PR #192, opened 2013-09-08
Textbox functions (commenting)

kanzure:
Name and comment some textbox-related functions.


kanzure:
Can't be merged, plus really old...

PR #193, opened 2013-09-08
Name functions related to game init

kanzure:
Maybe not the best names, but here you go. These functions setup wram and clear the screen and do other things. InitNewGame and SetupNewGame could probably be named better.


kanzure:
Can't be merged, plus really old.

PR #276, opened 2015-02-11
Blindly use wram labels for everything except mobile.

yenatch:
Also except maps/second_map_headers.asm, since it's easier to just redump later.


kanzure:
Does this build? Also, how do you know if the wrong label is being used in the wrong bank?

yenatch:
trial and error

PR #280, opened 2015-02-14
Fix make clean; Final music incbins

dannye:
no more .tx files
remove rgbds generated files, .map and .sym
also two unused music data incbins

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 PYTHON := python
 
 .SUFFIXES:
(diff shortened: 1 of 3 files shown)


kanzure:
> remove rgbds generated files, .map and .sym

Aren't those used by the tools?

dannye:
Afaik, the only reason we have ever made a .map file is because rgbds would segfault otherwise, but that has been fixed. I know, personally, I only use the .sym file for bgb.

Either way, cleaning up generated files from the build is a good idea, no?

PR #290, opened 2015-06-09
macros/event.asm: Fix macro changemap

PikalaxALT:
Syntax is now `changemap MapBlockDataBank, MapBlockDataPointer`


kanzure:
Was this macro already in use? And how much refactoring of the prior use is this going to require?

PR #294, opened 2015-06-26
minor, comment change

Guernouille:
The reason stated for egg moves removal in Crystal was wrong.


iimarckus:
Almost. The original comment is correct for Sweet Scent and Steel Wing, but you’re right that Charm and Lovely Kiss are unobtainable.

PR #296, opened 2015-06-29
Add a script to convert project-specific graphics.

yenatch:
Does this seem reasonable?


kanzure:
Hmm, why move it out of the submodule if there are still other submodule things being used?

yenatch:
Project paths should stay in the project. This intuitively makes sense to me. Besides that, they have no business being in a submodule, for a few reasons:
- The submodule has to be updated with any path changes
- The submodule can only be used with a parent repo that uses the same paths the same way, meaning it's not reusable, which is the entire point of a submodule
- Even more bloat

You could have the paths in a config file that the submodule reads, but why bother when it limits functionality, forces non-submodule logic into the submodule, increases reliance on the submodule, and is a lot more complicated than just having a script that calls the submodule? Also, when gfx.py is eventually replaced by rgbgfx, it will be easier to swap out in the project scripts.

The only other thing that the submodule is used for is scan_includes, which is a pretty simple script that doesn't really have any dependencies to justify having it in the submodule. It's only there because there's no other place for code that both pokecrystal and pokered use. But that's now the only piece of code that's shared (pokered can already be used with rgbgfx).

PR #311, opened 2015-08-24
Bank 4 time functions moved out of main.asm

PikalaxALT:


yenatch:
dunno what happened with 31de320, do `git reset --soft 726c4bc` and commit again

PR #313, opened 2015-08-28
'make pngs'-fix

JimB16:
I've fixed the 'gfx.py'-file so that we can use 'make pngs' now.
The main problem was that variabes like 'args['pic']' were set and then the functions in extras/gfx.py were called, which didn't expect this variable and throw errors.

The resulting pictures aren't correct, yet. The interleaving doesn't work everywhere correctly. I fixed it for the folder 'gfx/trainers' and for 'gfx/pics' (but only for the 1st part of the animation in the front.png-picture = the upper part of the picture).
for example gfx/pics/ampharos/front.png: front
And it also seems that the width of the picture is different for every Pkmn, that's the reason why not every front.png-picture looks right.
The palette for both Pkmn-pictures is now the normal one. The back-picture was created before with the shiny-palette.

I don't have tested if the pictures are now correctly compressed and converted back to the GBC. Since I've made a few changes, it shouldn't give the correct result.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -72,17 +72,12 @@ pokecrystal.gbc: $(crystal_obj)
 	rgblink -n $*.sym -m $*.map -o $@ $^
 	rgbfix -Cjv -i BYTE -k 01 -l 0x33 -m 0x10 -p 0 -r 3 -t PM_CRYSTAL $@
 
-compress := gfx/trainers/beauty
-compress2 := gfx/pics/egg/front
-
 
 pngs:
-#	-exec $(gfx) unlz $(compress2).2bpp.lz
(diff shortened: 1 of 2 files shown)


yenatch:
Sorry if someone misled you on irc. The issue is with `export_2bpp_to_png` (extras/pokemontools/gfx.py) missing a `**kwargs` argument in the function def. The changes in this pr aren't necessary.

JimB16:
I was unsure which gfx.py had to change. I also thought about changing the parameters of the functions, but I decided to make the changes here.
I searched a bit in the old versions of the files and couldn't find anything that changed that lead to this problem.

Issue #315, opened 2015-08-30
'make pngs'/gfx.py doesn't deinterleave the pictures


JimB16:
The 'make pngs'-command is fixed, but the resulting pictures aren't deinterleaved and look like this:
![front](https://cloud.githubusercontent.com/assets/13925587/9566995/6825e810-4f1b-11e5-82da-7a056d7aa67b.png) ![back](https://cloud.githubusercontent.com/assets/13925587/9566996/6a762364-4f1b-11e5-9b08-3c605fa15d5a.png)
In my 'pull request' I've also fixed that (partly).
If .2bpp are converted to .pngs the size of the picture must be set, becaus the script gets only the size of .pngs. So you have to add sth. like this if the fileformat isn't .png.

> args['pic_dimensions'] = 7, 7

For the trainer-pics it's always '7,7', the Pkmn-pics have different sizes: from 5 to 7 (that has to be added manually as far as I can tell).
I could add this myself, but I just wanted to be sure this time if this fix is wanted.

Update: I've tested now to include an edited .png of a trainer. And the result that came up in the emulator is the deinterleaved picture as 'make pngs' produces it now. So to it's necessary to fix this if you want to edit the graphics of the game, as far as I see it.

yenatch:
Does 8f74b553db37ddc1842476eb0d96ec3c8f1ac0ed fix this?

JimB16:
For the Poke- and Trainer-pics it's fixed now, but there are several pictures in gfx/ who can be interleaved as well.
At the moment I'm trying to find out these files and their dimensions

PR #316, opened 2015-08-31
added pictures to deinterleave

JimB16:
#315
I've added some pictures that have to be deintereaved to look normal.
It's not the most elegant way, but you can't just add all pictuers of one folder, because not every picture has to be deinterleaved.

### DIFF
--- a/gfx.py
+++ b/gfx.py
@@ -61,7 +61,34 @@ def filepath_rules(filepath):
 	elif 'gfx/intro' in filedir and name == 'suicune_jump':
 		args['pic'] = True
 
+	elif 'gfx/intro' in filedir and name == 'suicune_back':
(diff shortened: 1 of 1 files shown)


yenatch:
I'm not sure why this works. These graphics aren't interleaved, they're just missing width information.

yenatch:
This is also not required for uncompressed graphics. They should be committed as png instead.

JimB16:
I also did wonder a bit that in the graphics there were whole lines of the picture visible and not completely mixed up as the other graphics. This also didn't work for every graphic that has the wrong width.

If they're not interleaved, can it happen that gfx.py interleaves pictures that don't need it and show up corrupted in the ROM and then in the game?

PR #317, opened 2015-09-01
Pokemon and trainer pngs

yenatch:
Source graphics and their compressed counterparts can coexist, so long as the compressed versions aren't rebuilt. The `init` target attempts to remedy this.

Pokemon and trainer pics are low-hanging fruit.


yenatch:
Merging changes like this in future will break the dependency chain. To fix: `make; git checkout gfx; make`

PR #321, opened 2015-10-11
Changes in sound macros and commentating them

JimB16:
Some of the parameters for the sound macros can be split in smaller parts than byte-size. Pokered has some of this changes already implemented and I wanted to put it now into Pokecrystal.
This pull request is just a demonstration at the moment of the changes that I want to implement.

And while I'm playing with this code I try to write better comments for the different macros, so someone interested in using them doesn't have to look into the code of the audio engine (where most of the explanation is at the moment for the commands).

### DIFF
--- a/audio/music/titlescreen.asm
+++ b/audio/music/titlescreen.asm
@@ -7,16 +7,18 @@ Music_TitleScreen: ; eb808
 
 Music_TitleScreen_Ch1: ; eb814
 	tempo 134
-	volume $77
+	volume_red 7, 7
 	dutycycle $3
 	tone $0002
-	vibrato $10, $12
+	vibrato_red $10, 1, 2
 	stereopanning $f0
 	notetype $c, $a7
-	intensity $a0
+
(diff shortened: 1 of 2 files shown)


yenatch:
It would be good to have the macros be the same between red and crystal.

Please also mirror `music_commands` in https://github.com/pret/pokemon-reverse-engineering-tools/blob/master/pokemontools/crystal.py#L2470.

roukaour:
Merging this would break a lot of existing music asm files by Mmmmmm, FroggestSpirit, Pum, ShantyTown, and others.

Rangi42:
As long as macros/legacy.asm retains the old music macros, then old asm demixes will still work, and new ones can use more sensible commands. Plus when/if somebody writes a music tutorial it'll be nice to have better macros.

At least one more that could be changed: add some implicit state to `musicheader` so it doesn't need those redundant `1` arguments. (We've discussed this before, with Prism's `channelcount` and `channel` macros as an alternative example.)

PR #323, opened 2015-10-30
Use .lz.orig files to preserve compressed data.

yenatch:
#317 broke everything, this ought to fix most of it


yenatch:
known issues:
- girafarig has a repeated tile (first one in frames 3 and 5), so it's missing in generated data, including 2bpp
 - should the 'incorrect' 2bpp map to the correct lz?
- lugia and dewgong's bitmasks are supposed to be zeroed out
- several pokemon palettes are inverted
- trainers/swimmer_m.pal is inverted

yenatch:
i overthought this, the palette doesn't need to be sorted in the first place

yenatch:
#362

PR #325, opened 2015-11-04
Battle Tower, AI, battle labels

PikalaxALT:
Also dabbling in battle effects and mon pic animations


yenatch:
f58756c has nothing to do with the other commits

i also hate it a lot

yenatch:
where is `engine/evolution_animation.asm`

yenatch:
#326

PR #329, opened 2015-11-29
Clean up the command macros

yenatch:


macros/sound.asm
```diff
@@ -29,201 +29,207 @@ A_ EQU 10
 A# EQU 11
 B_ EQU 12
 
+NUM_OCTAVES = 8
 
 octave: macro
-	db $d8 - (\1)
+	if (\1) < 1 || (\1) > 8
```
PikalaxALT: Why is the 8 hard-coded?

PR #351, opened 2016-10-26
Change python to python2 in makefile for better compatability

lukeoftheaura:
Linux distros today are using `python` to mean python 3, so this as an edit to the makefile to specify python 2. So simple I doubt anyone cares but it fixed a broken build for me.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-PYTHON := python
+PYTHON := python2
 MD5 := md5sum -c --quiet
 
 .SUFFIXES:


iimarckus:
Some are, some aren’t. Others use different naming schemes (such as explicitly versioning python2.7, python3.5, etc.). That’s why it’s a variable—so you can run `make PYTHON=python2` without issue.

Not ideal, but so far there’s no solution that’s really any good, besides removing the need for Python completely.

yenatch:
Python is no longer needed.

PR #352, opened 2016-10-29
Comments

FredrIQ:
This fixes some incorrect comments, adds some that point out bugs/redundancies and fixes inverted labels in HandleStatBoostingItems.


YamaArashi:
These seem to check out.

PR #354, opened 2017-01-08
Fix a reversed label mistake and reference a major bug

FredrIQ:
I reversed LastEnemyCounterMove and LastPlayerCounterMove (they were in the wrong order).

I also added a comment to CheckTypeMatchup explaining how the rest of the code assumes the function works (the assumption is incorrect).


kanzure:
ACK 9b8f57a7308691f67f9025b5dcc621941dbc3afd and earlier

PR #357, opened 2017-03-14
Fix every section to their desired position

### LINKED ISSUE #356: Recent versions of rgbds no longer creates matching binaries
As of rgbds commit 1b05c43b979f8cc8c9bd571a7c96f13614724111 (this repository: github.com/rednex/rgbds ), the resulting binary of 'make' no longer match the actual binary.

Note that this repository links to another repo for rgbds, but it redirects to the above now. Also, not only is the result different, the resulting binary doesn't actually work (it runs but doesn't seem to respond to button input)

Ben10do:
Fixes #356. Hope this helps to recover from the recent changes to rgbds.

This guarantees that each section will be in the same place as the original game, without the risk of different linker versions repositioning sections.

This works with the current version of rgbds (`947d767c6487bbb131fa038ecc287ceb4c620fc5`), and will produce the correct v1.0 ROM (md5: `9f2922b235a5eeb78d65594e82ef5dde`), and the correct v1.1 checksum (`301899b8087289a6436b0a241fbbb474`) with this version.

As I've endeavoured to remove all 'floating' sections, there may be some unnecessary fixed sections (e.g. if there's only one section in a bank) – feel free to fix these if necessary.

### DIFF
--- a/audio.asm
+++ b/audio.asm
@@ -1,7 +1,7 @@
 INCLUDE "includes.asm"
 
 
-SECTION "Audio", ROMX, BANK[AUDIO]
+SECTION "Audio", ROMX[$4000], BANK[AUDIO]
 
 INCLUDE "audio/engine.asm"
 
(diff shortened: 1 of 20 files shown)


Ben10do:
#358 has the same goal as this PR, so take your pick!

Sanqui:
Thank you for your effort!

PR #358, opened 2017-03-15
Fix floating sections

AntonioND:
This is needed so that rgblink doesn't move them around after modifying
it's placement algorithm.


Sanqui:
Builds and matches under both rgbds 0.2.4 and master.

I'll accept this pull request over #357 because it also fixes VRAM and has lowercase hex (even though there isn't really a style guide).

Thank you!

PR #360, opened 2017-03-25
Implement a linkerscript (for rgbds 0.3.0 compatibility)

Ben10do:
A parallel to @AntonioND's pret/pokered#147. Again, this relies on rednex/rgbds#140, so wait until that's merged before merging this PR.

This linkerscript will sequentially allocate all the sections that were previously not fixed. This allows such sections to remain in place relative to each other, regardless of if their size is changed (e.g. by mods).

At present, any bank constants are included as comments next to the bank number (e.g. `ROMX $10 ; EVOS_ATTACKS`), and each section has a comment with its intended location in the ROM. If this is unhelpful, I'm happy to change this.

Only the minimum necessary ORGs are included. The need for these may be removed in the future by including more sections in the script.

As part of this change, I had to rename the “Audio” section in WRAM0 to “WRAM Audio“ (similar to some SRAM sections), as [there were conflicts](https://github.com/rednex/rgbds/pull/140#issuecomment-289124609) with the identically-named ROMX section.


Ben10do:
rednex/rgbds#140 has now been merged, and this linkerscript works correctly as of rgbds' master, producing the correct checksum. Feel free to look into suggesting improvements to and/or merging this PR.

NieDzejkob:
You need to change the rgbds version in `.travis.yml` to make it pass the checks ;)

Ben10do:
Good point, seeing as rgbds 0.3 has been out for a while – I shall rectify this now.

NieDzejkob:
ResidentSleeper

yenatch:
Is there a way to detect whether rgbds is 0.3.0 or higher?

AntonioND:
As far as I know, there is no direct way. The only way is to try to assemble a file with something only available in 0.3.0 (like the keyword ALIGN) and see if it returns error or not. I think that there were version strings in the command line help at some point but they were removed.

yenatch:
I supposed you could just check the usage for `-l`.

Ben10do:
I can't think of a way off the top of my head that isn't hacky, I'm afraid.

Here are a few more telltales I can think of:
- **Post-compile, pre-link:** Object files made by rgbasm 0.3.0 start with `RGB3`, and for later versions, `RGB4`. Previous versions will use `RGB0`-`RGB2`.
- **Pre-compile:** I know that OAM Sections were also added in rgbds 0.3.0, so you could detect if compilation succeeds with a file using such a section. You could also use ALIGNed sections (as per @AntonioND's suggestion), or see if WARN causes compilation to fail.
- **Mid-compile:** Check if the BANK of a label in HRAM is 0. However, as this was previously undefined, this would be *really* hacky, and could potentially result in some false positives.

And yes, of course, checking the usage for `-l` would be another method.

AntonioND:
Yeah, that's another option. Still very hacky.

yenatch:
#389

Thanks for your work on this feature.

Ben10do:
No problem, glad this has been sorted!

PR #365, opened 2017-06-09
Improvements to the tools and their Makefile

Ben10do:
This PR contains a number of tool-related changes:

- Make the tools implicitly when making the ROM
- Add a `clean` target to the tools Makefile
- Run said `clean` when cleaning the pokecrystal directory
- Add a `.gitignore` for the compiled tools
- Remove the `make tools` step in the installation instructions
- Fix warnings about assignments in 'if' statements (as produced by Clang)


yenatch:
202cea9 and by extension cf39024 do not work with `make -j`.

Ben10do:
So it doesn't… Forgive my lack of experience with parallelising Makefiles 😛

I'll add an extra commit with your cheat, if that makes `make -j` work properly.

yenatch:
Actually, I was wrong. That always evaluates it. You would have to use `=` and evaluate it in a recipe, which is too hacky for my liking.

Ben10do:
Fair enough; in which case, I won't add an extra commit with that cheat…

yenatch:
I think this is a reasonable compromise:

```make
ifeq ($(MAKECMDGOALS),)
_ := $(shell make -C tools)
endif
```

yenatch:
Ok, forget that.

```make
ifneq ($(MAKECMDGOALS),clean)
ifneq ($(MAKECMDGOALS),tools)
Makefile: tools
endif
endif
```

Only caveat is it has to go below the `all` rule, or it becomes the default target.

Ben10do:
OK, I'll give this a try.

PR #367, opened 2017-06-13
Missing a dependency

SkyrisBactera:
Without libpng-dev, sudo make install causes:
In file included from src/gfx/gb.c:20:0:
include/gfx/main.h:20:17: fatal error: png.h: No such file or directory
 #include <png.h>
                 ^
compilation terminated.
Makefile:90: recipe for target 'src/gfx/gb.o' failed
make: *** [src/gfx/gb.o] Error 1


yenatch:
Thanks.

PR #368, opened 2017-06-17
Fix section linking

PikalaxALT:
ROM builds to yield correct checksums.  Achieved by fixing sections.


xCrystal:
I honestly don't think this is desired. The main advantage of using the disassembly for hacking is not having to bother about fixed addresses and repointing, and these changes destroy that point. Correct me if I'm wrong, but I'm pretty sure rgbds v0.2.5 builds the ROM just fine; personally, I'd rather stick to an old/obsolete version of rgbds than hardcode sections until there is another solution. Sure, I probably shouldn't be complaining much since I've barely contributed to the disassembly but I don't think I'm speaking only for myself here.

PikalaxALT:
While in general I agree with you, this particular repository is intended to be a reference.  For it to serve as such, the assembler output needs to match the source material byte for byte.

The ideal solution would be to provide a linkerscript, which would allow sections to float while still building the correct ROM in an unmodified source.

yenatch:
This is redundant to both #358 (which I reverted!) and #360. In short, no. #360 will be merged eventually.

Ben10do:
One thing that this PR also does (which I forgot to do in #366) is replace `ld hl, [sp+0]` with `ld hl, sp+0`. If #360 is preferred over this, then it may also be worth making that change separately.

yenatch:
Please squash 2be9c11 into 25d8e25. I will merge this over #360 since it removes `BANK[]` and fixes unions. It would be a good idea to remove contents.asm as well.

yenatch:
Oh also please remove all the `org $4000` and `org $a000`.

yenatch:
I accidentally pushed master to PikalaxALT:master, truncating this pr.

PR #374, opened 2017-06-25
pokemon_animation's variable is not initialized

SnDream:
No initialization variables in pokemon_animation make the bitmask.asm and frames.asm generated in gfx/pics/*/ may contain wrong content and make Bank $34 out of space.


yenatch:
Nice catch. Thanks.

PR #381, opened 2017-08-29
Add constants for easier edits

i0brendan0:
A constant can be used to make it easier for someone to change the right numbers.

### DIFF
--- a/event/mom_phone.asm
+++ b/event/mom_phone.asm
@@ -56,7 +56,7 @@ MomTriesToBuySomething:: ; fcfec
 
 CheckBalance_MomItem2: ; fd044
 	ld a, [wWhichMomItem]
-	cp MOM_ITEM_LIST_2 ; 10
+	cp NUM_MOM_ITEMS_2 ; 10
 	jr nc, .nope
 	call GetItemFromMom
 	ld a, [hli]
@@ -99,7 +99,7 @@ CheckBalance_MomItem2: ; fd044
 
 .exact
 	call .AddMoney
-	ld a, MOM_ITEM_LIST_1 ; 5
+	ld a, NUM_MOM_ITEMS_1 ; 5
 	call RandomRange
 	inc a
 	ld [wWhichMomItemSet], a
@@ -194,7 +194,7 @@ GetItemFromMom: ; fd117
 
 .zero
 	ld a, [wWhichMomItem]
-	cp MOM_ITEM_LIST_2 ; 10
+	cp NUM_MOM_ITEMS_2 ; 10
 	jr c, .ok
 	xor a
 
--- a/event/mom_phone.asm
+++ b/event/mom_phone.asm
@@ -1,5 +1,5 @@
-MOM_ITEM_LIST_1 EQU 5  ; how many are in list 1?
-MOM_ITEM_LIST_2 EQU 10 ; how many are in list 2?
+NUM_MOM_ITEMS_1 EQUS ((MomItems_1End - MomItems_1) / 8)
+NUM_MOM_ITEMS_2 EQUS ((MomItems_2End - MomItems_2) / 8)
 
 MomTriesToBuySomething:: ; fcfec
 	ld a, [wMapReentryScriptQueueFlag]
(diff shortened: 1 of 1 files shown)


event/mom_phone.asm
```diff
@@ -191,7 +194,7 @@ GetItemFromMom: ; fd117
 
 .zero
 	ld a, [wWhichMomItem]
-	cp 10 ; length of MomItems_2
+	cp MOM_ITEM_LIST_1 ; 10
```
slx7R4GDZM: Shouldn't this be MOM_ITEM_LIST_2?

i0brendan0:
Darnit. Yes. I just copied and I forgot to change that to a 2. Ugh.

i0brendan0:
Fixed now.

xCrystal:
Don't want to be that guy, but MOM_ITEM_LIST_1_LENGTH and MOM_ITEM_LIST_2_LENGTH would probably be more appropriate. Also, I haven't been following pokecrystal's standards closely, but wouldn't the constant definitions belong to a constants file? Or maybe it's fine as long as their scope is only the file they are defined in?

i0brendan0:
No that's probably a better name convention. I'm terrible with coming up with names, so I just gave them the first thing that made sense. And they are never used anywhere else I think. So leaving them in that file probably is O.K. I personally would like them near the actual lists but I was told they have to be declared before being used.

yenatch:
Localized constants are fine.

yenatch:
It might be easier to edit if it automatically found the size of the list:
```assembly
NUM_MOM_ITEMS_1 EQUS ((MomItems_1End - MomItems_1) / 8)
NUM_MOM_ITEMS_2 EQUS ((MomItems_2End - MomItems_2) / 8)
```
```diff
 MomItems_1: ; fd136
 	momitem      0,   600, MOM_ITEM, SUPER_POTION
 	momitem      0,    90, MOM_ITEM, ANTIDOTE
 	momitem      0,   180, MOM_ITEM, POKE_BALL
 	momitem      0,   450, MOM_ITEM, ESCAPE_ROPE
 	momitem      0,   500, MOM_ITEM, GREAT_BALL
+MomItems_1End:
+
 MomItems_2: ; fd15e
 	momitem    900,   600, MOM_ITEM, SUPER_POTION
 	momitem   4000,   270, MOM_ITEM, REPEL
 	momitem   7000,   600, MOM_ITEM, SUPER_POTION
 	momitem  10000,  1800, MOM_DOLL, DECO_CHARMANDER_DOLL
 	momitem  15000,  3000, MOM_ITEM, MOON_STONE
 	momitem  19000,   600, MOM_ITEM, SUPER_POTION
 	momitem  30000,  4800, MOM_DOLL, DECO_CLEFAIRY_DOLL
 	momitem  40000,   900, MOM_ITEM, HYPER_POTION
 	momitem  50000,  8000, MOM_DOLL, DECO_PIKACHU_DOLL
 	momitem 100000, 22800, MOM_DOLL, DECO_BIG_SNORLAX_DOLL
+MomItems_2End:
; fd1ae
```

i0brendan0:
Are you sure each line is 8 bytes (I think that's what the 8 means)? Im pretty sure the item/doll I'd is 1 byte. Not sure what the MOM_(ITEM/DOLL) is. But I'm sure the money values are more then 3 bytes each. 4 Bytes is 65535 and the possible value is more than that. So either 6 or 8 bytes for each money value. So minimum would be 6+6+1+1= 14. But as I don't actually know off of the top of my head I could be wrong and 8 could be the right number.

yenatch:
65535 (0xffff) is two bytes. each money amount is three bytes, so 3+3+1+1

```assembly
 momitem: macro
 ; money to trigger, cost, kind, item
 	dt \1
 	dt \2
 	db \3, \4
ENDM
```
`dt` is 3 bytes

i0brendan0:
Right. I'm a little out of it. I was mistaking bytes for F.

i0brendan0:
So I did what @yenatch said for dynamic range of lists and now it doesn't work.

i0brendan0:
Ok so fixed the syntax error `EQU` vs `EQUS` and now it says that `MomItems_1End` and `MomItems_2End` aren't defined.

i0brendan0:
Ok this is fixed finally.

yenatch:
Merged in #423.

PR #384, opened 2017-08-31
Correct constant to remove a bug

i0brendan0:
If the ds 2 that is behind OTClassName is deleted or decreased, the game will continue to display characters for trainer classes that are 12 or 13 characters long, including the @ terminator. The correct constant has been put in place and the ds 2 removed, but no one would have known it was wrong if they hadn't have removed the ds 2.


yenatch:
Good catch. There are probably more like this. The 1/2-byte trailing spaces in wram.asm should be reexamined.

PR #385, opened 2017-09-02
Rename Psychic from type constant to move constant

i0brendan0:
Changes the TM to the move name constant and renames the TM/HM compatibility list of each affected Pokemon to take for the change. I don't think I missed any but if I did let me know.


yenatch:
how did aca7d6d pass...

PR #388, opened 2017-09-10
Moving a few images out of gfx/unknown

surskitty:


yenatch:
The filenames also need to be changed in the makefile.

surskitty:
I believe it's fixed :Db

PR #394, opened 2017-10-21
Adjusting Indigo League map labels

surskitty:
Apologies for this not being two separate pull requests; can't figure out how to delete a commit off of it.

flex had not been automatically installed on my box and so should be in the dependencies list.


yenatch:
Rather than delete commits you usually rewind them.
```
git reset HEAD^
git push -f origin master
```

surskitty:
That's fixed, then.  Thanks!

PR #397, opened 2017-11-15
A few more labels and comments

mid-kid:
Here I'll just request for my commits. They're pretty self-contained and well-explained, so I won't do it again in this description. Just cherry-pick whatever you think is worthy.


battle/moves/move_effects.asm
```diff
@@ -1387,10 +1229,6 @@ Whirlpool:
 	endmove
 
 SuperFang:
```
huderlem: What is the reason for leaving labels like SuperFang and Psywave?  There are a few other cases like that in this file.

mid-kid: Because while they share the same effect pointer, they use a different effect ID. It was either this or adding ` ; EFFECT_SUPER_FANG` behind the appropiate pointer. I thought this looked reasonably good.

PR #399, opened 2017-11-24
Early anim_obj documentation

henrxv:
Simple ways to use the anim_obj parameters

### DIFF
--- a/battle/anims.asm
+++ b/battle/anims.asm
@@ -288,21 +288,15 @@ BattleAnim_MirrorMove: ; c929b
 	anim_ret
 ; c929c
 
-;General info anim_obj command:
+;Basic anim_obj command info :
 ;******************************
-;anim_obj ANIM_OBJ_#	x1, x2,   y1, y2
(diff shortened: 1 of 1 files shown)


yenatch:
weird to have this in the middle of the file

the x1/y2 explanation is a bit complicated. x1 is just in tiles, and x2 is in px. i guess the idea must have been that you could pick whichever was more convenient, but i think these macros are gonna be reverted to just px soon, so this comment is gonna be edited anyway

there's no formal style guide but the style for the rest of the repo is to put a space between the semicolon and the comment: `; comment`

henrxv:
Ah sorry i really just wanted to provide some of the notable values for
easy use of the anims they way they were when I downloaded the project.

Removed the formula part and made the X2 Y2 part clearer.

Hope it's better but in any case it was to help you guys with the amazing
project! Modify it as you like :)

roukaour:
Thanks for your effort, @henrxv ! If you make any more progress, can you please fill out docs/move_anim_commands.md? So far it (and most of the other docs/ files) are just lists of command IDs, with no explanation.

PR #404, opened 2017-11-30
De-Magic-Numbers the Rocket Grunts

surskitty:


yenatch:
Thanks. This was initially done on purpose, since `GRUNTM, GRUNTM_20` didn't really add anything over `GRUNTM, 20`, but you're right that the constants are easier to change/remove.

PR #409, opened 2017-12-09
Document constants with comments, actual names, and more thorough usage

roukaour:
Constants are commented with the code and file they refer to.
Redundant constants like `NUM_WILDMONS_PER_AREA_TIME_OF_DAY` and `NUM_GRASSMON` have been merged.
sprite_constants.asm in particular has been filled out with meaningful names.


Sanqui:
I love this

macros/charmap.asm
```diff
@@ -152,7 +188,7 @@
 	charmap "♂",        $ef
 	charmap "¥",        $f0
 	charmap "×",        $f1
-	charmap "·",        $f2
+	charmap "<PERIOD>", $f2 ; same as "." in English
```
yenatch: `·` is a different character (https://en.wikipedia.org/wiki/Interpunct). `<DOT>` might be better.

roukaour: The tile looks identical to the period; it's used as a decimal point in PrintNum, and sometimes to abbreviate "`<ID>№`" (other times a period or slash is used). I'll rename it to "`<DOT>`".

yenatch: hmm, which one am i thinking of then. there's one that is a dot in the middle. or was that the ellipsis.

roukaour: You're thinking of `charmap "·", $74 ; middle dot, unused` in gfx/misc/font_extra.png.

Guernouille:
I'm getting a lot of "already defined" errors from the NEXTU ENDU and UNION constants after this merge

yenatch:
`UNION` was added in rgbasm 0.3.3. If you have an earlier version it will think it's a label and spit out that error.

Guernouille:
Ooooooh. I'll install the new version then. Thank you!

PR #411, opened 2017-12-10
Clean up more labels, constants, and file names

roukaour:
* Move lots of data tables into individual files (for now, until a final file structure is agreed on)
* Rename tilesets and overworld sprites, consistent with their constant names
* Consistent capitalization for map names, matching their constants
* Single files for tilesets/data.asm and maps/blockdata.asm
* Document macros/map.asm
* `Trainer*AfterBattleText` labels
* All map triggers use the `maptrigger` macro, not `dw`
* Comment palette colors and roof map groups
* Correct the `TEXT_SPEED` constants
* More accurate and user-friendly palette constants
* Use `MORN`/`DAY`/`NITE` shifts for time-based person_events
* Prefix `FLOOR` to the `elevfloor` constants


yenatch:
`Daycare` is correct.

roukaour:
Really? It's "DAY-CARE" in-game, and the map was already called DayCare.asm/DayCare.blk; I just changed the constant and various labels to match.

yenatch:
That's unfortunate. I was aware that the map files were called that but assumed it was a mistake.

roukaour:
So should I use the names `Daycare`/`DAYCARE` instead of `DayCare`/`DAY_CARE`? They're both correct names for real-world day cares/daycares.

yenatch:
Let's stick with the in-game name.

yenatch:
I just noticed that some labels were changed to PokeCenter. but i think the map constants are `POKECENTER`. There isn't an in-game name here but I think it should be one word.

roukaour:
The blk+asm files were already named PokeCenter; I did consider changing that to Pokecenter to match the constant, but assumed they were that way for a reason. So consistently one word is preferable?

yenatch:
yeah either way should be consistent. i wouldnt trust the blk/asm names generally

roukaour:
Okay, I don't expect to add any more major edits to this PR.

gfx/pics.asm
```diff
@@ -1,18 +1,21 @@
...
-
 SECTION "Pic Pointers", ROMX
+
 PicPointers:: INCLUDE "gfx/pics/pic_pointers.asm"
 
+
 SECTION "Unown Pic Pointers", ROMX
+
+; Unown pic pointers are assumed to start at the same address as PicPointers,
+; but in a different bank.
```
yenatch: This should be enforced in the linkerscript as well.

engine/phone.asm
```diff
@@ -165,7 +165,7 @@ CheckPhoneContactTimeOfDay: ; 900ad (24:40ad)
 
 	callba CheckTime
 	pop af
-	and (1 << MORN) + (1 << DAY) + (1 << NITE)
+	and MORN + DAY + NITE
```
yenatch: `MORN | DAY | NITE`

roukaour: Done; IIRC there are other places which use + for combining masks, will keep an eye out for them.

gfx/pics.asm
roukaour: Done.

engine/options_menu.asm
```diff
@@ -373,57 +380,56 @@ Options_Print: ; e4424
 ; e445a
 
 .Strings:
+; entries correspond to PRINT_* constants
```
yenatch: `GBPRINTER_*`

roukaour: No, the `PRINT_*` constants are the 0-5 indexes; `GBPRINTER_*` are the raw brightness values. Likewise for `TEXT_SPEED_*` (indexes) versus `TEXT_DELAY_*` (frame delays).

On second thought, those are bad names. I'm putting `OPT_` in from of the option index constants.

PR #419, opened 2017-12-13
More reorganization and documentation

roukaour:
* Document more bugs and glitches
* Start documenting effect commands
* Move code from main.asm into individual engine/ and event/ files
* Reorganize macros/
* Split items/ and trainers/ into their data/ and engine/ components
* Rename the Onix trade house to match the other trade houses
* Use constants for gender ratios
* Macro-ify the unused palette map

Also, `tools/toc.py` can auto-generate tables of contents in Markdown files, replacing `## TOC` with a list of links to headings and sub-headings. I've used it for FAQ.md and docs/bugs_and_glitches.md.


docs/bugs_and_glitches.md
```diff
@@ -1,6 +1,43 @@
...
+- [NPC use of Full Heal or Full Restore does not cure Nightmare status](#npc-use-of-full-heal-or-full-restore-does-not-cure-nightmare-status)
+- ["Smart" AI encourages Mean Look if its own Pokémon is badly poisoned](#smart-ai-encourages-mean-look-if-its-own-pokémon-is-badly-poisoned)
+- [A Disabled, PP Up–enhanced move may not trigger automatic Struggling](#a-disabled-pp-upenhanced-move-may-not-trigger-automatic-struggling)
+- [Counter and Mirror Coat still work if the opponent uses an item](#counter-and-mirror-coat-still-work-if-the-opponent-uses-an-item)
+- [Present damage is incorrect in link battles](#present-damage-is-incorrect-in-link-battles)
+- [BRN/PSN/PAR do not affect catch rate](#brnpsnpar-do-not-affect-catch-rate)
+- [Moon Ball does not boost catch rate](#moon-ball-does-not-boost-catch-rate)
+- [Love Ball boosts catch rate for the wrong gender](#love-ball-boosts-catch-rate-for-the-wrong-gender)
+- [Fast Ball only boosts catch rate for three Pokémon](#fast-ball-only-boosts-catch-rate-for-three-pokémon)
+- [Friend Ball catches sent to the PC overwrite the wrong Pokémon's happiness](#friend-ball-catches-sent-to-the-pc-overwrite-the-wrong-pokémons-happiness)
```
xCrystal: This is actually not a bug. The ``SetBoxMonCaughtData`` at https://github.com/pret/pokecrystal/blob/master/items/item_effects.asm#L623 happens before the box Pokemon happiness is written to.

roukaour: Interesting! This GlitchCity thread agrees: <https://forums.glitchcity.info/index.php?topic=8105.0> I do wonder why the comment has been there all this time though.

xCrystal: While we're at it, I'm lazy to properly submit a PR, but I know of two more battle bugs:
- Beat Up error, whose cause is located [here](https://github.com/pret/pokecrystal/blob/d0d61cb/battle/effect_commands.asm#L3194) (should be cp c), and a more detailed explanation can be found [here](https://www.youtube.com/watch?v=202-iAsrIa).
- [Substitute + Defense Down side effect error](https://www.youtube.com/watch?v=OGwKPRJLaaI) (see video description and https://github.com/pret/pokecrystal/blob/fd14253/battle/moves/move_effects.asm#L556-L577; the second ``effectchance`` shouldn't be there)

roukaour: Thanks! The Beat Up video is unavailable; is it private, or the link wrong?

roukaour:
I believe this is in a state to be merged whenever, assuming it passes code review.

docs/bugs_and_glitches.md
```diff
@@ -256,7 +256,18 @@ This is a bug with `CheckPlayerHasUsableMoves` in [battle/core.asm](/battle/core
 
 This bug affects Attract, Curse, Foresight, Mean Look, Mimic, Nightmare, Spider Web, Transform, and stat-lowering effects of moves like String Shot or Bubble during the semi-invulnerable turn of Fly or Dig.
 
-*To do:* Identify specific code causing this bug and fix it.
```
xCrystal: This bug affects virtually all moves and the problem seems to come from https://github.com/roukaour/pokecrystal/blob/2fa1e97a/battle/effect_commands.asm#L1823
More precisely, it seems to be that the line after the caller (https://github.com/roukaour/pokecrystal/blob/2fa1e97a/battle/effect_commands.asm#L1717) should've been ret z instead.

roukaour: `.LockOn` returns `nz` if the move will definitely hit, `z` if it might still miss. `BattleCommand_CheckHit` returns early if the move will definitely hit; if the move misses, it has to set `[AttackMissed]` to 1. So `call .LockOn; ret nz` looks correct.

xCrystal: in any case, `.lockOn` returns nz when the target is during Fly or Dig, causing BattleCommand_CheckHit to return early and ensure the hit.

roukaour: I'm just not sure L1717 `ret nz` is the one to change. Wouldn't that make moves always hit when they're *not* locked on? The control flow within `.LockOn` looks complex enough that it might be giving the wrong result in some cases.

xCrystal: Yeah, my bad. That change would lead to non-lock on scenarios being treated as locked on. I'm not sure about potential side effects but on the other hand replacing the ``ret nz``, with ``jp nz, .Miss`` could do the trick (although this does seem to be ``.FlyDigMoves`` work, so ``.LockOn`` itself may need a rework. Basically, you want to catch the cases where the target is locked on but not during Fly/Dig, in order to ensure the hit in those. That said, it would make more sense if ``.FlyDigMoves`` was called before ``.LockOn`` so that it doesn't have to handle more complex scenarios than necessary.

i0brendan0:
So the location that you guys are talking about for lock on isn't even related to status moves and secondary effects. That's for the damage moves. It works flawlessly. All moves hit except those 3 moves against flying foes. I checked this out before and it most likely is that game freak forgot to write the code for status moves to hit with lock on. So while damage still occurs, status moves and secondary effects aren't checking for lock on and thus don't work.

roukaour:
@i0brendan0 So the flaw is in `CheckHiddenOpponent` after all? (Note how it's called by `BattleCommand_StatDown`, and reportedly this bug affects "stat-lowering effects of moves like String Shot or Bubble".)

i0brendan0:
@roukaour yes exactly. As you can see there is no code to check for lock on. It's not like gamefreak was writing and used the wrong condition to jump like they normally do, but there's actually no repeated code at all. I have a feeling the writing of the fix is fairly simple. Also copy paste from the damage version.

home.asm
```diff
@@ -1035,13 +1035,8 @@ IsHM:: ; 34df
...
 ; 34e7
 
-IsHMMove:: ; 34e7
-	ld hl, HMMoves
-	ld de, 1
-	jp IsInArray
 
-
-INCLUDE "data/hm_moves.asm"
+INCLUDE "home/hm_moves.asm"
```
yenatch: I think the surrounding functions should be home/moves.asm and HMMoves should be put back into data/. Ignore if this is already done.

macros/code.asm
```diff
@@ -0,0 +1,55 @@
...
+; 	cp 30
+; 	jr nc, .loop
+x = 1
+rept 8
+IF \1 > x
+x = (x + 1) * 2 +- 1
+ENDC
+endr
+	and x
+endm
```
yenatch: something less likely to collide with other variables might save someone a lot of confusion later (`__x` or similar)

home.asm
roukaour: There was disagreement over whether the table should go in data/ because it's data, or home/ because it's in bank 0. Just let me know which it should be in; or just merge and move it where you prefer. I just want it to be in a file *some*where, not buried in home.asm, because it's an important table for users to edit.

macros/code.asm
roukaour: Agreed, now that the variable is local to the macro (previously the maskbits-calling code had to do `and x` afterwards). Will change it.

roukaour: I went to change this, but found that `x` is commonly used as a local variable name in macros. Should they each have unique names (`_maskbits_x`, `_sinewave_x`, `_frame_x`, `_dx_x`, etc) or just leave it as-is?

roukaour: (The `maskbits` callers no longer need to refer to `x` themselves, since `and x` is inside `maskbits`, so there shouldn't be interference with other macros.)

home.asm
roukaour: In the meantime I've put the surrounding functions in their own home/ file.

main.asm
```diff
@@ -1941,6 +1941,12 @@ ScrollBGMapPalettes:: ; 4c03f
 
 INCLUDE "tilesets/palette_maps.asm"
 
+Unknown_4ce05: ; unreferenced
+rept 26
+	db $06
+endr
+; 0x4ce1f
+
```
yenatch: ?

roukaour: Ask Game Freak. :P

constants/radio_constants.asm
```diff
@@ -123,3 +123,6 @@ NUM_PASSWORDS_PER_CATEGORY EQU 3
 	const BUENA_ITEM
 	const BUENA_MOVE
 	const BUENA_STRING
+
+; BuenaPrizeItems table length (see data/items/buena_prizes.asm)
+NUM_BUENA_PRIZES EQU 9
```
yenatch: `NUM_BUENA_PRIZES EQUS "(BuenaPrizeItemsEnd - BuenaPrizeItems) /  2"` and should go back into event/buena.asm

roukaour: Can't; the table is defined after the places where `NUM_BUENA_PRIZES` is used, and it won't compile. I can define `NUM_BUENA_PRIZES EQU 9` right after the table instead. (Previously there was no constant, just a hard-coded `db 9` and `db 1, 2, 3, 4, 5, 6, 7, 8, 9`.)

audio/engine.asm
```diff
@@ -537,9 +537,9 @@ _CheckSFX: ; e82e7
 
 PlayDanger: ; e8307
 	ld a, [Danger]
-	bit 7, a
+	bit DANGER_ON_F, a
 	ret z
-	and $7f
+	and $ff - (1 << DANGER_ON_F)
```
yenatch: `and ~(1 << DANGER_ON_F)` or `and ~DANGER_ON`

roukaour: Can't; `~` doesn't do 8-bit negation.

docs/bugs_and_glitches.md
```diff
@@ -943,6 +944,17 @@ The exact cause is unknown, but a workaround exists for `DexEntryScreen_MenuActi
...
 
+## In-battle “`…`” ellipsis is too high
+
+This is a mistake with the “`…`” tile in [gfx/battle/hp_exp_bar_border.png](/gfx/battle/hp_exp_bar_border.png):
+
+![image](https://i.imgur.com/hVK5ofi.png)
+
+**Fix:** Lower the ellipsis by two pixels:
+
+![image](https://i.imgur.com/qfMvhlG.png)
```
yenatch: These images should be in docs/.

audio/engine.asm
yenatch: weird

constants/radio_constants.asm
yenatch: The equs will still work, it doesn't have to be equ

yenatch: oh its rept

roukaour: I tried `EQUS` and that didn't work either. The expression's value is not defined at the time `rept` needs it.

wram.asm
```diff
@@ -2816,52 +2825,54 @@ SECTION "Battle Tower", WRAMX
...
 w3_d88f:: ds 5
 w3_d894:: ds 1
 w3_d895:: ds 11
 w3_d8a0:: ds 1
 w3_d8a1:: ds 1
 w3_d8a2:: ds 1
-w3_d8a3:: ds $19d
-w3_da40:: ds $1c0
+w3_d8a3:: ds 1
+ENDU ; d869
```
yenatch: looks like d8a4

PR #420, opened 2017-12-14
Partial Phone Script Labels

surskitty:
I'm going through these pretty much at random, so apologies for that.

I would much appreciate it if someone went through and split up engine/phone_scripts and engine/more_phone_scripts into some kind of logical order.  It's complicated by the fact they aren't consistent at all, and of course that most of them are still unlabeled, but they're a complete mess.


surskitty:
phone_scripts and more_phone_scripts could be sensibly broken up into two files each.

The first half of `engine/phone_scripts.asm` is the logic for each NPC to handle what they're calling about, while the second half is selecting the appropriate greeting based on the time of day.  Line 1908 is where I would break them up.

`engine/more_phone_scripts.asm`, meanwhile, has the first half as logic to select the correct NPC's text-block for the same kind of conversation, and then the second half is specific to various NPCs to pick what they're blathering on about when they've got multiple potential topics.  Line 963 is the transition there.

I'm not intending on splitting them up myself, but I think it's a good idea.  I haven't gone through it enough to figure out what's happening in `text/phone/extra.asm` and `text/phone/extra2.asm` yet, though.  I'm prioritising getting the logic clear and then I can label the text later.

PR #422, opened 2017-12-24
Split all of the footprints

mid-kid:
Split all the footprints into separate files in `gfx/footprints/`. Makes them
easier to edit and find for most people.


yenatch:
I don't think the double incbin is the right solution. I'll see if I can figure out a hacky macro that still passes as an incbin in scan_includes. Otherwise I think footprints.1bpp should be concatenated from the other files by a hypo `tools/pokemon_footprints`.

roukaour:
Personally I'm not bothered by the double INCBIN. There are other files designed for easy editing that get processed to match the internal format, like the animation graphics+data. And if it weren't for "whoever was editing footprints forgot to fix their tile editor" (as commented in engine/pokedex.asm), the footprints would naturally work as a series of normally-INCBINned 1bpp files.

yenatch:
`cat` wouldn't make sense since the order has to be defined somehow... not sure the best way to do that  either

roukaour:
How about ImageMagick/GraphicsMagick `montage`?

yenatch:
defeats the purpose. i've warmed up a bit to the double incbin idea. there's already a lot of tables to change and this one is at least localized.

Issue #345, opened 2016-06-16
Some label semi-misnomers:


luckytyphlosion:
Not much of an issue, just some things I noticed that should be changed:
- [x] GetMonFrontpic should be something more like "RequestEnemyFrontpic" as it only applies to the enemy and there are certain scenarios when the frontpic isn't needed (minimized, sub). This also applies to GetMonBackpic
- [x] FrontpicPredef should be renamed to something like "GetAnimatedFrontpic", and GetFrontpic should be renamed to "GetMonFrontPic"
- [x] CutAndPasteTilemap doesn't actually cut and paste the tilemap; it converts the 20x18 tilemap into a 32x18 tilemap so it can be HDMA'd.
- [ ] LoadEDTile is entirely errorneous, it should be something like "CopyTilemapAtOnce".

yenatch:
LoadEDTile still needs to be fixed.

PR #424, opened 2017-12-24
Rename a routine and some maps; remove all code from main.asm

roukaour:
- Rename: LoadEDTile → CopyTilemapAtOnce (fix issue #345)
- Remove ambiguity: Kanto's Underground Path vs. Goldenrod's Underground Warehouse


engine/routines/checkpokerus.asm
```diff
@@ -0,0 +1,25 @@
+CheckPokerus: ; 4d860
```
yenatch: I think this one can just be engine/check_pokerus.asm. Even though it's small it's more visible, and it's often the first thing people look for.

roukaour: Moved it to engine/pokerus/ along with the other Pokérus-related engine files. We can probably do the same for other engine/ subsystems.

PR #433, opened 2017-12-25
More code cleanup

roukaour:
- Clean up color/palette code
- Use `HIGH()` and `LOW()` directives
- Improve documentation


engine/experience.asm
```diff
@@ -6,7 +6,7 @@ CalcLevel: ; 50e1b
 .next_level
 	inc d
 	ld a, d
-	cp (MAX_LEVEL + 1) % $100
+	cp MAX_LEVEL + 1
```
yenatch: This was here so that a max level of 255 would work. Otherwise it's a "not 8 bit" error.

engine/item_effects.asm
```diff
@@ -892,10 +892,10 @@ HeavyBallMultiplier:
...
 .WeightsTable:
 ; weight factor, boost
-	db 2048 >> 8, 0
-	db 3072 >> 8, 20
-	db 4096 >> 8, 30
-	db 65280 >> 8, 40
+	db HIGH(2048),   0
+	db HIGH(3072),  20
+	db HIGH(4096),  30
+	db HIGH(65280), 40
```
yenatch: It looks like this function can infinite if the upper half of the weight is 0xff.

engine/experience.asm
roukaour: Oops, will fix that.

engine/item_effects.asm
roukaour: `HIGH(65280)` = `0xff`, so nothing higher is possible.

event/happiness_egg.asm
```diff
@@ -181,9 +181,9 @@ DayCareStep:: ; 7282
 	dec hl
 	inc [hl]
 	ld a, [hl]
-	cp 5242880 / $10000
+	cp HIGH($500000 >> 8)
 	jr c, .check_egg
-	ld a, 5242880 / $10000
+	ld a, HIGH($500000 >> 8)
```
yenatch: `MAX_EXP_DAY_CARE`? I always assumed you could go all the way to 100 in the daycare.

engine/item_effects.asm
yenatch: Right, and it only breaks on less than.

battle/effect_commands.asm
```diff
@@ -730,7 +730,7 @@ BattleCommand_CheckObedience: ; 343db
 
 	; risingbadge
 	bit RISINGBADGE, [hl]
-	ld a, MAX_LEVEL + 1
+	ld a, LOW(MAX_LEVEL + 1)
```
yenatch: This could cause nothing to obey if the max level was 255. `MAX(MAX_LEVEL, $ff)` would at least allow anything below 255 to obey. Not sure if `MAX()` exists.

event/happiness_egg.asm
roukaour: 5,242,880 is more than a Pokémon can ever gain, but yeah, I'll add a constant.

battle/effect_commands.asm
yenatch: I think in the other case the way the code was written allowed it to work. It's ok to error if it can't easily work.

PR #435, opened 2017-12-26
Split maps/ into maps/blks/ and maps/scripts/

roukaour:
All .blk files go in maps/blks, all .asm files go in maps/scripts. This is similar to pokered's organization of maps/\*.blk and scripts/\*.asm.

I kept the top-level maps/ directory, but it might even be preferred to move that to data/maps, or separate them as data/map_scripts/ and data/map_blks/.

Either way, I think this is a helpful approach because:

- There are many possible ways to divide maps into subdirectories (by group, environment, region, landmark, etc). Most of them still need a directory for exceptions to the rule, maps/shared/ or maps/misc/ or so on.
- It's easier to work with a single flat directory than with a tree of subdirectories. No need for `-r` flags in Bash, or the relatively poor recursive search/filter features of Explorer. (Personal bias: it's also easier to write Polished Map when assuming one location for .blks.)
- Keeping scripts and block data together can be unwieldy. 302 .blks + 388 .asms = 690 files in one directory, with the important (second_)map_headers files mixed in there (easy to accidentally modify when processing all maps with some Bash or Python script). Hacks can end up with over 1000 files there, which isn't even fully viewable on GitHub.com.
- This is consistent with the general principle of keeping related data together and separate data apart. text/phone/ does not have subdirectories for gender; data/pokemon/base_stats does not separate the Gen 1 and Gen 2 Pokémon; etc. At the same time, different kinds of Pokémon data are kept apart: there isn't one data/pokemon/pikachu directory that contains its base_data.asm, dex_entry.asm, footprint.png, front.png, and so on. (That could be a viable way to organize it, but it's not pokecrystal's current approach.)

(An exception to this principle: tilesets/, which mixes three kinds of data for each tileset. Arguably justified since tilesets/data.asm also mixes them up. I could still see moving them to data/tilesets/, and moving the palette maps to data/palettes/tilesets/—metatiles.bin and collision.asm files both have per-block data and are usually edited together, but the palette maps apply to tiles, not blocks, and are listed in their own file tilesets/palette_maps.asm. Anyway, this PR is just about maps.)

I don't expect this to be agreed with and merged right away, just advancing it as a possibility for reorganization later on.


yenatch:
i think both maps/ and tilesets/ should go in data/

you're right that this makes sense until we come to a decision on grouping data. it doesn't really solve the 500 files problem but it's less noise at least

roukaour:
Two more questions:

- Should the beta/unused maps go in blks/unused?
- Can I separate the tileset-related palette_map.asm files from the blockset-related collision.asm and metatiles.bin files?

yenatch:
- beta maps should be renamed to unused. dont know about a subdir. maybe.
- the palmap should probably be with the graphics either way, but is there a reason to have any of the tileset data separate

mid-kid:
I'm probably in the minority here, but here I'll write my humble opinion on the issue:
* Dividing things into subdirectories per filetype (especially when they're discernible by extension) isn't a good idea. If we're to keep the principle of "related data is together", the blocks and script of a single map should be considered as a single instance of related data.
* The `-r` option of many GNU userspace utilities is a given with most of the project anyway, there's no real benefit to having a flat directory for maps. Polished Map should still, imo, allow for manually setting the project directory instead of guessing it out of the location of the .blk file.
* You're right, we shouldn't have everything in a flat directory. map_headers should be somewhere without, or with very little map scripts around.
* We don't split base stats for gen1 and 2 pokémon because there's just 251 of them, which is just barely manageable, and are all really easy to find, due to having one single canonical name all across the codebase. The case with map scripts, is that when we start prefixing names (such as `GoldenrodCity*`), we might as well just create a subdirectory for them. Right now, they're not really easy to find, either (is it `NewBarkTownLab` or `ElmsLab`? What is the player's house even called?). I'd be fine with a flat directory if they were all consistently named, but I'm also of the opinion that per-area subdirectories would make it much easier to find things.
* The `maps/` directory is prevalent enough to warrant a top-level directory, and a separate `.o` file. We want to prevent very deep directory trees, which would be the case if we decide to split up the directory and put it in `data/`.

roukaour:
The reason I'd separate palette_maps from collisions and metatiles is because they aren't all "tileset data". gfx/tilesets/\*.2bpp.lz and tilesets/\*_palette_map.asm are tileset data; tilesets/\*_metatiles.bin and tilesets/*_collision.asm are blockset/metatileset data. (Should we standardize on one term for those?)

I'm not a fan of per-area directories for a few reasons:

- The filename prefixes make them redundant: `maps/goldenrodcity/GoldenrodCity.asm`
- Many areas like routes or caves would have just one or two maps each.
- Some maps are ambiguous as to location: `UndergroundPath`, `Pokecenter2F.asm`, `Pokecenter1F.blk`. They'd need a maps/shared/ directory.
- You can start making exceptions, like a single maps/routes/ directory or grouping caves with their nearby area, but that leads to arguable judgement calls. Like, which route do DarkCaveBlackthornEntrance and DarkCaveVioletEntrance go on? Should three-floor Sprout Tower get its own directory? What about ten-floor Tin Tower? Will DayCare be stuck in the large routes/ directory because it's on Route 34?
- Any system will have some amount of subjective judgement in naming and placing them—like, we have ElmsHouse and not NewBarkElmsHouse, or FastShip and not SSAqua—but (at least in my experience) it's easier for a new user to learn which maps are where by just scrolling through a flat list of them all and noticing what names there are. Clicking in and out of a lot of subdirectories makes that harder.
- The possible "areas" besides landmarks, like groups or environments, are unhelpful when looking for a map. Route7SaffronGate is in MapGroup_Celadon, BattleTower1F is in MapGroup_Cianwood, etc.

I agree that the current system has some hard-to-guess names, but the simplest solution is to pick better names. `CianwoodCityManiasHouse`, `NewBarkPlayersHouse`, `AzaleaCharcoalKiln`, etc. (I've gotten used to important named characters having maps named after them, like `BluesHouse` or `KarensRoom`, but prefixing them would be fine too.)

IMO you're right about maps/ being important enough for a top-level directory.

yenatch:
if we grouped maps into subdirs doing it by landmark makes the most sense (plus shared/). most of the problems grouping maps go away automatically. a map can only have one landmark, and it's usually the name the player remembers.

underground path isn't ambiguous, and it's not shared. it has its own landmark.

one-map folders is not a big issue. many routes have houses in them. if you want to add a house to a route you don't have to think about where to put it. if 300 .asm files in one folder is acceptable to you, 100 landmark folders should be ok. it's not going to be a surprise to open route_1/ and see the only map in there is route 1 (and it will still have two files). i agree that we shouldn't make exceptions. especially in the case of routes it defeats the purpose. the idea is to group maps logically, not have as many in one folder as possible.

immediately knowing what assets are shared is an advantage, not a drawback

finding a specific map is one thing, but the goal is also to make finding related maps and adding/editing/sharing maps easier. if you want to share a single map, you still can. that process doesnt change. if you want to share a lot of maps it's easier to pick and choose certain locations so you don't forget any maps by mistake.

usually when the existing tools (file managers) suck we try to make up for it. fortunately you can always search for file extension easily so i can't think of any cases where we need a separate dir for each filetype. it makes it harder to add/edit/share maps too

roukaour:
> it's not going to be a surprise to open route_1/ and see the only map in there is route 1 (and it will still have two files).

It's not surprising, but it's also not useful. With all maps in one directory, you can scroll down the list and make changes like correcting `Unknown` labels, or checking which sprites are used; and you can immediately see how many total maps there are. Subdirectories have neither advantage.

> the goal is also to make finding related maps and adding/editing/sharing maps easier. if you want to share a single map, you still can. that process doesnt change. if you want to share a lot of maps it's easier to pick and choose certain locations so you don't forget any maps by mistake.

The filename prefixes are already accomplishing this 90% of the time, and we can just fix the remainder (`BluesHouse`, `CharcoalKiln`, etc).

yenatch:
10% is a lot of edge cases. but obviously some maps should be renamed either way

for your example you can search *.asm in explorer (which you would already have to do in a flat dir)

also you should split the data/map_objects.asm changes out of this pr

mid-kid:
>With all maps in one directory, you can scroll down the list and make changes like correcting Unknown labels,
`grep -r Unknown`

>or checking which sprites are used;
`grep -rw SPRITE_LASS`

>and you can immediately see how many total maps there are.
`find -name "*.asm" | wc -l`

All of these are non-issues, really.

yenatch:
I think renaming the map files is going too far. We should stick with what we have now if that's the only way to keep them.

roukaour:
Sure, you can search for "\*.asm", or `grep maps/**/*.asm`; but how often would you benefit from the subdirectories, instead of working around them?

Their main benefit is that it keeps related blk+asm files together while avoiding a huge directory; but plenty of related files are separated anyway (like tilesets/ and gfx/tilesets/, or maps/ and constants/map_constants.asm which contains their dimensions). It's not like you can send someone the asm+blk files and have everything they need to use the map. And GitHub solves sharing problems anyway.

I count 38 maps that could use landmark prefixes: <https://pastebin.com/6cT2QiHr> This would be in keeping with the current convention; e.g. it's already `CianwoodCityPhotoStudio`, not just `PhotoStudio`. After that the only "hard-to-find" maps would be the three second-floor Pokémon Center ones, and even they could get "Pokecenter2F" as a prefix.

How about we just keep maps/ as-is for now, and consistently apply the landmark prefixes? Then wait and see if subdirectories would be more helpful, or if blk/asm separation is worth doing.

mid-kid:
>but how often would you benefit from the subdirectories, instead of working around them?


Often enough. Searching for stuff is the worst. I find myself constantly doing stuff like `$ cat maps/GoldenrodCity<tab><tab>` just to see everything in GoldenrodCity. Being able to just open a folder to see everything would help a lot. Likewise, it's _very_ helpful to be able to see which maps are reused in multiple locations.

>It's not like you can send someone the asm+blk files and have everything they need to use the map. And GitHub solves sharing problems anyway.

A lot of people don't know how to use git. Teaching them is much harder than telling them to zip up the files and send them my way.

I'm fine with fixing the landmark prefixes first, but this is something we'll have to figure out in the end regardless.

roukaour:
Would regional `johto` and `kanto` subdirectories be sufficient? (Plus `special` for the PC2F maps, and `unused` for the beta ones.) (Fast Ship shows up on Johto's map, so it would be in `maps/johto/`.) That would avoid having many subdirs with a few files each, and the filename prefixes would not have redundant information. And the max dir size would be roughly halved.

I'd suggest per-environment subdirs, since there are five of those, but splitting up landmarks would be annoying. (`maps/city/Violet.asm` plus `maps/indoor/VioletMart.asm` plus `maps/dungeon/SproutTower1F.asm`... no.)

mid-kid:
I find just using the region names as subdirectories to be a fine enough compromise, if we can't go all the way. I have the same issue with splitting them in per-environment subdirs as I have with splitting them in `blk` and `scripts`, though.

surskitty:
I like either per-landmark subdirs or subdirs for landmarks with multiple maps, so that routes without houses don't get subdir'd.

roukaour:
Looks like subdirectories will be per-landmark if they do happen.

PR #436, opened 2017-12-26
Made Get*Name functions use wNamedObjectIndexBuffer

mid-kid:
This seems like the appropriate name for these instances.
Also renamed all the easy-to-spot instances of wd265 where this is
applicable.

This one is going to need a slight more bit of review and thinking:
* Have I replaced all instances correctly?
* Is the name of this wram address even appropriate? It's used for storing IDs of things (moves, 'mon, tms/hms, types, etc), mostly, as well as some results of a few calculations.


yenatch:
`wNameIndex`

mid-kid:
It's not used only for names either.
It's often just used for temporary ID storage. Much like CurPartySpecies.

PR #437, opened 2017-12-26
Reorganize battle/; rename most "header" misnomers; gfx/pics → gfx/pokemon

roukaour:
I left `map_header` and `map_header_2` alone because (1) `map_data` and `map_data_2` sound generic, (2) pokered also calls them headers, and (3) technically `*_MapScriptHeader` and `*_MapEventHeader` aren't headers either. (`MapEventHeader` is more like `MapScriptFooter`.) Plus the fact that `MapEventHeader` (and `musicheader`s) come right before their data is convention/coincidence; they only need to be in the same bank. audio/cries.asm's `musicheader`s are all grouped together, separate from all the cry channel data. Anyway, updating the map-related labels can be done along with refactoring the `maps/` directory structure, however we decide to do that.


wram.asm
```diff
@@ -2941,21 +2941,16 @@ wBattleAnimTemp0:: db
...
-wBattleAnimTemp4:: db
-wBattleAnimTemp5:: db
-wBattleAnimTemp6:: db
-wBattleAnimTemp7:: db
-wBattleAnimTemp8:: db
 
 NEXTU ; d419
 wBattleAnimTempOAMFlags:: db
-	ds 1
+wBattleAnimTempField02:: db
```
yenatch: 02?

engine/mail.asm
```diff
@@ -397,11 +397,11 @@ MailboxPC: ; 0x44806
 
 	ld a, [wCurMessageIndex]
 	ld [wMenuCursorBuffer], a
-	ld a, [OBPals + 8 * 6]
+	ld a, [OBPals palette 6]
 	ld [wMenuScrollPosition], a
 	call ScrollingMenu
 	ld a, [wMenuScrollPosition]
-	ld [OBPals + 8 * 6], a
+	ld [OBPals palette 6], a
```
yenatch: these seem wrong

maps/AzaleaGym.asm
```diff
@@ -91,35 +91,35 @@ TrainerTwinsAmyandmay2:
 	closetext
 	end
 
-TrainerBug_catcherbenny:
-	trainer EVENT_BEAT_BUG_CATCHER_BENNY, BUG_CATCHER, BUG_CATCHER_BENNY, Bug_catcherbennySeenText, Bug_catcherbennyBeatenText, 0, .AfterScript
+TrainerBugCatcherbenny:
+	trainer EVENT_BEAT_BUG_CATCHER_BENNY, BUG_CATCHER, BUG_CATCHER_BENNY, BugCatcherbennySeenText, BugCatcherbennyBeatenText, 0, .AfterScript
```
yenatch: benny

engine/mail.asm
roukaour: Can you please elaborate? `make compare` is OK, and that's been a common pattern for loading Nth palettes.

yenatch: Why would the menu scroll position be the first byte of the palette?

wram.asm
roukaour: It corresponds to `BATTLEANIMSTRUCT_02`.

PR #443, opened 2017-12-28
Misc style cleanup and documentation

xCrystal:
- Use v prefix on VRAM addresses
- Add w prefix to a few WRAM addresses (ones from engine_flags)
- Uppercase MACRO, ENDM, and EQUS
- More Magikarp length documentation
- More slots constants and documentation


xCrystal:
Should wram addresses in banks 2 to 7 be labeled as wN_Address, or just wAddress? I assume it's the former but both styles are used right now.
`w5_MobileOpponentBattleMessages`
`wScratchTileMap` (bank 6)

aaaaaa123456789:
My $0.37, which comes down to 2 cents USD after converting:

* There's no reason to write explicit address components in labels. That includes bank numbers; `BANK()` is a thing. While it might be useful to separate between WRA1 and WRA2-7 labels, explicit bank numbers are a bad idea.
* The above point is made stronger by the fact that WRA0 and WRA1 labels, while being conceptually different, carry the same prefix.
* Address comments are of little use in general, considering the symbol file exists. I'd advocate for getting rid of all of them, and WRAM seems like a good place to start.

roukaour:
Explicit bank values `w5_MobileOpponentBattleMessages` always looked to me like leftovers that were never eliminated because nobody knows/cares about mobile code.

If address comments are going away, then can the .sym files be added to version control? That way you don't have to successfully build an unmodified repo to see where things are. The original sym files are useful for debugging shifted data, or finding GameShark codes.

aaaaaa123456789:
+1 to tracking .sym files

Make sure the file is sorted in some useful way before tracking it, though. RGBDS is free to emit symbols in any order it wishes (and it makes extensive use of that liberty), so an unsorted .sym file will end up causing artificial changes as the repo changes.

xCrystal:
Yeah, I also do like the idea of having the .sym file readily available. Much better than looking at the address comments in wram anyway. I don't care if those are removed, and I do find them useless for the most part, although I can imagine some poor souls that are still making Pokemon hacks by hex editing looking at them somehow.

xCrystal:
All the ``wReel1Whatever - wReel1`` offsets could use constants but for now I'm done with this PR.

constants/map_constants.asm
```diff
@@ -1,9 +1,9 @@
-newgroup: macro
+newgroup: MACRO
 const_value = const_value + 1
 	enum_start 1
 endm
```
yenatch: `endm` should match `macro` case

xCrystal: my bad, forgot those.

xCrystal: wen't ahead and removed random indents too #50fc9c3

roukaour:
Should `IF`/`ELIF`/`ENDC` and `REPT`/`ENDR` also be capitalized?

yenatch:
don't think it's necessary

Issue #441, opened 2017-12-28
Reorganize maps/


roukaour:
Subdirectories within `maps/` for each landmark, including `special/`, plus `unused/` for the beta maps. (Maybe?) See PR #446 .

Rangi42:
With landmark subdirectories, "<Landmark><Purpose>House" maps can be made less redundant. E.g. instead of "maps/goldenrod_city/GoldenrodBillsHouse.asm", "maps/goldenrod_city/BillsFamilyHouse.asm" (since "BillsHouse.asm" is the one on Route 25). Or instead of "maps/route_26/Route26DayofWeekSiblingsHouse.asm", just "maps/route_26/WeekSiblingsHouse.asm". Full list: https://pastebin.com/b0KmqRUj

Rangi42:
Both of the map reorganization PRs were closed, so this looks like an abandoned idea.

PR #446, opened 2017-12-29
group maps into directories by landmark

yenatch:
adding another map reorg to the pile


roukaour:
(Supersedes #386 .)

roukaour:
These maps can be renamed for consistency, regardless of whether this exact subdirectory PR is merged:

- `CianwoodCityPhotoStudio` → CianwoodPhotoStudio`
- `LavenderTownSpeechHouse` → LavenderSpeechHouse`
- `Route5SaffronCityGate` → Route5SaffronGate`
- `Route1718Gate` → Route17_18Gate`
- `Route2946Gate` → Route29_46Gate`
- `BetaTeakCity.blk` → BetaEcruteakCity.blk`
- `BetaSSAquaInsideCutOut` → `BetaFastShipInsideCutOut`
- `BetaCinnabarIslandPokemonLabHallway.blk → `BetaCinnabarPokemonLabHallway.blk`
- `BetaCinnabarIslandPokemonLabRoom1.blk → `BetaCinnabarPokemonLabRoom1.blk`
- `BetaCinnabarIslandPokemonLabRoom2.blk → `BetaCinnabarPokemonLabRoom2.blk`
- `BetaCinnabarIslandPokemonLabRoom3.blk → `BetaCinnabarPokemonLabRoom3.blk`

(Renaming the two-route gates makes them sort correctly when using a "natural sort" like File Explorer, and IMO is easier to read. But if underscores are too exceptional, then never mind them.)

TheMostCuriousThing:
I don't remember the consensus on this change (except that there more or less wasn't one). I'm ambivalent on the reorganization itself, but I *really really* appreciate `blockdata.asm`, `map_headers.asm`, `second_map_headers.asm` (whatever [their ultimate name](https://github.com/pret/pokecrystal/issues/441)) getting their own directory level.

roukaour:
I would kind of like, as an immediate partial fix, to move the "metadata" files into data/, for both maps/ and tilesets/.

This would mean:

* data/maps/map_headers.asm
* data/maps/map_headers_2.asm (from maps/second_map_headers.asm)
* data/maps/block_data.asm
* data/tilesets/tilesets.asm
* data/tilesets/data.asm
* data/tilesets/palette_maps.asm
* engine/routines/LoadMapGroupRoof.asm (from tilesets/roofs.asm)
* engine/tileset_palettes.asm (from tilesets/special_palettes.asm)
* engine/tileset_anims.asm (from tilesets/animations.asm)

That way, maps/ and tilesets/ would both be just catch-all directories for the not-yet-organized map and tileset data, and we can at least make the metadata easy to edit until the data's final organization is determined.

PR #449, opened 2017-12-29
Misc labeling and cleaning up

xCrystal:


constants/hardware_constants.asm
```diff
@@ -114,10 +114,16 @@ rWave_d     EQU $ff3d
...
+rLCDC_SPRITE_ENABLE  EQU 1
+rLCDC_SPRITE_SIZE    EQU 2
+rLCDC_WINDOW_TILEMAP EQU 6
+rLCDC_ENABLE         EQU 7
+rLCDC_DEFAULT        EQU %11100011
 rSTAT       EQU $ff41 ; LCDC Status (R/W)
 rSCY        EQU $ff42 ; Scroll Y (R/W)
 rSCX        EQU $ff43 ; Scroll X (R/W)
 rLY         EQU $ff44 ; LCDC Y-Coordinate (R)
+LY_VBLANK   EQU 144
```
roukaour: not `rLY_VBLANK`?

aaaaaa123456789: It's a constant, not a register. `rFOO` is used for the actual I/O registers.

engine/intro_menu.asm
```diff
@@ -1035,7 +1035,7 @@ StartTitleScreen: ; 6219
 	ld [rSVBK], a
 
 	ld hl, rLCDC
-	res 2, [hl]
+	res rLCDC_SPRITE_SIZE, [hl] ; 8x8
```
roukaour: The 8x8/8x16 meaning would also be useful next to the `rLCDC_SPRITE_SIZE` definition.

constants/hardware_constants.asm
xCrystal: yeah either the placement or prefix seems wrong

home/lcd.asm
```diff
@@ -58,11 +58,11 @@ DisableLCD:: ; 568
 .wait
 ; Wait until VBlank would normally happen
 	ld a, [rLY]
-	cp 145
+	cp LY_VBLANK + 1
 	jr nz, .wait
 
 	ld a, [rLCDC]
-	and %01111111 ; lcd enable off
+	and $ff ^ 1 << rLCDC_ENABLE
```
roukaour: `$ff ^ (1 << rLCDC_ENABLE)` is how similar lines are formatted.

constants/hardware_constants.asm
```diff
@@ -114,10 +114,16 @@ rWave_d     EQU $ff3d
 rWave_e     EQU $ff3e
 rWave_f     EQU $ff3f
 rLCDC       EQU $ff40 ; LCD Control (R/W)
+rLCDC_SPRITE_ENABLE  EQU 1
+rLCDC_SPRITE_SIZE    EQU 2
+rLCDC_WINDOW_TILEMAP EQU 6
+rLCDC_ENABLE         EQU 7
+rLCDC_DEFAULT        EQU %11100011
```
aaaaaa123456789: Should probably be `LCDC_DEFAULT`, for the same reason as `LY_VBLANK` below.

home/lcd.asm
aaaaaa123456789: Risking enraging the macro haters, I'm tempted to do `NOT EQUS "$ff ^ ", since `~` cannot be used. (Or just do the right thing and report the bug to rgbds.)

xCrystal:
Of course, I wrote "BP Map"

PR #450, opened 2017-12-29
Constants for TextBoxFlags

roukaour:
From home.asm:

```
PrintLetterDelay:: ; 313d
; Wait before printing the next letter.

; The text speed setting in Options is actually a frame count:
; 	fast: 1 frame
; 	mid:  3 frames
; 	slow: 5 frames

; TextBoxFlags[!0] and A or B override text speed with a one-frame delay.
; Options[4] and TextBoxFlags[!1] disable the delay.
```

### DIFF
--- a/constants/wram_constants.asm
+++ b/constants/wram_constants.asm
@@ -31,8 +31,8 @@ TEXT_DELAY_SLOW EQU 5
 
 ; TextBoxFlags:
 	const_def
-	const TEXT_DELAY_F    ; 0
-	const NO_TEXT_DELAY_F ; 1
+	const FAST_TEXT_DELAY_F ; 0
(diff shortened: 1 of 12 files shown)


aaaaaa123456789:
Should probably have better names than `TEXT_DELAY_F` and `NO_TEXT_DELAY_F`. The names imply that they are opposite, which makes one wonder how it would be possible for none or both to be set at the same time.

xCrystal:
``ENABLE_TEXT_SPEED`` (bit 0) and ``DISABLE_TEXT_DELAY`` (bit 1) ?

roukaour:
I'm done here unless anything else needs correcting.

PR #451, opened 2017-12-29
Fix and document some battle animation macros

yenatch:


roukaour:
Since you have a PR open already: `ld a, $6 ; OBJ 6` in engine/sprites.asm:AnimateEndOfExpBar should be `ld a, PAL_BATTLE_OB_BLUE`.

PR #455, opened 2018-01-01
Use more palette(s) constants, rename BG/OBP wram addresses, and rBGPI+rOBPI auto-increment constants (from Rangi)

xCrystal:
Merged two commits from roukaour/master to avoid late conflicts as we were dealing with some similar things, so this effectively makes #454 redundant.

The reason to rename the four WRAM palette buffers is that their usage appears to be just for redundancy, so it makes more sense to just name them 1 and 2.

`wBGPals1 palette PAL_BG_YELLOW + 2`
Expressions like this one can be changed to e.g. `wBGPals1 palette PAL_BG_YELLOW color 1` but maybe that's overkill?

Merge whenever.

### DIFF
--- a/mobile/mobile_42.asm
+++ b/mobile/mobile_42.asm
@@ -836,7 +836,7 @@ MobileTradeAnim_02: ; 108638
 	ld a, $5
 	ld [rSVBK], a
 	ld hl, Palette_109107
-	ld de, UnknBGPals
+	ld de, wBGPals1
 	ld bc, 8 palettes
 	call CopyBytes
 	pop af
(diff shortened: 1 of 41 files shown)


mobile/mobile_42.asm
```diff
@@ -1485,7 +1485,7 @@ Function108b45: ; 108b45
 	ld a, $5
 	ld [rSVBK], a
 	ld de, (31 << 10) + (31 << 5) + 31 ; $7fff
```
roukaour: This could use palred/palgreen/palblue. Likewise for line 1503 below.

PR #456, opened 2018-01-01
OAM data documentation; move more content into data/; move palettes into gfx/; more code+script constants; consistent map naming

roukaour:
Also merges:

- #452 Menu work
- #457 sort the symfile by symbol type
- #459 Consistent unused content references, and predef and special label names
- #460 Replace hardcoded banks & addresses in non-mobile code


roukaour:
I'm done with this PR.

roukaour:
Okay, I'm actually done with this PR.

roukaour:
Next change will be documenting battle anim constants, but that'll take time and be a separate PR, so merge this whenever.

yenatch:
please put map2sym.py in another pr and rename includes.asm to constants.asm

xCrystal:
I don't mind text/ being moved into data/, but I think all the text should remain inside a unique text/ folder.  Personally, I don't mind "deep" directory trees (3? 4? levels) nearly as much as I dislike directories with loads of files and folders all over the place. It's about organization and being able to locate content as easily as possible by navegating through directories using common sense. Whether something like phone_text.asm would belong more to a data/phone/ directory than a data/text/ directory is arguable and up to preference, but I don't think this approach or scattering all text files inside an already dense directory is a good idea; I think we should be going into the opposite direction.

roukaour:
I'm OK with relocating the text files; will wait to see what @yenatch says.

home/names.asm
```diff
@@ -7,7 +7,7 @@ NamesPointers:: ; 33ab
 	dbw 0, PartyMonOT
 	dbw 0, OTPartyMonOT
 	dba TrainerClassNames
-	dbw $4, $4b52 ; within PackMenuGFX
+	dbw $4, PackMenuGFX + 60
```
yenatch: This comment was just to indicate that it wasn't a valid address. It's not actually referencing PackMenuGFX. My guess would be it's a pointer to MoveDescriptions (0b:4b52). The `$4, $4b52` is `$4, $4000` in Gold. The address of MoveDescriptions in Gold is also at 0x4000 (6d:4000). The bank is wrong still, but 04:4000 in Gold was GetPlayerMovement, so the bank was probably wrong to begin with.

engine/battle/core.asm
```diff
@@ -26,7 +26,7 @@ DoBattle: ; 3c000
 	and a
 	jr z, .not_linked
 
-	ld a, [hLinkPlayerNumber]
+	ld a, [hSerialConnectionStatus]
 	cp $2
```
xCrystal: Some hSerialConnectionStatus constants (like USING_INTERNAL_CLOCK) are missing in code, more specifically in engine/battle/core.asm, home/init.asm, mobile/mobile_40.asm, and a few in engine/link.asm. Pointing out in case you haven't noticed, because you may already be aware.

roukaour: Thanks. Even with that the link code has much left to do (e.g. the `wPlayerLinkAction` and `wOtherPlayerLinkMode` values, the many `wXXXX` labels, hardcoded sizes like `$1b9`), but I don't have the knowledge of GB serial hardware to do it.

engine/pokedex.asm
```diff
@@ -15,6 +15,9 @@
 	const DEXSTATE_UPDATE_UNOWN_MODE
 	const DEXSTATE_EXIT
 
+POKDEX_SCX EQU 5
```
yenatch: pokdex?

engine/types.asm
```diff
@@ -91,9 +91,32 @@ GetTypeName: ; 50964
...
-	ld bc, 13
+	ld bc, MOVE_NAME_LENGTH
 	jp CopyBytes
 ; 5097b
 
 
 INCLUDE "data/types/names.asm"
+
+
+Unreferenced_GetGen1TrainerClassName: ; 50a28
```
yenatch: what? why is this here?

Rangi42: Should it go in engine/routines/unreferenced_getgen1trainerclassname.asm?

wram.asm
```diff
@@ -1565,7 +1565,7 @@ wDaysSince:: db
 
 SECTION "WRAM 1", WRAMX
 
-wd000:: ds 1
+wDecompressBuffer:: ds 1
```
yenatch: should this be a union?

Rangi42: No, there are no uses of unidentified `wd000` left.

yenatch: pretty sure wDecompressBuffer is not 1 byte large

Rangi42: Okay, I see. You could, but there's a big union right after it at d002, and every NEXTU would need a `ds 2` at the beginning. How about a comment noting that `wDecompressBuffer` is larger than that? (Max size: `GBCOnlyGFX` takes $540 bytes uncompressed, and ~~I think `LZ_170d16` takes $c10 in mobile_5c.asm~~. Oh, mobile_5c.asm should have used `wDecompressScratch`, not `wDecompressBuffer`. So yeah, $540 bytes, and `wGBCOnlyDecompressBuffer` would be a more accurate name.)

charmap.asm
```diff
@@ -207,17 +208,29 @@
...
+	charmap "<TTE>",     $1e ; "って"
+	charmap "<O>",       $1e ; "を "
+	charmap "<TA!>",     $22 ; "た！"
+	charmap "<KOUGEKI>", $23 ; "こうげき"
+	charmap "<WA>",      $24 ; "は　"
+	charmap "<NO>",      $25 ; "の　"
+	charmap "<ROUTE>",   $35 ; "ばん　どうろ"
+	charmap "<WATASHI>", $36 ; "わたし"
+	charmap "<KOKO_WA>", $37 ; "ここは"
+	charmap "<GA>",      $4a ; "が "
```
yenatch: I have a feeling these are a `'d` situation. It's fine to leave them for now but they were probably invisible.

engine/link_2.asm
```diff
@@ -56,7 +56,7 @@ LinkTextbox2: ; 4d35b
 
 .PlaceBorder: ; 4d37e
 	push hl
-	ld a, $76
+	ld a, "ぁ" ; $76
```
yenatch: this doesnt seem like a border tile

engine/link_trade.asm
```diff
@@ -139,7 +139,7 @@ _LoadTradeScreenBorder: ; 16d696
 
 LinkComms_LoadPleaseWaitTextboxBorderGFX: ; 16d69a
 	ld de, LinkCommsBorderGFX + $30 tiles
-	ld hl, vTiles2 tile $76
+	ld hl, vTiles2 tile "ぁ"
```
yenatch: this doesnt seem right either

charmap.asm
Rangi42: They're dummied out in the English version, but home/text.asm and a couple of control character tables still mention them, so it's either this or `JP_ROUTE_CHAR EQU $35`.

engine/link_2.asm
Rangi42: Yeah, I undid that later.

charmap.asm
yenatch: What I mean is this:
```asm
    charmap "ばん　どうろ", $35
```

Rangi42: ? `charmap "ばん　どうろ", $35` doesn't exist anywhere. The `"ばん　どうろ"` is just a comment explaining what the control character was for.

PR #457, opened 2018-01-02
sort the symfile by symbol type

yenatch:


roukaour:
map2sym.py in #456 can do this without edge cases, and with extra data in comments (bank and section metadata, and free space). Just needs to be incorporated into the Makefile. (Supports Python 2 or 3.)

PR #458, opened 2018-01-02
Use macro for all occurrences of box address tables.

TheMostCuriousThing:
The exact same address table appears six times and should be consolidated:

```
	dba sBox1
	dba sBox2
	dba sBox3
	dba sBox4
	dba sBox5
	dba sBox6
	dba sBox7
	dba sBox8
	dba sBox9
	dba sBox10
	dba sBox11
	dba sBox12
	dba sBox13
	dba sBox14
```

I have no idea where the macro should go, though.

### DIFF
--- a/engine/billspc.asm
+++ b/engine/billspc.asm
@@ -2216,21 +2216,7 @@ GetBoxPointer: ; e3396 (38:7396)
 ; e33a6 (38:73a6)
 
 .boxes ; e33a6
-	;  bank, address
-	dba sBox1
-	dba sBox2
-	dba sBox3
-	dba sBox4
-	dba sBox5
-	dba sBox6
-	dba sBox7
-	dba sBox8
-	dba sBox9
-	dba sBox10
-	dba sBox11
-	dba sBox12
-	dba sBox13
-	dba sBox14
+	box_address_table
 ; e33d0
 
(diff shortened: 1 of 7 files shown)


xCrystal:
``box_pointer_table`` is probably more accurate? Might as well consolidate all the local label names to something like ``.BoxPointers:`` or whichever name you like.

aaaaaa123456789:
Probably a bad idea.

First of all, replacing data by a macro to avoid repetition is a sure way of going towards macro hell; it won't be the only case of repeated data. (I'm aware that `dba` is already a macro, but it doesn't conceal the data; it just shows its structure.)

Moreover, this is a data structure that is easily computed at runtime, and a useful thing to do is to replace it entirely by a pointer calculation.

That being said, adding a new box does require modifying multiple pointer tables, assuming they are kept — but it also requires modifying multiple functions and _other_ correlated data structures, and no macro will save a ROM hacker from that. The issue of possibly missing something is better solved by documenting how to add/remove boxes, just like other "easy" modifications (e.g., how to replace a Pokémon) are (or can be) documented.

TheMostCuriousThing:
The complexity of adding a new box is the reason I'm here. I'll look at how the process can be documented; that's a good idea.

Issue #453, opened 2017-12-30
Remove the `mobile` string equate


yenatch:
`mobile` was used for stubbed mobile functions.

```asm
Example: mobile
    ...
```

What it should look like instead:
```asm
Example_Stubbed:
    ret
    ...
```

mid-kid:
I'd prefer a prefix over a suffix.

yenatch:
yeah, do whatever makes sense

xCrystal:
There are three different cases to address: Unreferenced code, single ret functions that are still referenced (these are rare, but I've seen a few at some point), and the kind of stubbed functions mentioned above that may or may not be referenced. I'd probably handle the last two cases in a similar way (e.g. ``Stubbed_`` or ``CommentedOut_`` prefix/suffix), while for unreferenced functions that don't begin with ret I'd use ``Unreferenced_``.

Currently, there appears to be a mix of ``mobile``, ``; XXX`` and ``; unreferenced`` used to address any of those three cases indiscriminately.

xCrystal:
``Label: mobile`` -> ``Stubbed_Label\n ret``.
``Label: ; unreferenced``, ``Label: ; Unreferenced`` and ``Label: ; XXX`` -> ``Unreferenced_Label`` where Label is ``Function1a2b3c`` or ``Data4d5e6f`` except maybe specific cases like the unreferenced predef ``Predef22``.
``; unreferenced``, ``; Unreferenced`` and ``XXX`` without label -> Same as above (create label).
``Unused_Label`` -> Same as above (change ``Unused_`` to ``Unreferenced_``).

Does this look good?

yenatch:
Many unused labels are referenced. Some examples:
```
AI_Smart_Unused2B
CheckUnusedTwoDayTimer
CreditsUnusedCD21
ElmPhoneUnusedText
EnemyMonUnused
ForUnusedCursor
MapRoute34UnusedSignpostText
ReelAction_Unused
Route32CooltrainerMText_UnusedSproutTower
Special_CheckUnusedTwoDayTimer
sTrainerRankingUnused1
sTrainerRankingUnused2
sTrainerRankingUnused3
UnusedGuySpriteGFX
UnusedPhoneScript
UnusedPhoneText
UnusedSfx_Ch5
UnusedSfx_Ch6
UnusedSfx_Ch7
UnusedSfx_Ch8
UnusedTitleBG_GFX
UnusedTitleBG_Palettes
[...]

PR #448, opened 2017-12-29
Move tilesets/ into data/

roukaour:
I'd actually prefer for `tilesets/` to stay as a top-level directory if `maps/` is too; but if it does get moved, here's the PR.

PR #459, opened 2018-01-02
Consistent unused content references, and predef and special label names

xCrystal:
Ready to merge. However, it's going to conflict somewhere with #456, so feel free to merge that one first and then I could look at merging and resolving conflicts. There *shouldn't* be too many though.

Addresses issue #453. Gets rid of the ``mobile`` string, consolidates unused content with ``Unreferenced_`` or ``Stubbed_`` label prefixes, or with an ``; unused`` comment. These are mostly consistent and should be easy to locate with a simple search, even though words like ``unused`` and ``dummy`` are still scattered in many labels and comments of unused content, particularly for constants, text and scripts.

As for ``predef`` and ``specials``, I added ``Predef_`` and ``Special_`` and a prefix in function labels (for consistency with most current specials), using also the word ``Dummy`` or ``Unused`` when appropriate. Unecessary bank 0 specials were not renamed and instead commented with ``; bank 0``.

Also took out leftover unused data from main.asm and changed some ``UnknownText_xxx`` that are unreferenced to ``UnusedText_xxx``.

### DIFF
--- a/engine/battle/ai/move.asm
+++ b/engine/battle/ai/move.asm
@@ -92,7 +92,7 @@ AIChooseMove: ; 440ce
 
 	push bc
 	ld d, BANK(TrainerClassAttributes)
-	predef Predef_Flag
+	predef Predef_FlagAction
 	ld d, c
 	pop bc
 
--- a/engine/battle/ai/move.asm
+++ b/engine/battle/ai/move.asm
@@ -92,7 +92,7 @@ AIChooseMove: ; 440ce
 
 	push bc
 	ld d, BANK(TrainerClassAttributes)
-	predef Predef_FlagAction
+	predef Predef_SmallFarFlagAction
 	ld d, c
 	pop bc
 
--- a/engine/battle/ai/move.asm
+++ b/engine/battle/ai/move.asm
@@ -92,7 +92,7 @@ AIChooseMove: ; 440ce
 
 	push bc
 	ld d, BANK(TrainerClassAttributes)
-	predef FlagPredef
+	predef Predef_Flag
 	ld d, c
 	pop bc
 
--- a/engine/card_flip.asm
+++ b/engine/card_flip.asm
@@ -2,6 +2,13 @@ CARDFLIP_LIGHT_OFF EQU $ef
 CARDFLIP_LIGHT_ON  EQU $f5
 CARDFLIP_DECK_SIZE EQU 4 * 6
 
+; two labels below called from inside ./dummy_game.asm
+Unknown_e00ed: ; e00ed (38:40ed)
+; Graphics for an unused Game Corner
+; game were meant to be here.
+Ret_e00ed: ; e00ed (38:40ed)
+	ret
+
 _CardFlip: ; e00ee (38:40ee)
 	ld hl, Options
 	set 4, [hl]
--- a/engine/card_flip.asm
+++ b/engine/card_flip.asm
@@ -6,7 +6,7 @@ CARDFLIP_DECK_SIZE EQU 4 * 6
 Unknown_e00ed: ; e00ed (38:40ed)
 ; Graphics for an unused Game Corner
 ; game were meant to be here.
-Ret_e00ed: ; e00ed (38:40ed)
+ret_e00ed: ; e00ed (38:40ed)
 	ret
 
 _CardFlip: ; e00ee (38:40ee)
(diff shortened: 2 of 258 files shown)


xCrystal:
I'm done with this PR

engine/card_flip.asm
```diff
@@ -2,6 +2,13 @@ CARDFLIP_LIGHT_OFF EQU $ef
 CARDFLIP_LIGHT_ON  EQU $f5
 CARDFLIP_DECK_SIZE EQU 4 * 6
 
+; two labels below called from inside ./dummy_game.asm
+Unknown_e00ed: ; e00ed (38:40ed)
+; Graphics for an unused Game Corner
+; game were meant to be here.
+Ret_e00ed: ; e00ed (38:40ed)
```
roukaour: What about `Stubbed_` or `Dummy_` here?

roukaour: Also, the name of `Unknown_e00ed` could reflect that it was some 4-tile image for the dummy game. (Most likely a cursor, given that `SPRITE_ANIM_INDEX_DUMMY_GAME` uses 4 tiles at $00 just like other cursors.)

xCrystal: I left empty functions with generic names, and only used meaningful names when there is actual code (even if unreferenced or dummied out after a ret) to back it up. If this one's so evident though, I guess it could be changed to something to reflect what it was supposed to be there, but I feel it's enough if reflected in a comment.

maps/CopycatsHouse2F.asm
```diff
@@ -70,7 +70,7 @@ Copycat:
 .Default_Merge_3a:
 	faceplayer
 	variablesprite SPRITE_COPYCAT, SPRITE_LASS
-	special MapCallbackSprites_LoadUsedSpritesGFX
+	special Special_MapCallbackSprites_LoadUsedSpritesGFX
```
roukaour: Can this have a shorter name, like `Special_LoadUsedSpritesGFX`?

xCrystal: Sure thing, will do it tomorrow along with anything else that might come up. Assuming #456 gets merged first, I'll also try to do the merge myself

xCrystal: 91f97f08

engine/battle/ai/move.asm
```diff
@@ -92,7 +92,7 @@ AIChooseMove: ; 440ce
 
 	push bc
 	ld d, BANK(TrainerClassAttributes)
-	predef FlagPredef
+	predef Predef_Flag
```
roukaour: `FlagActionPredef` would be more clear, and matches `FlagAction` and `EventFlagAction`.

engine/routines/flagpredef.asm
```diff
@@ -1,5 +1,5 @@
-Predef_FlagAction: ; 4d7c1
-; Perform action b on flag c in flag array hl.
+Predef_SmallFarFlagAction: ; 4d7c1
```
roukaour: This is not always far, only when d = 1.

xCrystal: I know... it's kinda like the FarCopyBytes kinda functions. You can use them to copy a byte from the current bank. I know, there's always a bankswitch in those regardless, but you are still forced to set d to the bank argument here as well (even if it's just setting it to 0 to avoid bankswitch).
I mean, I don't really care. I can rename it again if wanted.

roukaour: Eh, it's fine. Come to think of it, `FlagAction` does *not* support bank switching, so "SmallFar" accurately describes this one's differences.

I would, however, either rename the file to predef_smallfarflagaction.asm (ugly) or find a better place for it than engine/routines/*lowercaselabelname*.asm (which was never meant to be permanent anyway).

xCrystal: Yes, the file should be renamed (along with many others), but I don't think it's a good idea to do more in this PR until the current open ones are merged. I know it's just a single file that won't cause any conflict, but it would probably become renamed/moved again when that stuff is dealt with in another PR anyway.

roukaour:
The conflict with #456 is very minor: my copy of data/sprite_anims/framesets.asm subsumes your changes; and there's a one-line manual merge needed in engine/item_effects.asm.

xCrystal:
Good to hear, feel free to merge this PR into your repo, as I'm not going to push any more commits to it.

As far as I know, #386 is obsolete, #446 and #448 aren't settled yet, #457 is not a problem, and #452 isn't finished yet. If #460 is finished too, it could be merged as well to get a head start fixing potential conflicts.

roukaour:
Yeah, #456, #457, #459, #460, and #469 can be merged, and #386 can be closed.

Issue #413, opened 2017-12-11
Many banks are still hard-coded


roukaour:
Some have comments which ought to have been used in the first place, e.g. `ld a, $5 ; BANK(UnknBGPals)`. Others are not commented, usually when they're loaded into `[rSVBK]`. Using relevant `BANK()` directives will make it easier for projects to move WRAM around, such as for allowing larger sprite animations or improving the map engine.

mid-kid:
As an add-on to this: There's still a few hardcoded memory addresses hanging around, mostly pointing into WRAM. These should be assgned labels as well.

mid-kid:
A lot of these were fixed in #460. What is the current status of this? Is there anything missing? If it's done, this issue should be closed.

Rangi42:
Yeah, might as well close this. The remaining hard-coded banks are part of the larger problems of (a) hard-coded numbers in less-well-disassembled engine code, and (b) barely-disassembled-at-all mobile code.

PR #460, opened 2018-01-03
Replace hardcoded banks & addresses in non-mobile code

TheMostCuriousThing:
Addresses [#413](https://github.com/pret/pokecrystal/issues/413).

"Replace" is defined here as replacing constant numbers with `BANK` functions (in the case of banks) or labels (in the case of addresses), *except* where the use of a bank or address appears unnecessary and therefore can't be clearly labeled (such cases should be explicitly commented, though).

"Non-mobile code" is defined here as all code outside `mobile/*.asm`, `lib/mobile/main.asm` , and `home/mobile.asm`. (There are still thousands of hardcoded banks & addresses in mobile code.)

- ~~Replace hardcoded ROM banks.~~ (Possibly unneeded?)
- [x] Replace hardcoded SRAM banks.
- [x] Replace hardcoded WRAM banks.
- [x] Replace hardcoded ROM addresses.
- [x] Replace hardcoded SRAM addresses.
- [x] Replace hardcoded WRAM addresses.
- ~~Replace hardcoded VRAM banks & addresses.~~ (This isn't advisable while VRAM is still [vaguely and incompletely labeled](https://github.com/pret/pokecrystal/issues/466)).

([fb825a8](https://github.com/pret/pokecrystal/commit/fb825a800e44c0795225949658b58a0522a9c998) is irrelevant and a relic I committed before I knew the scope of this PR. But I left it in since it just makes one minor improvement to mobile code.)

### DIFF
--- a/engine/gbc_only.asm
+++ b/engine/gbc_only.asm
@@ -10,7 +10,7 @@ GBCOnlyScreen: ; 4ea82
 	call ClearTileMap
 
 	ld hl, GBCOnlyGFX
-	ld de, $d000
+	ld de, WRAM1_Begin
 	ld a, [rSVBK]
 	push af
 	ld a, 0
@@ -19,7 +19,7 @@ GBCOnlyScreen: ; 4ea82
 	pop af
 	ld [rSVBK], a
 
-	ld de, $d000
+	ld de, WRAM1_Begin
 	ld hl, vTiles2
 	lb bc, BANK(GBCOnlyGFX), $54
 	call Get2bpp
--- a/engine/gbc_only.asm
+++ b/engine/gbc_only.asm
@@ -10,7 +10,7 @@ GBCOnlyScreen: ; 4ea82
 	call ClearTileMap
 
 	ld hl, GBCOnlyGFX
-	ld de, WRAM1_Begin
+	ld de, wd000
 	ld a, [rSVBK]
 	push af
 	ld a, 0
@@ -19,7 +19,7 @@ GBCOnlyScreen: ; 4ea82
 	pop af
 	ld [rSVBK], a
 
-	ld de, WRAM1_Begin
+	ld de, wd000
 	ld hl, vTiles2
 	lb bc, BANK(GBCOnlyGFX), $54
 	call Get2bpp
--- a/engine/gbc_only.asm
+++ b/engine/gbc_only.asm
@@ -13,7 +13,7 @@ GBCOnlyScreen: ; 4ea82
 	ld de, wd000
 	ld a, [rSVBK]
 	push af
-	ld a, 0
+	ld a, 0 ; this is tantamount to selecting Bank 1
 	ld [rSVBK], a
 	call Decompress
 	pop af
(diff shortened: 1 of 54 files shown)


home/names.asm
```diff
@@ -7,7 +7,7 @@ NamesPointers:: ; 33ab
 	dbw 0, PartyMonOT
 	dbw 0, OTPartyMonOT
 	dba TrainerClassNames
-	dbw $4, $4b52 ; within PackMenuGFX
+	dbw $4, PackMenuGFX + 60
```
roukaour: Make that `dba PackMenuGFX + 60` since `$4` is `BANK(PackMenuGFX + 60)`.

TheMostCuriousThing: `BANK()` can't evaluate arithmetic (at least I think that's the problem). `dbw BANK(PackMenuGFX), PackMenuGFX + 60` is the best that can be done.

roukaour: Oh well, `dbw BANK(PackMenuGFX), PackMenuGFX + 60` is good enough.

mobile/mobile_5f.asm
```diff
@@ -3035,86 +3035,55 @@ Function17e349: ; 17e349
...
-	call Function17e41e
+inc_crash_check_pointer_farcall: MACRO
+	call IncCrashCheckPointer
+	call HlToCrashCheckPointer ; redundant
 	ld a, [rSVBK]
 	push af
 	ld a, $1
 	ld [rSVBK], a
-	farcall SaveGameData_
+REPT _NARG
```
xCrystal: ``rept``, ``shift``, and ``endr`` should be lowercase!

engine/gbc_only.asm
```diff
@@ -13,7 +13,7 @@ GBCOnlyScreen: ; 4ea82
 	ld de, wd000
 	ld a, [rSVBK]
 	push af
-	ld a, 0
+	ld a, 0 ; this is tantamount to selecting Bank 1
```
roukaour: Please clarify that this isn't a typo; writing 0 to rSVBK has the same effect as 1. ([SVBK](http://gbdev.gg8.se/files/docs/mirrors/pandocs.html#videodisplay))

xCrystal:
While at it, there may be more hardcoded WRAM banks in calls to ``FarCopyWRAM`` and ``GetFarWRAMByte`` (``GetFarWRAMWord`` appears to be unreferenced).

engine/menu.asm
```diff
@@ -730,15 +730,15 @@ Function24423: ; 24423
 	ld a, [VramState]
 	bit 0, a
 	ret z
-	xor a
+	xor a ; effectively ld a, BANK(sScratch)
```
roukaour: Other parts of the code just comment the implied zero value: `xor a ; PARTYMON`, `xor a ; OAKS_POKEMON_TALK`, `xor a ; LOW(vBGMap0)`, etc.

TheMostCuriousThing:
I'm currently tracing backward from `MBC3RomBank` and don't think there are any hardcoded ROM banks left *from what I'm seeing*. If anyone has any leads, please share, since I'm about done this otherwise.

PR #469, opened 2018-01-12
Label frameset 09

mid-kid:
It seems to be used as the "big" text entry cursor (i.e. for the bottom
3 buttons in the text entry screen).
It is never referenced explicitly, instead used when adding 1 to
SPRITEANIMSTRUCT_0E, in NamingScreen_AnimateCursor.


engine/namingscreen.asm
```diff
@@ -562,10 +562,10 @@ NamingScreen_AnimateCursor: ; 11a3b (4:5a3b)
 .ok
 	cp d
 	ld de, .LetterEntries
-	ld a, $0
+	ld a, $0 ; SPRITE_ANIM_FRAMESET_TEXT_ENTRY_CURSOR
 	jr nz, .ok2
 	ld de, .CaseDelEnd
-	ld a, $1
+	ld a, $1 ; SPRITE_ANIM_FRAMESET_TEXT_ENTRY_CURSOR_BIG
```
roukaour: `ld a, SPRITE_ANIM_FRAMESET_TEXT_ENTRY_CURSOR_BIG - SPRITE_ANIM_FRAMESET_TEXT_ENTRY_CURSOR ; 1`

PR #475, opened 2018-01-16
A few map labels 

surskitty:


maps/Route44.asm
```diff
@@ -26,29 +26,29 @@ TrainerBirdKeeperVance1:
 	end_if_just_battled
 	opentext
 	checkflag ENGINE_VANCE
-	iftrue UnknownScript_0x19d86a
+	iftrue VanceWantsBattle
 	checkcellnum PHONE_BIRDKEEPER_VANCE
-	iftrue UnknownScript_0x19d8eb
+	iftrue NumberAcceptedM
```
roukaour: Don't use such generic labels; if Bird Keeper Vance has `NumberAcceptedM`, what will the many other phone-call trainers use? *Local* generic names are OK, like `.AskedAlready`, since they can be reused.

surskitty: The other phone call trainers use the exact same thing; all NumberAcceptedM does is `scall numberacceptedm`.  It doesn't build with simply .NumberAcceptedM because it's referenced by both Vance and Wilton.

roukaour:
No other map script uses "Rt" to abbreviate "Route"; and if you make labels local with a ".", you don't need the location prefix anyway. Look at the labels used for other phone-call trainers, like Joey on Route 30 or Ralph and Liz on Route 32, and try to stay consistent with them.

surskitty:
Merged in again ....

yenatch:
Sorry. I was gonna do it at some point.

Issue #440, opened 2017-12-28
Rename and document menus


yenatch:
- `MenuDataHeader` -> `Menu`
- `MenuData2` -> `MenuItems`
- figure out the purpose of each flag
- possibly use a macro to define menus
- docs/menus.md (see scripting compendium)
- `loadmenudata` -> `loadmenu`?

roukaour:
With a name as short and generic as `Menu`, they'll also need more meaningful names. Like `.MenuDataHeader_Water` could be `.UnownWaterMenu`, but plenty of plain `.MenuDataHeaders` are left.

mid-kid:
wMenuData2Flags (scrolling):
```
7: Select is functional
6: Start is functional
5: Run function 3 in the structure
4: Show arrows on the right-hand side
3: D-Left is functional
2: D-Right is functional
1: ?????
0: ?????
```

roukaour:
`MenuDataHeader` is a header the same way `musicheader` is: it appears right before the actual data and contains some metadata. Simply renaming `MenuData2` to `MenuData` would make this clear, since there'd commonly be `MenuDataHeader`s containing a pointer to the subsequent `.MenuData`.

Rangi42:
Renaming `MenuDataHeader` to just `Menu` would get confusing because `ThingMenu` is sometimes already a different Thing-menu-related concept, like a routine to draw the menu (example: `BattleMenu` and `BattleMenuDataHeader`).

Rangi42:
Is anything left to be done regarding menus, or can this be closed?

mid-kid:
Don't think so. Time to close.

PR #476, opened 2018-01-22
Resolve some issues; rename some maps; move more tables into data/; warp_def → warp_event (ready to merge)

### LINKED ISSUE #429: Use constants for bitfields
The `bit`, `res`, and `set` instructions are frequently used with hard-coded arguments and comments explaining their meaning (or without even the comments).

Rangi42:
Issues resolved:

- #429 Use constants for engine flags' bit fields
- #440 Rename `MenuDataHeader` and `MenuItems2`
- #478 Remove `Special_` prefix from special routines
- #481 `EVENT_GAVE_KURT_APRICORNS` and the `EVENT_DRAGON_SHRINE_QUESTION_2` are misnamed


data/special_pointers.asm
```diff
@@ -133,7 +133,7 @@ SpecialsPointers:: ; c029
 	add_special Function1700ba
 	add_special Function170114
 	add_special BattleTowerBattle
-	add_special UnusedSpecial_Function1704e1
+	add_special Function1704e1
```
xCrystal: Why did you remove ``Unused`` here?

Anyway wouldn't it be preferable if unused specials and predefs kept the ``Special_`` and ``Predef_`` keyword? Given that they are never referenced they won't look redundant anywhere and you can notice which special/prefef slots can be reused more "transparently". Not a big deal though.

home/battle.asm
```diff
@@ -136,100 +136,7 @@ UpdateBattleHuds:: ; 39d4
...
-	dw wPlayerMoveStructPower,       wEnemyMoveStructPower
-	dw wPlayerMoveStructType,        wEnemyMoveStructType
-	dw CurPlayerMove,                CurEnemyMove
-	dw LastPlayerCounterMove,        LastEnemyCounterMove
-	dw LastPlayerMove,               LastEnemyMove
-; 3a90
+INCLUDE "home/battle_vars.asm"
 
 
 FarCopyRadioText:: ; 3a90
```
xCrystal: ``FarCopyRadioText`` should probably be moved elsewhere too

Rangi42: I'm pretty sure the home bank is full of one-off "misplaced" routines like that.

Rangi42:
@yenatch This PR is my last reorganizing one; what's the status on reviewing + merging its changes? (Likewise for #477 and #486.) I know @mid-kid is waiting on this PR to do another one that turns `mapconst name, height, width` into the more consistent `map_const name, width, height`. And anyone forking prior to this will have to merge from upstream for tutorials and docs to match their code.

mid-kid:
I'm not waiting. I've already added it to my PR. I'll have to rebase my entire PR when this is merged anyway.

yenatch:
i'm up to date. here are my notes

fa3192e3da7e3ce9550b40a2ab2c3640c996ea46 rept 4
ead103d2c6d21cd4774e5c32b9c66f9435f7adbb loadmenuheader -> loadmenu
132fe4692581891802c38420bdd4744d58bf78d5 greaterthanthatlevel -> abovelevel
87514598bedf66be222d85ebbc7565c14d79e0f4 engine/battle/effect_commands $ffff -> MAX_DAMAGE
006d9f644168a09ba147643e29a599bc668bf094 if theres BIGDOLLSYM and BIGDOLLASYM, what is BIGDOLL
0eeb00d6ac67199b3343c5a6ddc739b251f7a68f indicate water_1-3 meaning (amphibian, fish, invertebrate), humanlike -> humanshape, undiscovered -> none
b2970e75f4d1a7a2bc7ac58a656ffd350a80cf0a makes it harder to add more scenes
c834bdadf2f3e38d684f798dfa67d7cd751a6a65 objectface -> faceobject
bf138d94d42ba0d092d5028e40a85260c250a4ee scene_var isnt related to loadvar

not sure the best way to merge this with #477 and #486

Rangi42:
Thanks for reviewing these changes! Re: #477 and #486, if you merge them first, I'll merge them into my fork.

fa3192e, 132fe46, 0eeb00d: Fixed.

ead103d: @mid-kid is familiar with the menu code+data and pointed out that this really is a header, just like music headers: it's metadata that comes right before the actual menu data. Renaming every instance of "MenuHeader" to just "Menu" also turns up name collisions where there's already a "FooMenu" defined (e.g. `BattleMenu`). Anyway, can we 

8751459: A problem with `MAX_DAMAGE EQU $ffff` is that frequently maximum damage is loaded as "ld a, $ff; ld [hli], a; ld [hl], a" or similar (see `.CriticalMultiplier`, `.finish_ohko`, `DoubleDamage`, etc). So somebody altering `MAX_DAMAGE` wouldn't affect all those sites.

006d9f6: The only thing that generically applies to either kind of big doll is `OBJECT_ACTION_BIG_DOLL` (see `SetFacingBigDoll` in engine/map_object_action.asm; it checks the sprite ID and uses `FACING_BIG_DOLL_SYM` or `FACING_BIG_DOLL_ASYM` accordingly). All other big doll related constants have been renamed to make it clear which kind they apply to, including the ones that had overly-specific Snorlax- or Lapras-related names.

b2970e7: No more than the `MAPNAME_OBJECTNAME` constants make it harder to add new `object_event`s. If a map has the typical pattern of "scene 0 autoruns when you first enter, it sets the scene to 1, scene 1 does nothing", then `SCENE_DEFAULT` and `SCENE_FINISHED` exist for that common case. If not, you can choose to define meaningful constants, or use hard-coded ID numbers. They're mostly helpful for maps with three or more scenes, and maps whose scenes get modified by other maps (e.g. in CianwoodCity.asm, `setmapscene ROUTE_42, SCENE_ROUTE42_SUICUNE` is immediately obvious).

c834bda: `faceobject` is already defined. There's `objectface OBJECT, DIR` and `faceobject OBJECT, OTHEROBJECT`. I just changed "person" to "object" in them, but would welcome a less confusing pair of names.

bf138d9: With `map_def` renamed to `map` and `warp_def` to `warp_event`, I felt that `scene_def` was an outlier. It is kind of related to `loadvar` since `loadvar` takes a WRAM variable as an argument, and so does `scene_var`; pretty sure you could even use `loadvar wMyMapSceneID, N` in place of `setscene MY_MAP, N`. But, is there a name you'd prefer besides `scene_var` or `scene_def`?

aaaaaa123456789:
Re: `MAX_DAMAGE` (and anything similar), it's a fixable issue:

```
  ld a, low(MAX_DAMAGE)
  ld [hli], a
if low(MAX_DAMAGE) == high(MAX_DAMAGE)
  ld [hl], a
else
  ld [hl], high(MAX_DAMAGE)
endc
```

Rangi42:
If something like this is acceptable, I'll commit it:

```
	ld a, LOW(MAX_DAMAGE)
	ld [de], a
	inc de
if HIGH(MAX_DAMAGE) != LOW(MAX_DAMAGE)
	ld a, HIGH(MAX_DAMAGE)
endc
	ld [de], a
```

```
; Cap at MAX_DAMAGE.
	ret nc

	ld a, LOW(MAX_DAMAGE)
	ld [hProduct + 2], a
if HIGH(MAX_DAMAGE) != LOW(MAX_DAMAGE)
	ld a, HIGH(MAX_DAMAGE)
endc
	ld [hProduct + 3], a

	ret
```

yenatch:
ead103d2c6d21cd4774e5c32b9c66f9435f7adbb no problem calling them headers, but the command should still be `loadmenu`
006d9f644168a09ba147643e29a599bc668bf094 there's also`SPRITEMOVEDATA_BIGDOLL`, but i guess it acts the same way
c834bdadf2f3e38d684f798dfa67d7cd751a6a65 dunno, maybe `turnobject`
bf138d94d42ba0d092d5028e40a85260c250a4ee the name is fine but the commit message implied it was related to `loadvar`

Rangi42:
@yenatch Alright, ready to merge. :)

@SatoMew "Barrier Station" was brought up initially and rejected. And the "Beta" suffix is useful because those maps *are* beta, inaccessible leftovers.

SatoMew:
@Rangi42 I see. And I'd go with "unused" instead as it is a more accurate term but maybe it's not too important.

PR #477, opened 2018-01-23
Medium-sized cleanups

mid-kid:
Built files, pet-peeves and shenanigans.


engine/move_mon.asm
```diff
@@ -201,24 +223,37 @@ endr
...
 	xor a
+	; Status
 	ld [de], a
 	inc de
+	; Unused
 	ld [de], a
 	inc de
+
+	; Initialize HP.
 	ld bc, 10
```
xCrystal: This ``ld bc, 10`` can probably use some constant? (not sure if there are more of these hardcoded values around, but it would be neat to fix it/them like you've done with a few others)

mid-kid: I'm having trouble tracing the origins and current value of `hl`, but if I were to know that it'd not be much more than `structure_target - structure_base`.

PR #484, opened 2018-01-25
Added support for building a Virtual Console patch

mid-kid:
Further described in the commit messages themselves.

The program I've made for this is far from perfect. It won't check for additional differences between the two ROMs other than the ones described in the patch template, and won't check if the offsets of the labels in the original and patched rom actually match.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,6 +9,7 @@ RGBFIX := rgbfix
 RGBGFX := rgbgfx
 RGBLINK := rgblink
 
+patches := pokecrystalvc.patch
 roms := pokecrystal.gbc pokecrystal11.gbc
 
 crystal_obj := \
@@ -28,6 +29,7 @@ gfx/sprites.o \
 lib/mobile/main.o
 
 crystal11_obj := $(crystal_obj:.o=11.o)
+crystalvc_obj := $(crystal_obj:.o=vc.o)
 
 
 ### Build targets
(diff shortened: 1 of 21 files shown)


yenatch:
I don't think this should be closed. It's interesting and probably useful. It's a lot of bloat for a niche feature, so I don't think it should be merged either. I get that it becomes even less useful by not being part of everyone's repo even if they aren't aware it's there. I guess it's not too much work to remove, you'd just kill all the .VC labels and _CRYSTALVC ifdefs. I'm still undecided.

mid-kid:
>I'm still undecided.

Thank you for answering. For me, there's either one of two possible outcomes to this: It's either merged, or I make a .patch for this in a separate repo. 158 added lines in the main repo is too small to warrant the maintenance burden of a fork/branch, and the difficulty it brings for including it in on-going repos (especially since there's too many people who don't know git). I am not bothered about discoverability, as it could be linked to in the wiki.

>The question is whether it will have to be maintained by the user.

The idea is that it shouldn't. If it does introduce any problems when building anything unrelated to the patch file, it's a bug, plain and simple. I am willing to continue developing and documenting the utility as long as necessary. I've yet to find out what most of the `ConditionValue`s mean, and not all of the format is properly documented.

I am (and have been made) perfectly aware that my solution is far from perfect. The fact I'm trying to generate a byte-perfect patch file brings with it a lot of baggage, such as a largely unreadable patch template and a format that largely sucks. It can be partially mitigated by, for example, moving problematic lines to different files, but that doesn't completely solve the problem. We could have a tool to "clean it up", or provide a "clean" file, but I'm not sure if that really solves the problem either. There's also some improvements that could be done to the template format, as well as removing the necessity of having 7 FPA\_\*\_Begin labels by providing a more generic name in the template format.

Would any of these be good improvements and/or make you more comfortable merging this?

Makefile
```diff
@@ -41,20 +43,25 @@ crystal11_obj := $(crystal_obj:.o=11.o)
 all: crystal
 crystal: pokecrystal.gbc
 crystal11: pokecrystal11.gbc
+crystalvc: pokecrystalvc.patch
```
yenatch:
```makefile
crystalvc: pokecrystalvc.gbc pokecrystalvc.patch
```

mid-kid: That's implied by `pokecrystalvc.patch: pokecrystalvc.gbc` later in the file, but ok.

mid-kid:
I'm going to be trying to improve make_patch.c in the next few days, to try to make it more flexible and hopefully future-proof. (Mostly in relation to the HRAM PR https://github.com/pret/pokecrystal/pull/515)
I'll re-open this when it's done.

PR #489, opened 2018-02-19
Add Fix for Lock-On Bug

i0brendan0:
These changes should fix the status move and secondary effects not working if the opponent is flying or digging but has been locked on.

### DIFF
--- a/docs/bugs_and_glitches.md
+++ b/docs/bugs_and_glitches.md
@@ -304,6 +304,7 @@ Fix:
 CheckHiddenOpponent: ; 37daa
 	ld a, BATTLE_VARS_SUBSTATUS5_OPP
 	call GetBattleVar
+	ld a, [hl]
 	and 1 << SUBSTATUS_LOCK_ON
 	cp 1 << SUBSTATUS_LOCK_ON
 	ret z
--- a/docs/bugs_and_glitches.md
+++ b/docs/bugs_and_glitches.md
@@ -302,17 +302,10 @@ Fix:
 
 ```asm
 CheckHiddenOpponent: ; 37daa
-	ld a, BATTLE_VARS_SUBSTATUS5_OPP
-	call GetBattleVar
-	ld a, [hl]
-	and 1 << SUBSTATUS_LOCK_ON
-	cp 1 << SUBSTATUS_LOCK_ON
-	ret z
-	ld a, BATTLE_VARS_SUBSTATUS3_OPP
-	call GetBattleVar
-	and 1 << SUBSTATUS_FLYING | 1 << SUBSTATUS_UNDERGROUND
 	ret
 ```
+
+The code in `CheckHiddenOpponent` is completely redundant as `CheckHit` already does what this code is doing. Another option is to remove `CheckHiddenOpponent` completely, the calls to `CheckHiddenOpponent`, and the jump afterwards.
 
 ## Beat Up can desynchronize link battles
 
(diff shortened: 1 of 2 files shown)


xCrystal:
I'm a bit confused. Shouldn't all moves miss against a target using Fly or Dig regardless of Lock-On/Mind Reader? Just by logic I mean.

i0brendan0:
No. Lock On bypasses Fly and Dig. But the developers didn't add any code to make status moves and secondary effects bypass with Lock On.

Rangi42:
Four lines of original code is a lot to leave commented in the .asm itself. Usually fixes are kept in the doc .md; in-source fix comments are limited to single lines that are obvious mistakes, like swapping c/nc or moving a line up/down a bit.

i0brendan0:
I removed the commented code as you asked.

Rangi42:
Thanks; now the doc code just has to match the asm code.

```asm
CheckHiddenOpponent: ; 37daa
; BUG: This routine should account for Lock-On and Mind Reader.
	ld a, BATTLE_VARS_SUBSTATUS3_OPP
	call GetBattleVar
	and 1 << SUBSTATUS_FLYING | 1 << SUBSTATUS_UNDERGROUND
	ret
```

**Fix:**

```asm
CheckHiddenOpponent: ; 37daa
	ld a, BATTLE_VARS_SUBSTATUS5_OPP
	call GetBattleVar
	and 1 << SUBSTATUS_LOCK_ON
	ret z
	ld a, BATTLE_VARS_SUBSTATUS3_OPP
	call GetBattleVar
	and 1 << SUBSTATUS_FLYING | 1 << SUBSTATUS_UNDERGROUND
	ret
```

i0brendan0:
Added the Protect miss message bug and fix.

Rangi42:
None of the other bug fixes mention specific line numbers, because they can easily change when the file is updated. Please just put one code snippet showing the wrong code and a second one showing the fixed code, like the rest.

mid-kid:
Setting wEffectFailed simply isn't enough to fix that "bug".

Rangi42:
Can this be closed? It's a deeper issue with all failure text, and questionable whether it's a "bug" or just how Gen 2 was intended.

i0brendan0:
The failure text bug can be disregarded. It's not gamebreaking. But the Lock On bug should be tested and committed if it works. I can remove the failure commit if need be.

yenatch:
The lock on check seems backwards. It should return if it's 1, not 0.

Rangi42:
`and`+`cp` for a single bit isn't necessary; as you guessed before, you can just use `bit`, which also makes the intention clearer (testing whether a single bit is set).

i0brendan0:
I’m still fuzzy on the bit instruction. The command needs to return early with the z flag set if the opponent is under lock on. As it is right now it does that. But if I’m understanding the bit instruction correctly, bit clears the z flag if the given bit is set, which is not what I want to do.

docs/bugs_and_glitches.md
```diff
@@ -298,8 +298,20 @@ CheckHiddenOpponent: ; 37daa
...
 ```
 
-*To do:* Fix this bug.
+Fix:
 
+```asm
+CheckHiddenOpponent: ; 37daa
+	ld a, BATTLE_VARS_SUBSTATUS5_OPP
+	call GetBattleVar
+	and 1 << SUBSTATUS_LOCK_ON
```
mid-kid: If this works, please replace this with `cpl; bit SUBSTATUS_LOCK_ON, a`.

i0brendan0:
so turns out all this work to fix this bug was unnecessary as the entire section that introduces the bug is unnecessary.

mid-kid:
This is now ready for merge.

Rangi42:
You've verified that dummying out `CheckHiddenOpponent` lets Lock-On bypass Fly/Dig for those stat-lowering moves?

mid-kid:
Yeah, he has.

mid-kid:
To explain: Fly and Dig are already checked for in CheckHit, and there it happens properly, i.e. the move will always hit if locked on and won't if not and the target is flying/digging.

Rangi42:
As long as you've actually ran the fix and seen it happening properly, then that's fine.

PR #493, opened 2018-02-26
add the wiki as a submodule

yenatch:
is this a good idea


Pokechu22:
I don't think this is a good idea.  Among other reasons, the wiki submodule would need to manually be updated for each edit the wiki, which would be easy to forget to do.  Plus I don't see any real value in doing this.

Rangi42:
Anyone can view the wiki, and it's been linked in the README and FAQ. If someone wants to edit it, it's not complicated to check out separately, or use the online editor.

Issue #491, opened 2018-02-26
Rename OlivineCityBattleTowerSignText, UnknownText_0x1a8cba, BattleTowerOutsideText_BattleTower, BattleTowerOutsideText_UltimateChallenge


SatoMew:
Another part of the code that can be explained by the Japanese version.

`OlivineCityBattleTowerSignText` replaces `UnknownText_0x1a8cba` when the Mobile System GB features are available. Likewise, `BattleTowerOutsideText_BattleTower` and `BattleTowerOutsideText_DoorsClosed` are respectively replaced by `BattleTowerOutsideText_UltimateChallenge` and `BattleTowerOutsideText_DoorsOpen`.

The Battle Tower in Western Crystal is always already open so both door texts are unused and the "Opening Now!" line should be considered a localization oversight in a comment due to it not making sense outside Japanese Crystal (for comparison, HeartGold/SoulSilver's text for the sign, which is now on Route 40, simply says _Battle Frontier Ahead_).

https://github.com/pret/pokecrystal/blob/c60f1331fef22b886adf3472d2b4f348832cfaf7/maps/OlivineCity.asm#L281-L288
https://github.com/pret/pokecrystal/blob/c60f1331fef22b886adf3472d2b4f348832cfaf7/maps/BattleTowerOutside.asm#L109-L127

SatoMew:
Shouldn't `BattleTowerOutsideText_BattleTower` and `BattleTowerOutsideText_UltimateChallenge` be renamed? They are sign texts but their labels don't make this obvious. https://github.com/pret/pokecrystal/commit/711e2d6143d59093a763f9bbd4a527f349fdeff8 changed the label of an unrelated sign text that is explicitly named as such so it's a bit inconsistent.

Are you guys OK with `BattleTowerOutsideSignText_NotYetOpen` and `BattleTowerOutsideSignText`?

PR #495, opened 2018-02-26
Use commonly-used egg group names

mid-kid:
Changed some egg group names to fully match the names used by most of
the community, including Bulbapedia and the veekun dex.
These names are sourced from Stadium 2, Pokédex 3D, as well as strategy
guides, according to Bulbapedia, so it's about as close to "official" as
we can get them.
Sadly, some of these names don't properly describe what is contained in
them. As such, I've written a short description for those in a
comment, that should help clarify their contents. For more details,
see https://bulbapedia.bulbagarden.net/wiki/Egg_group#Egg_Groups.

Note: Also included a small commit for https://github.com/pret/pokecrystal/issues/491
Note2: More commits have been and/or will be added. See the commit messages for info.


iimarckus:
IMO it would be best to use the Stadium 2 names.

mid-kid:
@iimarckus Why? Because of them being from the same gen?

iimarckus:
Yes, exactly.

SatoMew:
Seems like https://github.com/pret/pokecrystal/pull/495/commits/ebfb3e776660b967b0dbc464a76fbea87f46e0c5 didn't catch everything (not including comments here).

If the Stadium 2 Egg Group names are used, then a comment explaining that certain Pokémon belong to the Gender unknown Egg Group in Stadium 2 might be useful.

@iimarckus Is [this list](https://bulbapedia.bulbagarden.net/wiki/Gender_unknown_(Egg_Group)#Pok.C3.A9mon) correct? I am unable to check but, if accurate and complete, then it's those particular 8 Pokémon from Generations I and II that have distinct Egg Groups between Gold/Silver/Crystal and Stadium 2.

mid-kid:
@SatoMew Why would you want a list of them? `grep -rw GENDER_UNKNOWN data/` should do more than enough to provide curious people with a list.

I want to get this PR out of the way. Some people want the stadium 2 names, other people prefer the modern names. I couldn't care either way, as long as we consistently use either convention. What will it be?

mid-kid:
Okay, I have a proposal:
I'm going to revert the `GENDERLESS -> GENDER_UNKNOWN` commit. There's no reason to call some 'mon GENDER_UNKNOWN when they're not in the stadium2 egg group, which is the origin of that term in the first place
Then, I'm going to change the egg group names to the stadium2 ones, because that's the ones that match up with the direct counterpart to this game that names them, and thus feel more accurate, except for gender unknown, because that one simply doesn't exist in this game.

If you have any improvements/objections to this it's time to make your voice heard.

Rangi42:
Stadium 2 names are fine with me, but put the alternate Pokédex 3D names in comments, and consider using `AMPHIBIAN`/`INVERTEBRATE`/`FISH` instead of `WATER 1`/`2`/`3` since base stats are clearer that way.

SatoMew:
As I previously stated, "gender unknown" is the official term used for Pokémon with no assigned gender. Stadium 2 so happens to include an Egg Group called Gender unknown which does not exist in the core games. It's simply a matter of the same term being used in different contexts but it is indeed the correct name for what we want on pokecrystal.

If the goal is to be consistent and use official naming wherever possible, then reverting the change doesn't exactly help us reaching it.

As to the list, I proposed it because sites like Bulbapedia use and even link to the disassembly as a source for some of the technical information and thus have no interest in setting up git and the repo. This way, people who come here for documentation will get the required information without much effort.

mid-kid:
@SatoMew there's no need to have a list for the gender unknown egg group when it doesn't exist in the game. Stadium 2 is the source.

mid-kid:
In the end, it was decided that GENDER_UNKNOWN should stay. Even if it doesn't have the same meaning as the stadium 2 egg group, it's used consistently across official documentation to refer to what would otherwise be called "genderless".

PR #496, opened 2018-02-26
Label more map scripts; document more bugs [ready to merge]

### LINKED ISSUE #494: Patch to Hall of Fame counter for entries no. 201 and higher
I'd like to add this to docs/bugs_and_glitches.md.

Due to a bug, the game will not display the intended message to players who enter the Hall of Fame at least 201 times and will keep repeating the 200th entry message ([video](https://www.youtube.com/watch?v=iHkWubvxmSg)).

https://github.com/pret/pokecrystal/blob/725148443f0ed0f70af747259ef49643359e92a2/engine/events/halloffame.asm#L373-L395

**Fix:** Replace `cp 200 + 1` with `cp 200`.

Rangi42:


SatoMew:
`MAX_STAT_VALUE` is a battle constant used by the Max Potion and Full Restore, which is equal to 999.

https://github.com/pret/pokecrystal/blob/725148443f0ed0f70af747259ef49643359e92a2/constants/battle_constants.asm#L72

https://github.com/pret/pokecrystal/blob/c60f1331fef22b886adf3472d2b4f348832cfaf7/data/items/heal_hp.asm#L8-L9

However, the game has no issue restoring full health to a Pokémon with a maximum HP value between 1000 and 65535. Shouldn't there be a cap for these two items somewhere?

Note that Red/Green/Blue/Yellow also behave like Gold/Silver/Crystal in this regard but there is no `MAX_STAT_VALUE` in the Generation I games.

Rangi42:
I'm not sure what you mean. The code had a 999 there, so I replaced it with a named constant.

SatoMew:
What I mean is that shouldn't the items stop restoring after 999 HP as the expectation is for the maximum HP value to never exceed 999? The game clearly proves otherwise but the information on Bulbapedia hints at there being a 999 HP cap so I'd like to know which one is right.

Rangi42:
It's possible that they ignore the value and have a separate code path which fully restores HP, even above 999.

SatoMew:
The Max Potion uses `RestoreHPEffect` (which calls `ItemRestoreHP`) while the Full Restore uses `FullRestoreEffect`.

https://github.com/pret/pokecrystal/blob/725148443f0ed0f70af747259ef49643359e92a2/engine/item_effects.asm#L31-L32

https://github.com/pret/pokecrystal/blob/725148443f0ed0f70af747259ef49643359e92a2/engine/item_effects.asm#L1639-L1676

Unlike the Full Restore, the Max Potion won't restore more than 999 HP at once so multiple Max Potions have to be used on a Pokémon with 65535 maximum HP but, like the Full Restore, it does not seem to check whether a Pokémon's maximum HP value is 999 and/or higher than 999.

Since I don't see `MAX_STAT_VALUE` at all in engine/item_effects.asm, does this mean that they ignore the value and, if so, is it intentional or a bug given what data/items/heal_hp.asm says? I feel a bit confused about what is supposed to happen.

xCrystal:
Just to point out a few unrelated random things that probably don't justify opening their own PR (also I have poketcg set up and I'm lazy to set up pokecrystal just for these things...).

data/phone/text/

- bill.asm: ``BillPhoneGeneriText`` -> ``BillPhoneGenericText``
- trainers1.asm -> trainers.asm
- It could be a good idea to move all *_overworld.asm files into their own overworld folder (and remove "_overworld" from their filenames.

home/rst.asm

- Space between comma and ROM0 in SECTIONs

data/wild/

- This is more of a suggestion, but maybe split in subfolders (grass, fish, tree, contest, swarm(?) ), and possibly do something about rock mons being blatantly conidering tree mons in filenames and labels (though I understand it's annoying because rock mons run off the same engine and pointer table as rock mons).

Rangi42:
data/wild/x/ would be a bit too much hierarchy for so few files. It's the maximum depth we have, and usually for grouping many files, like data/pokemon/base_stats or engine/battle/move_effects. data/wild's 15 files are simple enough on their own.

engine/events/halloffame.asm
```diff
@@ -376,7 +376,7 @@ _HallOfFamePC: ; 86650
 	pop hl
 	call DisplayHOFMon
 	ld a, [wHallOfFameTempWinCount]
-	cp 200 + 1
+	cp HOF_MASTER_COUNT
```
yenatch: I think this is backwards. The compare should be `HOF_MASTER_COUNT + 1` and the counter should stop at `HOF_MASTER_COUNT`.

Rangi42: I was uncertain about this too, but the video shows that the Japanese version displayed "200-Time Famer" and then "HOF Master" for 201 wins.

SatoMew: Before I patched the Japanese ROM, it'd display the equivalent to "200-Time Famer" every time for the 201th entry and later, not only the 200th. Changing `cp 200 + 1` there to `cp 200` fixed it for me.

SatoMew: Before I patched the Japanese ROM, it'd display the equivalent to "200-Time Famer" every time for the 201st entry and later, not only the 200th. Changing `cp 200 + 1` there to `cp 200` fixed it for me: if the counter is at 199, entering the Hall of Fame 200 times will update the counter to 200 and use the normal message and, once at 200, any new entry (201+) displays the special message.

Rangi42: Oh, I didn't realize the Japanese one had to be patched too. In that case, yeah, special-casing 200 makes more sense than 201. Thanks.

PR #509, opened 2018-05-05
A fix-various-little-things PR

mid-kid:


data/wild/flee_mons.asm
```diff
@@ -31,6 +31,5 @@ OftenFleeMons: ; 3c5a8
 AlwaysFleeMons: ; 3c5b1
 	db RAIKOU
 	db ENTEI
-;	db SUICUNE
```
kanzure: why remove suicune here?

mid-kid: All the other suicune comments had already been removed, since Suicune simply does not roam in Crystal. Just finishing the job here.

macros/scripts/events.asm
```diff
@@ -315,15 +315,15 @@ giveegg: MACRO
 	db \2 ; level
 ENDM
 
-	enum givepokeitem_command ; $2f
-givepokeitem: MACRO
-	db givepokeitem_command
+	enum givepokemail_command ; $2f
```
Rangi42: Please add `givepokeitem` and `checkpokeitem` to macros/legacy.asm since they're renamed. Also update docs/event_commands.md, and rename `Script_givepokeitem` and `Script_checkpokeitem` in engine/overworld/scripting.asm to match.

mid-kid: Fixed.

PR #510, opened 2018-05-05
Use explicit HELD_NONE and NO_LIMITS, not 0s; also fix #511

### LINKED ISSUE #511: "Left shift of negative value" warning with rgbds 0.3.7
```
warning: main.asm(312) -> engine/events/magnet_train.asm(7) -> lb(1):
    Left shift of negative value: -4
```

The relevant line:

```
    lb de, (11 * 8) - (11 * 8 + 4), -$60
````

Which macro expands to:

```
    ld de, ((11 * 8) - (11 * 8 + 4) & $ff) << 8 + (-$60 & $ff)
```

### PR DESCRIPTION
**Pull request by Rangi42**
`item_attribute` has the held effect, parameter, and property value together, and sometimes they're all 0. This makes it easier to tell which is which.

Used, for instance, here: <https://github.com/pret/pokecrystal/wiki/Add-different-kinds-of-new-items>

### CODE REVIEW

**Comment by yenatch** [comment]
There should be a similar constant for property too. Or even `CAN_SELECT` and `CAN_TOSS`.

**Comment by Rangi42** [comment]
If I add `CAN_SELECT` and `CAN_TOSS`, would it be too verbose to rename the existing ones `CANNOT` instead of `CANT`, so the difference is more obvious?

**Comment by Rangi42** [comment]
I went with `NO_LIMITS` over `CAN_SELECT` and `CAN_TOSS` because it gives shorter lines, and `CAN_SELECT | CAN_TOSS` would be a redundant `0 | 0`.

File: engine/events/magnet_train.asm [hunk 1]
```diff
@@ -4,13 +4,13 @@ MagnetTrain: ; 8cc04
...
 	lb bc,  $40,  $60
-	lb de, (11 * 8) - (11 * 8 + 4), -$60
+	lb de, ((11 * 8) - (11 * 8 + 4)), -$60
 	jr .continue
 
 .ToGoldenrod:
 	ld a, -1 ; backwards
 	lb bc, -$40, -$60
-	lb de, (11 * 8) + (11 * 8 + 4), $60
+	lb de, ((11 * 8) + (11 * 8 + 4)), $60
```
yenatch: this is a problem with the macro, not the usage

Rangi42: Should I bother undoing it, then?

PR #513, opened 2018-05-18
Add Counter and Mirror Coat Bugfix

hyperdriveguy:
A fix to check if the opponent's move did any damage. I've updated both the docs and added comments in the code.

### DIFF
--- a/docs/bugs_and_glitches.md
+++ b/docs/bugs_and_glitches.md
@@ -246,7 +246,35 @@ DefenseDownHit:
 
 ([Video](https://www.youtube.com/watch?v=uRYyzKRatFk))
 
-*To do:* Identify specific code causing this bug and fix it.
+This is a bug with `BattleCommand_Counter` in [engine/battle/move_effects/counter.asm](/engine/battle/move_effects/counter.asm) and `BattleCommand_MirrorCoat` in [engine/battle/move_effects/mirror_coat.asm](/engine/battle/move_effects/mirror_coat.asm):
+
+```asm
+	; BUG: Move should fail with all non-damaging battle actions
+	ld hl, wCurDamage
+	ld a, [hli]
+	or [hl]
+	ret z
+```
+
+**Fix:**
+
+```diff
+	ld hl, wCurDamage
+	ld a, [hli]
+	or [hl]
+-	ret z
++	jp z, .failed
+```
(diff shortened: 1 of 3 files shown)


Rangi42:
Have you actually tested this fix? You put the `jr z, .capped` line right after `ld hl, wCurDamage`, before the contents of `wCurDamage` have even been inspected.

I think you meant to replace `ret z` with `jr z, .capped`.

hyperdriveguy:
I swear I tested it, but you're probably right. I'll double check to see what works and update it.

hyperdriveguy:
After extensive fiddling and testing, I think I've found the proper fix.

docs/bugs_and_glitches.md
```diff
@@ -246,7 +246,34 @@ DefenseDownHit:
...
+	ld hl, wCurDamage
+	ld a, [hli]
+	or [hl]
+	jp z, .failed
+```
+
+Add this to the end of each file:
+
+```asm
+	.failed
```
mid-kid: This shouldn't be indented. Otherwise these fixes look OK.

docs/bugs_and_glitches.md
```diff
@@ -246,7 +246,34 @@ DefenseDownHit:
 
 ([Video](https://www.youtube.com/watch?v=uRYyzKRatFk))
 
-*To do:* Identify specific code causing this bug and fix it.
+This is a bug with `BattleCommand_Counter` in [engine/battle/move_effects/counter.asm](/engine/battle/move_effects/counter.asm) and `BattleCommand_MirrorCoat` in [engine/battle/move_effects/mirror_coat.asm](/engine/battle/move_effects/mirror_coat.asm):
+
+```asm
```
mid-kid: Oh, and this should be in diff format.

docs/bugs_and_glitches.md
hyperdriveguy: Fixed in latest commit.

docs/bugs_and_glitches.md
hyperdriveguy: Changed it to the diff format. But I think it should be noted that none of the other bugfixes use the diff format.

hyperdriveguy:
Ready to merge.

PR #514, opened 2018-05-18
[Ready to merge] Named constants for sprite movement data bit flags + other minor fixes

Rangi42:
Found while writing more tutorials.


engine/events/pokecenter_pc.asm
```diff
@@ -39,12 +39,19 @@ PokemonCenterPC: ; 1559a
 	dw PlaceNthMenuStrings
 	dw .JumpTable
 
+PCPC_PLAYERS_PC   EQU 0
```
xCrystal: ``PCPC``? Is this a typo?

Rangi42: "Pokémon Center PC".

xCrystal: My bad, thought it was some form of find/replace gone wrong.

engine/gfx/load_pics.asm
```diff
@@ -263,30 +263,30 @@ GLOBAL PICS_FIX
...
+	db BANK("Pics 1") + 14
+	db BANK("Pics 1") + 15
+	db BANK("Pics 1") + 16
+	db BANK("Pics 1") + 17
+	db BANK("Pics 1") + 18
+	db BANK("Pics 1") + 19
+	db BANK("Pics 1") + 20
+	db BANK("Pics 1") + 21
+	db BANK("Pics 1") + 22
+	db BANK("Pics 1") + 23
```
yenatch: This should probably be `BANK("Pics 2")` and so on. The remaining sections could be empty. That might also make it so this function doesn't have to be immediately killed to work with pics.

Rangi42: I did consider doing it that way from the start, but it seems like that would imply more flexibility than actually exists.

`FixPicBank` takes the "unfixed-bank" value from `dba_pic`, subtracts `BANK("Pics 1") - PICS_FIX`, and uses that as an index into `FixPicBank.PicsBanks`. If the array were `BANK("Pics 1")`, `BANK("Pics 2")`, etc, it would look like you can add `BANK("Pics N")` at the bottom and not worry about what N is; but in fact they'd still have to be contiguous since `N - (BANK("Pics 1") - PICS_FIX)` needs to be the corresponding array index.

(Also, the last five entries don't correspond to existing pic banks.)

Regardless, the whole thing is a design flaw no matter how it's written, so I've changed it. Just let me know if it should be different.

PR #516, opened 2018-06-03
Don’t use GNU extensions when calling sed

BenHetherington:
Apparently, GNU `sed` has a few extensions that aren’t supported by the version of BSD `sed` that currently comes with Macs.

This would cause `sort_symfile.sh` to fail on macOS, causing the build to appear to fail at the last minute.

Admittedly, I’m not very familiar with `sed`, but this seems to do the trick on both macOS and Ubuntu. I'm open to slicker solutions, if anyone more experienced wants to give this a shot.

- The input file must be last in the arguments list.
- The `-i` option, allowing the same file for input and output, doesn’t appear to be supported. Instead, I’m writing the output to a temporary file, and replacing the original file with that temporary file.
- Apparently `\w` isn’t supported, so I’m simply using `.` instead, as it appears to match “0_ROM0@” etc. just as well.

### DIFF
--- a/tools/sort_symfile.sh
+++ b/tools/sort_symfile.sh
@@ -8,6 +8,6 @@ sed \
     -e "s/^00:[C-D]/4_WRAM0@&/g" \
     -e "s/^..:[D-D]/5_WRAMX@&/g" \
     $1 \
-| sort -o $TEMP_FILE
-sed -e "s/^.*@//g" $TEMP_FILE > $1
-rm $TEMP_FILE
+| sort \
+| sed -e "s/^.*@//g" > $TEMP_FILE
+mv $TEMP_FILE $1


luckytyphlosion:
`mv: cannot stat ‘temp.sym’: No such file or directory`

mid-kid:
Writing a `temp.sym` might screw with a user's `temp.sym`. If you do so, please do it with an `$(mktemp)`'d file or something.

Ben10do:
@luckytyphlosion Thanks for pointing that the Travis CI failure, I guess I didn't test this with `make -j`… @mid-kid's suggestion would solve this, so I'll give that a go.

Ben10do:
This is now using a file created with `mktemp`.

Earlier, I somehow missed that it still kept the "0_ROM0@" prefixes when running it in macOS, so I've also updated the regular expression to address this.

PR #518, opened 2018-06-06
[RTM] Specify DVs individually

Rangi42:


data/trainers/dvs.asm
```diff
@@ -1,72 +1,76 @@
+dvs: MACRO
+; atk, def, spd, spc
+	dn \1, \2
+	dn \3, \4
+ENDM
```
yenatch: `dn` can take any number of arguments

PR #522, opened 2018-06-09
Convert code samples to diffs

ketsuban:
This affects both bugs_and_glitches.md and design_flaws.md.


Rangi42:
PR #513 was changed to use diff format, like the wiki tutorials do, so this should be merged for consistency.

PR #524, opened 2018-06-12
Provide fix for battle transition bug

ketsuban:
This is based on the fix committed (and, I assume, written) by roukaour for Polished Crystal, which can be found in two commits [here](https://github.com/Rangi42/polishedcrystal/commit/85ab1f875bd4ea9f473c5826a93601e9ce666b0c) and [here](https://github.com/Rangi42/polishedcrystal/commit/dcee95bdb7b559b1b9eef48b6647c56756bf43f1).

### DIFF
--- a/docs/bugs_and_glitches.md
+++ b/docs/bugs_and_glitches.md
@@ -936,7 +936,7 @@ This is a bug with `CalcMagikarpLength.BCLessThanDE` in [engine/events/magikarp.
 
 ([Video](https://www.youtube.com/watch?v=eij_1060SMc))
 
-This is a cluster of bugs centred on `StartTrainerBattle_DetermineWhichAnimation` in [engine/battle/battle_transition.asm](/engine/battle/battle_transition.asm).
+This is a cluster of bugs centered on `StartTrainerBattle_DetermineWhichAnimation` in [engine/battle/battle_transition.asm](/engine/battle/battle_transition.asm).
 
 **Fix:**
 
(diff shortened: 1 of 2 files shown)


Rangi42:
Thank you for taking the time to solve and document this bug! Comments:

- `StartTrainerBattle_DetermineWhichAnimation` is edited three different times. It would be less confusing as a single edit. You don't have to show the actual sequence of edits that were made while solving the bug, just your final product.
- "Let's change the hardcoded 6 to `PARTY_LENGTH`" should be applied to `FindFirstAliveMonAndStartBattle` in the original code.
- If `Function39990` needs deleting to make room in the bank, then show that in a diff. Or if it's harmless to just leave it there, then don't mention deleting it, just say to put `SetTrainerBattleLevel` at the bottom of the file.

ketsuban:
I felt it was important to make sure the reader understood what was going into this bugfix, since it's more extensive than the others on the page which mostly have a single change of only a few lines in one file. The tradeoff is the same file is altered multiple times.

Rangi42:
You're right, this isn't the usual one- or two-line change, so it's important to understand what's going on. But I think seeing the final changes would be easier to understand than being repeatedly told to go back and re-change a file. Comments are good for explaining each piece of the whole edit. Example: <https://pastebin.com/saQRUfWH> (And then after that something like "We used `farcall SetTrainerBattleLevel` but haven't defined that yet, so here it is, etc").

Rangi42:
Re: the way this fix is written: assuming it's based on the wiki tutorials' format of step-by-step instructions, only two places there involve re-editing a file: [once](https://github.com/pret/pokecrystal/wiki/Add-a-new-trainer-class#9-fix-bank-overflow-errors) to fix a bug that's only noticed when you try `make`, and [once](https://github.com/pret/pokecrystal/wiki/Add-different-kinds-of-new-items#pokeballeffect-dusk-ball) because the first edit was part of a summarized "add the essential data" step.

PR #534, opened 2018-06-25
[RTM] Fix triple newlines left over from removing address comments, and other improvements

Rangi42:
Example:

```
Function1337: ; 1337
    ret
; 1338

Function1338: ; 1338
    ...
```

became:

```
Function1337:
    ret


Function1338:
    ...
```

so this reformats it to:

```
Function1337:
    ret

Function1338:
    ...
```


mid-kid:
I actually removed the entire line if I removed the address comment on an otherwise blank line.
It's probable, however that things like
```
Function1337: ; 1337
    ret

; 1338

Function1338: ; 1338
    ...
```

became
```
Function1337: ; 1337
    ret


Function1338: ; 1338
    ...
```

I thought it was fine because there's a lot of other places where two newlines are used between functions for some reason.

engine/battle/ai/scoring.asm
```diff
@@ -232,7 +230,6 @@ AI_Types:
 	jr .checkmove
```
mid-kid: There's still three newlines here?

engine/battle/ai/scoring.asm
```diff
@@ -260,7 +257,6 @@ AI_Offensive:
...
```
mid-kid: Here as well? Not going to point all of them out, but there's more.

Rangi42:
I don't know why those were missed the first time, but now they're done.

engine/pokemon/correct_nick_errors.asm
```diff
@@ -70,5 +70,5 @@ CorrectNickErrors::
 	db "<ROUTE>",  "<GREEN>" + 1
 	db "<ENEMY>",  "<ENEMY>" + 1
 	db "<MOM>",    "<TM>"    + 1
-	db "<ROCKET>", "┘"       + 1
+	db "<ROCKET>", " "
```
yenatch: `" "` has no relation other than being one after. If border tiles don't count as control characters the problem is the labelling.

engine/games/card_flip.asm
```diff
@@ -1,5 +1,5 @@
-CARDFLIP_LIGHT_OFF EQU $ef
-CARDFLIP_LIGHT_ON  EQU $f5
+CARDFLIP_LIGHT_OFF EQUS "\"♂\"" ; $ef
+CARDFLIP_LIGHT_ON  EQUS "\"♀\"" ; $f5
```
yenatch: This seems like an assembler bug. `EQU "♂"` should be allowed.

Rangi42: True; but until then the escapes are necessary, and were already used in some other files.

engine/pokemon/correct_nick_errors.asm
Rangi42: I guess border tiles can count as control characters, then.

home/audio.asm
```diff
@@ -50,8 +50,6 @@ UpdateSound::
 
 _LoadMusicByte::
 ; wCurMusicByte = [a:de]
-GLOBAL LoadMusicByte
```
yenatch: This was here because `BANK()` used to not flag labels for import. This was fixed a while ago.

PR #546, opened 2018-07-27
Reword bug docs now that they use diffs

Rangi42:


docs/bugs_and_glitches.md
```diff
@@ -313,119 +311,105 @@ The code in `CheckHiddenOpponent` is completely redundant as `BattleCommand_Chec
...
+ 	ld a, BATTLE_VARS_SUBSTATUS3
+ 	call GetBattleVarAddr
+ 	res SUBSTATUS_IN_LOOP, [hl]
+ 	call BattleCommand_BeatUpFailText
 +	call BattleCommand_RaiseSub
-	jp EndMoveEffect
+ 	jp EndMoveEffect
 ```
 
+For the cosmetic fix, also edit `BattleCommand_FailureText` in [engine/battle/effect_commands.asm](/engine/battle/effect_commands.asm).
```
mid-kid: This is a completely separate fix, and can (and should) be used in combination with the "breaking" fix above.
Originally, it was stated as simply "Fix:", but in the last commit I thought it'd be nice to specify it won't break compatibility with an original game.

It is, in essence, a different bug in a different place, but it also fits with the "Beat Up may fail to raise substitute", just under different circumstances than the bug above. I'm not entirely sure how to make this crystal clear.

docs/bugs_and_glitches.md
```diff
@@ -628,161 +594,139 @@ CalcExpAtLevel:
...
+ ; DRAGON_FANG
 -	item_attribute 100, HELD_NONE, 0, CANT_SELECT, ITEM, ITEMMENU_NOUSE, ITEMMENU_NOUSE
 +	item_attribute 100, HELD_DRAGON_BOOST, 0, CANT_SELECT, ITEM, ITEMMENU_NOUSE, ITEMMENU_NOUSE
-```
-
-And:
-
-```diff
-; DRAGON_SCALE
+ ...
```
mid-kid: I don't like dotdotdots... They add very little to the context since the preceding part could be anywhere in the file.
I'd avoid them if possible.

docs/bugs_and_glitches.md
```diff
@@ -290,21 +273,36 @@ This is a bug with `CheckPlayerHasUsableMoves` in [engine/battle/core.asm](/engi
...
+```diff
+ CheckHiddenOpponent:
+-; BUG: This routine is completely redundant and introduces a bug, since BattleCommand_CheckHit does these checks properly.
+-	ld a, BATTLE_VARS_SUBSTATUS3_OPP
+-	call GetBattleVar
+-	and 1 << SUBSTATUS_FLYING | 1 << SUBSTATUS_UNDERGROUND
+ 	ret
+```
 
+**Alternate fix:**
```
mid-kid: I honestly think this entire section is redundant. There's always better and more efficient ways of fixing a bug, no need to list them all. Either remove this solution or remove the sloppy solution.
Now I think about it, wouldn't it be a good idea to `and a` before returning in `CheckHiddenOpponent`? Just to make absolutely sure the zero flag is unset. Not all moves and possible entrypoints for `CheckHiddenOpponent` have been tested.

docs/bugs_and_glitches.md
Rangi42: I think "`; DRAGON_SCALE`" provides sufficient context here. The alternative would be two separate code blocks, but that looks like you're editing two different files, and clarifying "Edit the same file again" is more wordy than just these ellipses.

docs/bugs_and_glitches.md
Rangi42: Agreed about both ways being unnecessary.

If ~~xor a~~ `and a` isn't necessary in `CheckHiddenOpponent`, IMO the fix shouldn't have it. If it *is* necessary, then the current fix isn't a fix at all, which is what testing was supposed to catch. (And yeah, if testing hasn't actually verified that the current fix works, then the `and a` is probaby necessary.)

PR #549, opened 2018-08-05
Update lzcomp to the newest version

aaaaaa123456789:
This pull request updates lzcomp to the newest version, the one currently in use in Prism. I don't remember when I fixed a small bug in it, but it was a very long time ago; apparently that fix was not ported here, which causes some pixels to render incorrectly. 
It does not affect the actual ROM, but it affects ROM hackers who attempt to use the tool.

Also, since the code works properly, and there's little point in fixing what isn't broken just because a tool falsely believes it is, I updated the tools' Makefile to remove the unnecessary `-Wall -Wextra` for lzcomp only. I updated the main Makefile as well, since the new version of lzcomp accepts some command-line options; I added `--` to the invocation of lzcomp to ensure that it doesn't interpret the first argument as an option if it happens to begin with a dash (unlikely, but possible).

### DIFF
--- a/tools/lzcomp.c
+++ b/tools/lzcomp.c
@@ -11,13 +11,24 @@ struct command {
   signed value:    17;
 };
 
+struct options {
+  const char * input;
+  const char * output;
+  unsigned char mode;
+};
+
 int main(int, char **);
+struct options get_options(int, char **);
+void usage(const char *);
 void error_exit(int, const char *, ...);
 void bit_flip(const unsigned char *, unsigned short, unsigned char *);
 unsigned char * read_file_into_buffer(const char *, unsigned short *);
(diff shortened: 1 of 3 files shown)


tools/lzcomp.c
```diff
@@ -220,7 +326,8 @@ unsigned short scan_backwards (const unsigned char * data, unsigned short limit,
   unsigned short position;
   for (position = 0; position < real_position; position ++) {
     if (data[position] != data[real_position]) continue;
-    for (current_length = 0; (current_length < limit) && (data[position - current_length] == data[real_position + current_length]); current_length ++);
+    for (current_length = 0; (current_length <= position) && (current_length < limit) &&
```
aaaaaa123456789: For the record, this is the single-line fix. The rest are new features I added eventually (including improved compression), but I figured there was no reason to hold them back in pokecrystal.

PR #550, opened 2018-08-07
Misc things

mid-kid:
Rationale for `Split print_move_description and mon_menu`:
The functions for the "main" pokemon menu are all called MonMenu, yet they were stored in menus/start_menu.asm. While this makes slight sense, since the Pokémon menu is implemented very tightly with the start menu, it makes the file unecessarily big and the things hard to find. Therefor, I moved all of the MonMenu and related functions to its own mon_menu.asm, and moved the existing mon_menu.asm (which contained the submenu) to mon_submenu.asm.

Other rationales are self-explanatory or included in the commit message.


macros/scripts/battle_commands.asm
```diff
@@ -21,7 +21,7 @@ ENDM
 	command applydamage             ; 0e
 	command criticaltext            ; 0f
 	command supereffectivetext      ; 10
-	command checkdestinybond        ; 11
+	command checkfaint              ; 11
```
Rangi42: This needs replacing in docs/bugs_and_glitches.md too.

PR #551, opened 2018-08-10
Remove dedoubled group sprites

igidrau:
Every group sprites with the exact same sprites were merged.
I also merged Lavender, Vermillion and Fuchsia with Indigo, because the only sprite that differed was a second "SPRITE_BIG_SNORLAX"


ketsuban:
Possibly I'm missing something but it seems to me this isn't really appropriate because it doesn't match the original game.

PR #553, opened 2018-08-19
Avoid warnings from gcc for lzcomp

Rangi42:
- `get_options`: unused parameter `[-Wunused-parameter]`
- `write_command_to_textfile`: comparison is always false due to limited range of data type `[-Wtype-limits]`
- `write_command_to_file`: comparison between signed and unsigned integer expressions `[-Wsign-compare]`
- `find_best_copy`: suggest parentheses around assignment used as truth value `[-Wparentheses]`
- `optimize`: this statement may fall through `[-Wimplicit-fallthrough]`

### DIFF
--- a/tools/lzcomp.c
+++ b/tools/lzcomp.c
@@ -18,7 +18,7 @@ struct options {
 };
 
 int main(int, char **);
-struct options get_options(int, char **);
+struct options get_options(char **);
 void usage(const char *);
 void error_exit(int, const char *, ...);
 void bit_flip(const unsigned char *, unsigned short, unsigned char *);
@@ -43,8 +43,8 @@ struct command * select_command_sequence(struct command **, const unsigned short
 struct command * merge_command_sequences(const struct command *, unsigned short, const struct command *, unsigned short, unsigned short *);
 unsigned short compressed_length(const struct command *, unsigned short);
 
-int main (int argc, char ** argv) {
-  struct options options = get_options(argc, argv);
+int main (int argc __attribute__((unused)), char ** argv) {
+  struct options options = get_options(argv);
   unsigned short size;
   unsigned char * file_buffer = read_file_into_buffer(options.input, &size);
   struct command * compressed = compress(file_buffer, &size);
@@ -57,7 +57,7 @@ int main (int argc, char ** argv) {
   return 0;
 }
 
-struct options get_options (int argc, char ** argv) {
+struct options get_options (char ** argv) {
   struct options result = {.input = NULL, .output = NULL, .mode = 0};
   const char * program_name = *argv;
   for (argv ++; *argv; argv ++) {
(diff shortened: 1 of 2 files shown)


aaaaaa123456789:
I have to ask, what's the point? It's obvious that these warnings have been completely useless, as they point to problems that don't exist, and were also unable to locate the bug that did.

tools/lzcomp.c
```diff
@@ -43,8 +43,8 @@ struct command * select_command_sequence(struct command **, const unsigned short
 struct command * merge_command_sequences(const struct command *, unsigned short, const struct command *, unsigned short, unsigned short *);
 unsigned short compressed_length(const struct command *, unsigned short);
 
-int main (int argc, char ** argv) {
-  struct options options = get_options(argc, argv);
+int main (int argc __attribute__((unused)), char ** argv) {
```
iimarckus: I’m fine with the others, but is this one really necessary?

Rangi42:
The bug was a missing `current_length <= position`; I'm unaware of any static checker that would have caught it. Handling these warnings is more about clarifying your intent for people reading the code who can't be certain it's bug-free, e.g. "did they mean to fall through here?" or "should this be comparing or assigning?" or "is get_options missing an argc check somewhere?"

Also the `-Wtype-limits` warning gets rid of some genuinely pointless code in `write_command_to_textfile`. (Although it might be worthwhile to remove the text output feature entirely. The only reason Prism has it, as far as I can tell, is to avoid using INCBINs for some compressed tilemaps—why didn't it just go with pokecrystal's method of versioning the uncompressed .tilemap files and letting `make` handle the compression?)

tools/lzcomp.c
Rangi42: Would you prefer using `-Wno-unused-parameter`?

aaaaaa123456789:
The text output feature is also very useful when debugging lzcomp's output. It's how I fixed that bug in the first place. While Prism does have the macro support for actually incorporating LZ text files (since I actually compressed some data manually in one place or another), that's not the reason why it is there.

The only case which may cause confusion is the fall through near the end. I can't see how the other examples can cause someone to get confused, since:

* `int main(int, char **)` is the prototype for `main` mandated by the standard, and the compiler won't let you omit the first argument even if you wanted to;
* the `command.value > 65535` comparison isn't incorrect, just unnecessary (I might remove it myself from Prism's version someday; I just can't see why this may be confusing);
* the intent of the comparison of the result of `fwrite` with `pos - buf` isn't made any clearer by the cast, and overflow is actually impossible since `fwrite` cannot return negative values and `pos - buf` can never be greater than four;
* the assignments with an extra pair of parentheses cannot possibly be comparisons, since the variable that is being assigned to is declared right before (not to mention that `if ((foo == bar))` is perfectly valid).

tools/lzcomp.c
iimarckus: I mean, does an unused argc actually trigger a warning?

mid-kid: I mean, why not actually use it by simply `if (argc <= 1) usage(argv[0]);`?

aaaaaa123456789: That's a redundant check, since the lack of a filename is already tested for by checking for the null pointer that the C standard says must exist at `argv[argc]`.

Rangi42: @iimarckus There is a warning, which is why I silenced it with `__attribute__((unused))`. @mid-kid @aaaaaa123456789 True, instead of `const char * program_name = *argv; <argv parsing loop> if (!*argv) usage(program_name);`, you could have `if (argc == 1) usage(argv[0]); <argv parsing loop>`.

mid-kid: Not really, since it can have multiple arguments and still lack the one required argument. It was a crappy suggestion as an alternative to `__attribute__((unused))`

PR #557, opened 2018-08-26
Miscellaneous changes

Rangi42:
- Use rgbasm's GameBoy graphics literal for initializing `wCreditsFaux2bpp`

In the 2BPP format, two bytes `%ABCDEFGH %abcdefgh` define eight pixels `%Aa %Bb %Cc %Dd %Ee %Ff %Gg %Hh`.  
Four valid colors: `%00` = 0 = white, `%01` = 1 = dark, `%10` = 2 = light, `%11` = 3 = black.  
Thus `` `22222222 `` = pixels `%10 %10 %10 %10 %10 %10 %10 %10` = bytes `%11111111 %00000000` = `$ff00`.

- `_Divide` results go in `hQuotient` and `hRemainder`, not in `hDividend` and `hDivisor`

Note that the result quotient is four bytes, but `hQuotient` is a three-byte buffer. Issue #558 explains this.

### DIFF
--- a/engine/battle/core.asm
+++ b/engine/battle/core.asm
@@ -7868,7 +7868,6 @@ CalcExpBar:
 	ld [hld], a
 	xor a
 	ld [hl], a
-; multiply by 64
 	ld a, 64
 	ldh [hMultiplier], a
 	call Multiply
--- a/engine/battle/core.asm
+++ b/engine/battle/core.asm
@@ -352,7 +352,7 @@ HandleBerserkGene:
 
 .reverse
 	call .enemy
-;	jr .player
+	; fallthrough
 
 .player
 	call SetPlayerTurn
@@ -366,7 +366,7 @@ HandleBerserkGene:
 	ld de, wOTPartyMon1Item
 	ld a, [wCurOTMon]
 	ld b, a
-;	jr .go
+	; fallthrough
 
 .go
 	push de
--- a/engine/battle/core.asm
+++ b/engine/battle/core.asm
@@ -3715,23 +3715,23 @@ TryToRunAwayFromBattle:
 	inc a
 	ld [wNumFleeAttempts], a
 	ld a, [hli]
-	ldh [hPartyMon1Speed + 0], a
+	ldh [hMultiplicand + 1], a
 	ld a, [hl]
-	ldh [hPartyMon1Speed + 1], a
+	ldh [hMultiplicand + 2], a
 	ld a, [de]
 	inc de
 	ldh [hEnemyMonSpeed + 0], a
 	ld a, [de]
 	ldh [hEnemyMonSpeed + 1], a
 	call Call_LoadTempTileMapToTileMap
-	ld de, hPartyMon1Speed
+	ld de, hMultiplicand + 1
 	ld hl, hEnemyMonSpeed
 	ld c, 2
 	call CompareBytes
 	jr nc, .can_escape
 
 	xor a
-	ldh [hMultiplicand], a
+	ldh [hMultiplicand + 0], a
 	ld a, 32
 	ldh [hMultiplier], a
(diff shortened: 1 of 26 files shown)


engine/math/math.asm
```diff
@@ -172,18 +172,18 @@ _Divide::
...
-	ldh [hDividend + 2], a
+	ldh [hQuotient + 1], a
 
 	ldh a, [hMathBuffer + 2]
-	ldh [hDividend + 1], a
+	ldh [hQuotient + 0], a
 
 	ldh a, [hMathBuffer + 1]
-	ldh [hDividend + 0], a
+	ldh [hQuotient - 1], a
```
mid-kid: I thought the idea was to move the hQuotient label back one byte. I don't mind this either, however.

Rangi42: Eventually yes; for now I at least wanted to not use the wrong label entirely.

Rangi42:
This is ready to merge.

engine/battle/core.asm
```diff
@@ -3715,23 +3715,23 @@ TryToRunAwayFromBattle:
...
-	ldh [hPartyMon1Speed + 1], a
+	ldh [hMultiplicand + 2], a
 	ld a, [de]
 	inc de
 	ldh [hEnemyMonSpeed + 0], a
 	ld a, [de]
 	ldh [hEnemyMonSpeed + 1], a
 	call Call_LoadTempTileMapToTileMap
-	ld de, hPartyMon1Speed
+	ld de, hMultiplicand + 1
```
mid-kid: Are you really sure about this one? It really isn't used as a multiplicand at this point, though I see a `call Multiply` a little further below.

engine/battle/core.asm
```diff
@@ -7869,7 +7869,7 @@ CalcExpBar:
 	xor a
 	ld [hl], a
 ; multiply by 64
```
mid-kid: Redundant comment...

hram.asm
```diff
@@ -62,41 +62,36 @@ ENDU ; ffb1
...
-hPartyMon1Speed:: dw ; ffb5
-	ds 5
-hMGStatusFlags:: db ; ffbc
+; math-related values
 
-NEXTU ; ffb3
+UNION ; ffb3
 ; inputs to Multiply
-	ds 1
+    ds 1
```
mid-kid: We don't take kindly to spaces in 'ese parts.

engine/battle/core.asm
Rangi42: Yes. The `call Multiply` at line 3737 only initializes `hMultiplicand + 0` because `+ 1` and `+ 2` were initialized with the party mon speed. And `hMultiplier` is 32, which makes sense given the [escape chance formula](https://bulbapedia.bulbagarden.net/wiki/Escape#Generation_I_and_II), which is just what this block is doing.

engine/battle/core.asm
Rangi42: Yeah...

engine/battle/core.asm
aaaaaa123456789: what's the point of `+ 0`?

Rangi42: Lots of multi-byte buffers use `+ 0` for consistency with nearby `+ 1`, `+ 2`, etc, and to avoid implying that `[wSomeBuffer]` is a useful single-byte quantity. Although we probably haven't applied this convention everywhere.

aaaaaa123456789: It seems rather obscure to me, as the first reaction to seeing something like `+ 0` is "what is this even for?". It's similar to the old `+- 1` in that regard; its meaning isn't really obvious.

It reminds me of some SQL versions where `foo` and `+foo` behaved differently.

PR #560, opened 2018-09-01
Fix #537: Palettes generated from PNGs use .gbcpal directly.

### LINKED ISSUE #537: .gbcpal
Currently palettes make it through the build like this:

`image.png` -> `image.gbcpal` -> `image.pal`

This is wrong. It should be like pokeruby:

`image.png` -> `image.gbcpal`
`image.pal` -> `image.gbcpal`

Currently any .pal file will be overwritten by a newer png with the same name. Instead .pal files should override the png palette. This means all `.pal` includes meant to be generated from a png should become `.gbcpal` incbins.

This also means .pal doesn't necessarily need to be asm anymore, but I'm not sure if there's any reason to change this yet.

Rangi42:
This avoids overwriting a .pal file from a .png with the same name.  
It also prevents people from trying to edit generated .pal files.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -99,6 +99,20 @@ pokecrystal11.gbc: $(crystal11_obj) pokecrystal.link
 		tools/lzcomp -- $< $@)
 
 
+### Pokemon pic graphics rules
+
+gfx/pokemon/%/front.animated.2bpp: gfx/pokemon/%/front.2bpp gfx/pokemon/%/front.dimensions
+	tools/pokemon_animation_graphics -o $@ $^
+gfx/pokemon/%/front.animated.tilemap: gfx/pokemon/%/front.2bpp gfx/pokemon/%/front.dimensions
+	tools/pokemon_animation_graphics -t $@ $^
+gfx/pokemon/%/bitmask.asm: gfx/pokemon/%/front.animated.tilemap gfx/pokemon/%/front.dimensions
+	tools/pokemon_animation -b $^ > $@
+gfx/pokemon/%/frames.asm: gfx/pokemon/%/front.animated.tilemap gfx/pokemon/%/front.dimensions
+	tools/pokemon_animation -f $^ > $@
+gfx/pokemon/%/back.2bpp: gfx/pokemon/%/back.png
+	$(RGBGFX) -h -o $@ $<
+
+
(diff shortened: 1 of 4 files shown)


mid-kid:
If the goal is to allow a user to override the palettes though a manual .pal file, this isn't the right way to do it.
Instead, I propose modifying scan_includes to output dependencies in a build directory if they don't exist in the source tree (like pokegold-spaceworld does), and move to building everything in a build directory. This'd allow a user to override any otherwise built files by placing them in-tree, including palettes.
That said, I wouldn't personally override palettes through custom .pal files, so if the intent is simply removing the intermediate (and rather useless without doing something like I detailed above) .pal file, this PR is correct.

Rangi42:
The goal isn't to let .pal files override auto-generated .gbcpal←.png files; it's to prevent auto-generated .png→.pal files from overwriting the original .pals.

In general, .pal files are unrelated to any .png/.2bpp files. They provide palette data, entered directly by the user, for whatever reason. For instance, the various palettes loaded by CGB layouts. Some of them have the same name as adjacent .png files, like pack.pal and pack.png, but that doesn't mean we could derive pack.gbcpal from pack.png if pack.pal didn't exist.

The only cases where it even makes sense to derive a palette from an image is for Pokémon and trainer sprites. In those cases, a .gbcpal file is generated from the .png file. (I kept the name "normal.gbcpal", but actually "front.gbcpal" would be better so the Makefile won't need an explicit rule for front.png→normal.gbcpal. Will change that.)

(Yes, @yenatch suggested having two build rules, `image.png` → image.gbcpal` and `image.pal` → `image.gbcpal`, but then if they disagree you're relying on the modification times to determine which gets used. It's simpler to just, if you want to manually define a palette for some Pokémon or trainer, edit palettes.asm to say `INCLUDE "normal.pal"` instead of `INCBIN "normal.gbcpal", middle_colors`.)

A separate build directory would be nice, but is unrelated to fixing this issue, which is just that editing a .png can overwrite a .pal of the same name. (Like, I think if you edit pack.png, then pack.pal could turn into just `RGB 31, 31, 31 / RGB 20, 20, 20 / RGB 10, 10, 10 / RGB 0, 0, 0`.)

PR #561, opened 2018-09-02
Revise documentation

Rangi42:
All the INSTALL.md sections have similar structure

### DIFF
--- a/FAQ.md
+++ b/FAQ.md
@@ -32,7 +32,7 @@ You need to install `gcc`. If you're using Cygwin, re-run its setup, and at "Sel
 
 ### "ERROR: `UNION` already defined"
 
-Download [the latest **rgbds** release][rgbds]. Versions earlier than 0.3.3 will not work.
+Download [**rgbds 0.3.7**][rgbds]. Older versions will not work.
 
 ### "Segmentation fault" from `rgbgfx`
 
@@ -50,9 +50,9 @@ Run `make clean` to remove all the old `o` files, then re-run `make`.
 
 ### "Syntax error"
 
-You've made a mistake while writing some of the code. Re-read the modifications you've made to the file it complains about and try to compare them with other code.
+If you have not changed any of the asm, make sure you have the latest version of pokecrystal and the correct version of rgbds (see [INSTALL.md](INSTALL.md)).
 
-If you're not using the latest version of [**rgbds**][rgbds], update this as well. Don't forget to `make clean` afterwards.
+If you added or changed any code, you've made a mistake while writing some of it. Re-read the modifications you've made to the file it complains about and try to compare them with other code.
 
 
 ## How do I edit maps?
--- a/FAQ.md
+++ b/FAQ.md
@@ -52,7 +52,7 @@ Run `make clean` to remove all the old `o` files, then re-run `make`.
 
 You've made a mistake while writing some of the code. Re-read the modifications you've made to the file it complains about and try to compare them with other code.
 
-If you're not using the latest version of [**rgbds**](https://github.com/rednex/rgbds/releases), update this as well. Don't forget to `make clean` afterwards.
+If you're not using the latest version of [**rgbds**][rgbds], update this as well. Don't forget to `make clean` afterwards.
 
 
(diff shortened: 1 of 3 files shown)


Rangi42:
@mid-kid Please comment with any revisions. This is more consistent but maybe too verbose. Though I couldn't find a way to get rid of the per-distro headers either.

INSTALL.md
```diff
@@ -67,18 +76,30 @@ make crystal11
 ```
 
 
-## Mac
+## Mac OS X
+
+Open **Terminal** and enter the following commands.
```
mid-kid: I'd move this to below the first header.

INSTALL.md
```diff
@@ -67,18 +76,30 @@ make crystal11
...
+```
 
+To install **rgbds**:
+
+```bash
 git clone --depth=1 https://github.com/rednex/rgbds
-cd rgbds
-sudo make CFLAGS=-O2 install
-cd ..
+sudo make -C rgbds CFLAGS=-O2 install
```
mid-kid: Can you confirm mac's version of make supports the -C option? I left it because I wan't entirely sure...

INSTALL.md
```diff
@@ -89,63 +110,107 @@ To build **pokecrystal.gbc**:
...
+#### Debian or Ubuntu
+
+To install the software required for **pokecrystal**:
 
 ```bash
-sudo apt-get install make gcc bison git libpng-dev
+sudo apt-get install make gcc git
+```
 
+To install **rgbds**:
```
mid-kid: Technically, make gcc and git are also required to build rgbds... I don't really see the point in splitting this off.

FAQ.md
```diff
@@ -52,7 +52,7 @@ Run `make clean` to remove all the old `o` files, then re-run `make`.
 
 You've made a mistake while writing some of the code. Re-read the modifications you've made to the file it complains about and try to compare them with other code.
 
-If you're not using the latest version of [**rgbds**](https://github.com/rednex/rgbds/releases), update this as well. Don't forget to `make clean` afterwards.
+If you're not using the latest version of [**rgbds**][rgbds], update this as well. Don't forget to `make clean` afterwards.
```
luckytyphlosion: Specify the current version of rgbds to download? There may be some backwards compatibility issues in the future.

INSTALL.md
Rangi42: To make it clear which packages are only required for building rgbds, and so the `git clone` and `sudo make` commands common to all distros are more obvious. (I was considering a final section that just has the `git clone`, `cd pokecrystal`, `make pokecrystal`, `make pokecrystal11` commands, and every OS section would link to it; does that sound like an improvement?)

mid-kid: Yeah, that does sound like an improvement, with proper links.
Just say "continue by [installing pokecrystal]".

Makefile
```diff
@@ -230,5 +227,13 @@ gfx/unknown/unknown_egg.2bpp: rgbgfx += -h
 
 %.gbcpal: %.png
 	$(RGBGFX) -p $@ $<
+
 %.dimensions: %.png
 	tools/png_dimensions $< $@
+
+
+### These don't get built from other files, but have to exist on their own.
```
mid-kid: There is little point in doing this. Make will fail if a required file isn't found, and not do anything to try to create them, unless a rule for them already exists, in which case this will override and clear said rule.

I propose that we should simply not define any rules for these files if we don't want make to make them ;)

Rangi42: I assumed those rules existed for a reason, like to give better error messages. @yenatch added `pal` and `bin` in aaf26de8edab0c814463c24674c8a6f04c2a80f4, and `blk` and `tilemap` in 6122490cc339b5cab2ef9b25a7cf82b56de16413. ...Hmm, the `tilemap` rule is gone now, so maybe the rest can be deleted too?

Rangi42: Extensions that don't get generated: `asm`, `bin`, `blk`, `pal`, `tilemap`, `tilemap.rle`, `attrmap`

PR #564, opened 2018-09-16
abloobloobloobloobloo

mid-kid:


engine/tilesets/tileset_palettes.asm
```diff
@@ -123,12 +123,12 @@ LoadMansionPalette:
 	call FarCopyWRAM
 	ld a, BANK(wBGPals1)
 	ld de, wBGPals1 palette PAL_BG_WATER
-	ld hl, MansionPalette1 + 6 palettes
+	ld hl, MansionPalette1 palette PAL_BG_ROOF
```
Rangi42: This should be "`MansionPalette1 palette 6`"; it's the `WATER` palette, not the `ROOF` palette, and just happens to be at index 6.

home.asm
```diff
@@ -56,9 +56,11 @@ INCLUDE "home/predef.asm"
 INCLUDE "home/window.asm"
 INCLUDE "home/flag.asm"
 
-Unreferenced_Function2ebb::
-	ld a, [wMonStatusFlags]
-	bit 1, a
+Unreferenced_CheckBPressedDebug::
+; Used in debug ROMs to walk through walls and avoid encounters.
```
Rangi42: There seems to be no walk through walls code left, so you could mention pokegold-spaceworld here, like data/pokemon/unused_pic_banks.asm.

engine/tilesets/tileset_palettes.asm
mid-kid: My reasoning was that the roof palette is later overwritten with palette 8, so it's nice to know you're sourcing it from the location of the ROOF palette in case you want to re-order or otherwise modify it. On second thought, that's not really useful here.

home.asm
mid-kid: :+1:

mid-kid: This isn't specific to pokegold-spaceworld, as it's rather generic, and should exist in all of the Debug ROMs if I'm not mistaken.
I'm not sure what I'd mention about spaceworld here, as there's not much more to link to about it.

PR #565, opened 2018-09-16
Identify some Spaceworld beta content

Rangi42:


engine/menus/naming_screen.asm
```diff
@@ -27,7 +27,7 @@ NamingScreen:
 	ldh [hMapAnims], a
 	ldh a, [hInMenu]
 	push af
-	ld a, $1
+	ld a, TRUE
```
mid-kid: There's a lot more cases of this. If you're doing one of them, do all of them, or delegate it to a different PR.

Issue #556, opened 2018-08-26
Improve music macros


Rangi42:
PR #321 (closed for being years out of date) had some possibilities for changes to sound macros, mostly taken from pokered. In general it would be nice to have meaningful parameter values for each macro, and internally combine them into the raw bytes. Macros like `sound_duty` and `slidepitchto` already do this.

Prism also has an improved header structure. Instead of defining each channel with `musicheader`, which takes a parameter for the channel count that's only meaningful for the first one, it sets the channel count with `channelcount` and defines each channel with `channel`.

mid-kid:
Ideas thrown around on the Discord:
- `volume`: Split into the left and right components, drop the useless hex notation, warn people about values higher than 7.
- `dutycycle`: Make `DUTY_X` constants for the different percentages, warn the user about values greater than 3.
- `stereopanning`: Unpack all the hex values into bitwise notation (`%00000000`)
- `tone`: Reduce the amount of hexadecimal characters to 3 (`$xxxx` → `$xxx`), and warn users about values bigger than 2047.
- `vibrato`: Just do this: https://github.com/pret/pokecrystal/wiki/Sound-engine-commands#e1-vibrato-byte-byte
- Actually, for a lot of commands, that document already provides good macros: `soundinput`, `intensity`, `notetype`...
- any: Use a constant for true/false arguments that have a meaning beyond a boolean expression.
- any: Warn the user if using too big values, and truncate them properly.

mid-kid:
Just a note about stereopanning, it's only used with either `$f0` `$f` or `$ff` as arguments. If this only applies the bits relevant to the current channel, the bitwise constants would actually be useless.

Monstarules:
@Rangi42 the macros names are fine. We could do with a few updates on the wiki though, I don't think the actual guide for the effects are written in plain enough english, though.

@mid-kid stereopanning in Pokecrystal only uses $f0 $f and $ff , but the effect that it provides (sound panning when stereo is enabled) is used in many songs for romhacks, so perhaps that wouldn't be entirely useless, although for the sake of vanilla Crystal it is.

My suggestion is to start by removing the useless things and updating the wiki. That would be removing: - `unknownmusic0xe2`
-`unknownmusic0xe7`
- `unknownmusic0xe8`
- `unknownmusic0xf9` 
[...]

Issue #540, opened 2018-07-08
rename var/byte/mem script commands


yenatch:
before | after | what it actually does
---|---|---
writebyte | setval `value` | `[wScriptVar] = value`
addvar | addval `value` | `[wScriptVar] += value`
copybytetovar | readval `address` | `[wScriptVar] = [address]`
copyvartobyte | writeval `address` | `[address] = [wScriptVar]`
loadvar | write `address`, `value` | `[address] = value`
checkcode | getvar `var` | `[wScriptVar] = [var]`
writevarcode | setvar `var` | `[var] = [wScriptVar]`
writecode | setvar `var`, `value` | `[var] = value`
vartomem | printval `buffer` | `[wStringBuffer{buffer+3}] = PrintNum([wScriptVar])`
loadmemtrainer | settrainer | `[wOtherTrainer] = [wTempTrainer]`
pokenamemem | pokename `species`, `buffer` | `[wStringBuffer{buffer+3}] = GetPokemonName(species \|\| [wScriptVar])`
itemtotext | itemname `item`, `buffer` | `[wStringBuffer{buffer+3}] = GetItemName(item \|\| [wScriptVar])`
mapnametotext | mapname `buffer` | `[wStringBuffer{buffer+3}] = GetLandmarkName()`
landmarktotext | landmarkname `landmark`, `buffer` | `[wStringBuffer{buffer+3}] = GetLandmarkName(landmark)`
trainertotext | trainername `class`, `trainer`, `buffer` | `[wStringBuffer{buffer+3}] = GetTrainerName(class, trainer)`
stringtotext | print `string`, `buffer` | `[wStringBuffer{buffer+3}] = CopyName1([wScriptBank], string)`
readcoins | printcoins `buffer` | `[wStringBuffer{buffer+3}] = PrintNum([wCoins])`
readmoney | printmoney `buffer` | `[wStringBuffer{buffer+3}] = PrintNum([wMoney])`

mid-kid:
`setval` -> `loadval`?

`getvar` -> `readvar` and `setvar` -> `writevar` for consistency with the `val` equivalents.

`printval` -> `printnumval` or `printvalnum`. Distinguish it from `print` properly.
Or, make it generate from `print` if _NARG < 2.

Also, the last column you just wrote is really useful to have. Might want to update `docs/event_commands.md` with them.

Rangi42:
Reminder to keep the old names in macros/legacy.asm.

[...]

PR #573, opened 2018-11-05
 text_jump → text_far 

mid-kid:


mid-kid:
I seriously would like to clear up the confusion about text commands vs control characters. The function in question I was editing deals with filtering out control characters, and `TX_START` and `TX_BOX` are both text commands, thus really not appropriate as constants for this function.

Text commands are used as part of the scriptable text engine. TX_START simply means a `@`-terminated string follows, which may contain as many characters and/or control characters as it wants. (This is also why I'd love to prefix every text command in `macros/scripts/text.asm` with `text_`, to clear up the confusion between these and control characters such as `para`, `line`, `done`, etc, especially since they're used interchangeably in a lot of text scripts.)

mid-kid:
Anyway, this is ready to merge, yes.

Rangi42:
Thanks for the explanation. How about opening an issue for renaming the text macros? We already have #556 for music and #540 for script commands.

PR #574, opened 2018-11-06
Make the `connection` macro simpler to use

mid-kid:
It now only takes one numerical parameter, which is the offset of the
target map relative to the source map, much like in AdvancedMap. This
makes it easier to make connections and avoids having to calculate these
values by hand, and/or "mess with the values 'till it works", as many
have been doing thus far. It's just one, easy-to-understand value.

To convert from the old macro to the new macro, just take the fourth and
the fifth parameter, and calculate `<4th_param> - <5th_param>`. The
result is the value required for the new macro.

### DIFF
--- a/data/maps/attributes.asm
+++ b/data/maps/attributes.asm
@@ -17,81 +17,78 @@ CURRENT_MAP_HEIGHT = \2_HEIGHT
 ENDM
 
 ; Connections go in order: north, south, west, east
-_connection: MACRO
-;\1: direction
-;\2: map name
-;\3: map id
-;\4: x offset for east/west, y offset for north/south
-;\5: distance offset?
-;\6: strip length
-if "\1" == "north"
-	map_id \3
-	dw \2_Blocks + \3_WIDTH * (\3_HEIGHT - 3) + \5
-	dw wOverworldMapBlocks + \4 + 3
-	db \6
-	db \3_WIDTH
-	db \3_HEIGHT * 2 - 1
-	db (\4 - \5) * -2
-	dw wOverworldMapBlocks + \3_HEIGHT * (\3_WIDTH + 6) + 1
-elif "\1" == "south"
-	map_id \3
-	dw \2_Blocks + \5
-	dw wOverworldMapBlocks + (CURRENT_MAP_HEIGHT + 3) * (CURRENT_MAP_WIDTH + 6) + \4 + 3
-	db \6
-	db \3_WIDTH
-	db 0
-	db (\4 - \5) * -2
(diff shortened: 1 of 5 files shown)


Rangi42:
@mid-kid Here's my revision to your `connection` macro: https://pastebin.com/RyTfdtm1

It makes these changes:

- Legacy support for the old 6-parameter version
- Error message for invalid directions
- Splits the macro into two phases, (a) calculating the direction-specific variables and (b) declaring the data (so declaration only has to be done once)

The variables all have placeholder names, `_r`, `_s`, `_t`, and `_foo1-6`. I'll leave naming them to you, since you may want to rename the actual WRAM labels at the same time.

Your variable `_l` is now `_foo3`. I chose to have four separate north/south/east/west cases for setting it, with the north+south and east+west cases identical, instead of having a separate north+south/west+east condition above the main four-direction conditional. I also refactored it to make the `min` operation obvious, in case rgbds ever adds `MIN` and `MAX` functions.

mid-kid:
I have a couple of notes on it:
- I don't think the `_r` thing is necessary. It adds a couple of lines extra that doesn't really make the intent any clearer. If `_t` is smaller than 0, it should underflow into `_s`.
- I've been considering writing all the values as variables, and using them only at the end, but is that really at all necessary? Unless it really helps clarify what they mean, though I'm not sure `_blk` and `_map` are really that much more descriptive.

mid-kid:
I adapted the macro based on the suggestions above, but I've also made another attempt, to contrast the use of variables with just regular old `db`s: https://pastebin.com/raw/VqvUhbr0

There's a couple of things these variables don't properly explain, and while these things really can't be explained in such a macro, the fact they're named here, might actually be confusing:
- `_y` and `_x` are used differently depending on the kind of connection. In horizontal connections, `_y` is the player's new absolute coordinate, while `_x` is added to the player's current position, making it an offset. It's exactly the other way around with vertical connections.
- `_win` is also used differently depending on the direction of the connection, but I'm not even sure what `wOverworldMapAnchor`, which this value is used for, even means.
I'm pretty sure `_blk`, `_map` and `_len` are clear, though.

Rangi42:
* Agreed about `_r`.
* I do think it helps to have one definitive instance of `db`+`dw` uses, to make it clear that all four directions have the same data structure and to factor out pieces that are common to all four. It means having to define some extra variables, but their names can correspond to the WRAM labels they go with.

Rangi42:
Can you change the comment in macros/scripts/audio.asm to `LEGACY: Support for old sound_duty macro` too?

Issue #575, opened 2018-11-12
Rename text commands


mid-kid:
Text commands are used as part of the scriptable text engine. `TX_START` simply means a `@`-terminated string follows, which may contain as many characters and/or control characters as it wants.

There's a lot of such "text commands", ranging from running an asm script to playing a sound. These are used interchangeably in "text scripts" with control characters such as `para`, `line`, `cont` and such. To clear up confusion about which is which, I'd love to get all of the text commands that aren't already, prefixed with `text_`. This doesn't work as well for every single text command, so we should figure out what to rename some of the commands to. Maybe it's a good idea to make some exceptions, for example for the `sound_` "text commands".

Rangi42:
[docs/text_commands.md](../blob/master/docs/text_commands.md) lists all the current script and command macros with brief descriptions. A summary:

- `TX_START` == `text_start`
- `TX_RAM` == `text_from_ram` *address*
- `TX_BCD` == `text_bcd` *address*, *flags*
- `TX_MOVE` == `text_move` *address*
- `TX_BOX` == `text_box` *address*, *height*, *width*
- `TX_LOW` == `text_low`
- `TX_WAIT_BUTTON` == `text_waitbutton`
- `TX_SCROLL` == `text_scroll`
- `TX_START_ASM` == `start_asm`
- `TX_NUM` == `deciram` *address*, *bytes*, *digits*
- `TX_EXIT` == `interpret_data`
- `TX_SOUND_DEX_FANFARE_50_79` == `sound_dex_fanfare_50_79`
- `TX_DOTS` == `limited_interpret_data` *n*
- `TX_LINK_WAIT_BUTTON` == `link_wait_button`
- `TX_SOUND_DEX_FANFARE_20_49` == `sound_dex_fanfare_20_49`
- `TX_SOUND_ITEM` == `sound_item`
- `TX_SOUND_CAUGHT_MON` == `sound_caught_mon`
- `TX_SOUND_DEX_FANFARE_80_109` == `sound_dex_fanfare_80_109`
- `TX_SOUND_FANFARE` == `sound_fanfare`
- `TX_SOUND_SLOT_MACHINE_START` == `sound_slot_machine_start`
- `TX_STRINGBUFFER` == `text_buffer` *id*
- `TX_DAY` == `current_day`
- `TX_FAR` == `text_far` *address*

[...]

PR #577, opened 2018-11-17
Resolve issue #575: Rename text commands

Rangi42:


macros/scripts/text.asm
```diff
@@ -136,3 +136,10 @@ text_far: MACRO
 	dw \1
 	db BANK(\1)
 ENDM
+
+	enum_start $50
```
mid-kid: Either use `enum_set` or remove `enum_set` from the macros.
Maybe it'd be a good idea to remove this line and test the game. If nothing breaks, this commit is correct ;)

mid-kid:
I've double-checked everything and haven't been able to find anything wrong. I haven't checked if any `@`s around the code have been missed, however.

macros/scripts/text.asm
Rangi42: Nothing I've tested is breaking without it. :)

PR #572, opened 2018-11-03
Install pkg-config & flex with make, gcc, git

Fontbane:
May require putting flex and pkg-config in other parts of the instructions too

PR #578, opened 2018-11-19
Fixes in the animation subsystem.

mid-kid:
Supersedes https://github.com/pret/pokecrystal/pull/572


audio/engine.asm
```diff
@@ -1136,7 +1136,7 @@ ReadNoiseSample:
 ParseMusic:
 ; parses until a note is read or the song is ended
 	call GetMusicByte ; store next byte in a
-	cp $ff ; is the song over?
+	cp endchannel_cmd ; is the song over?
 	jr z, .endchannel
 	cp $d0 ; is it a note?
```
Rangi42: `$d0` is used twice in audio/engine.asm as the first non-note byte value; it should be named in constants/audio_constants.asm or macros/scripts/audio.asm, not sure which.

mid-kid: The problem is that the actual audio engine commands start at `$d8`, weirdly enough. I have no idea where the substraction by `8` is done, however.

mid-kid: Oh, I just noticed `octave` covers those 8 values.

mid-kid:
I wasn't able to put `FIRST_MUSIC_CMD` in `constants/audio_constants.asm` because the macros are included before the constants are.

mid-kid:
Okay, this is epic. Having "cleaned up" `BattleAnimCmd_EnemyFeetObj` and `BattleAnimCmd_PlayerHeadObj` it's very clear that they're extremely inappropriately named.
Oh, and so is everything related to them. Would suck if I had to rename them now when I just wanted to make some cool anims. Hahahahahaha... :sob:

`anim_enemyfeetobj`, `BattleAnimCmd_EnemyFeetObj`, `anim_playerheadobj` and `BattleAnimCmd_PlayerHeadObj` all load the graphics for both `ANIM_GFX_PLAYER` (the player's head) and `ANIM_GFX_ENEMY` (the enemy's feet). The former two only load one row of tiles, the latter two load **two rows** of tiles. For the enemy's feet it loads bottom-most tiles of the sprite, and for the player's head it loads the top-most tiles of the sprite. These are both vram-to-vram transfers. 

Now, these graphics must be used somewhere, right? Well, let's start with a simple example. I assume you're familiar with the move "Tackle". The animation plays as follows: The user's HP bar disappears, the user moves forward, a "hit marker" appears on the enemy, and the user moves back. Let's follow what it does:

First of all, it calls `BattleAnim_FollowPlayerHead_0`. This function loads the graphics through `anim_playerheadobj`, which means we'll have two rows loaded for each. Next, it uses `ANIM_BG_HEAD_FOLLOW`.

`ANIM_BG_HEAD_FOLLOW` does the following:
* On the player's turn, it loads and activates `ANIM_OBJ_PLAYERHEADFOLLOW`, on the enemy's turn, it loads and activates `ANIM_OBJ_ENEMYHEADFOLLOW`. It puts either object at different x coordinates, but the same y coordinate. This is probably fixed by offsetting in the appropriate OAMSET.
* On the next frame, it clears the appropriate area of the tilemap for whoever's the target of this attack.

(NOTE: Whoever's side has to be loaded/cleared on whose turn is decided through `BG_EFFECT_STRUCT_BATTLE_TURN`. I assume this is the third parameter to `anim_bgeffect`. This can be used if you want to flip the logic around. Calling the labels used around `BGEffect_CheckBattleTurn`, `.player_turn` is misleading, since it's only used to know which _side_ this effect has to take place in.)

This is basically the result:


All of the objects created are static single-frame objects that do _nothing_. In this case they're used so the background around the user can be offset to create the effect that it's "moving". This is done later in the move effect, through `ANIM_BG_TACKLE`.

TL;DR:
* `ANIM_GFX_PLAYER`: Graphics for the player's head
* `ANIM_GFX_ENEMY`: Graphics for the enemy's feet.
* `anim_enemyfeetobj`/`BattleAnimCmd_EnemyFeetObj`: Load one row for `ANIM_GFX_PLAYER`/`ANIM_GFX_ENEMY`.
* `anim_playerheadobj`/`BattleAnimCmd_PlayerHeadObj`: Load two rows for `ANIM_GFX_PLAYER`/`ANIM_GFX_ENEMY`.
* `ANIM_OBJ_PLAYERHEADFOLLOW`: Object for ***two rows*** of the enemy's feet.
* `ANIM_OBJ_ENEMYHEADFOLLOW`: Object for ***two rows*** of the player's head.
* `ANIM_OBJ_PLAYERFEETFOLLOW`: Object for ***one row*** of the enemy's feet.
* `ANIM_OBJ_ENEMYFEETFOLLOW`: Object for ***one row*** of the player's head.
* `ANIM_BG_HEAD_FOLLOW`: Replace the ***two row*** head or feet of either the user or the target with an object.
* `ANIM_BG_FEET_FOLLOW`: Replace the ***single row*** head or feet of either the user or the target with an object.
* `BattleAnim_FollowPlayerHead_0`: Load graphics and replace the ***two row*** head or feet of the target with an object.
* `BattleAnim_FollowPlayerHead_1`: Load graphics and replace the ***two row*** head or feet of the user with an object.
* `BattleAnim_FollowEnemyFeet_0`: Load graphics and replace the ***single row*** head or feet of the target with an object.
* `BattleAnim_FollowEnemyFeet_1`: Load graphics and replace the ***single row*** head or feet of the user with an object.
* `LoadBattleAnimObj`: loads an ANIM_GFX, not an object.

Caveats when using `anim_enemyfeetobj` or `anim_playerheadobj`:
* They have to be loaded after any other GFX, as loading GFX afterwards will override the se.
* They won't work with `anim_4gfx` or `anim_5gfx`, since they take two of the available 5 slots in `wBattleAnimTileDict`. (Using them regardless might cause writing where it shouldn't!)
* You have to make sure there's enough space to load them at the end of the VRAM area, otherwise they'll overwrite existing tiles. (`loaded_tiles <= 0x80 - 0x31 - 7 * 2 - 6 * 2` for `playerhead`, and `loaded_tiles <= 0x80 - 0x31 - 7 - 6` for `enemyfeet`)

mid-kid:
Alright, proposal:

Original|New
-|-
`ANIM_GFX_PLAYER`|`ANIM_GFX_PLAYERHEAD`
`ANIM_GFX_ENEMY`|`ANIM_GFX_ENEMYFEET`
`anim_enemyfeetobj`|`anim_battlergfx_2row`
`anim_playerheadobj`|`anim_battlergfx_1row`
`BattleAnimCmd_EnemyFeetObj`|`BattleAnimCmd_BattlerGFX_1Row`
`BattleAnimCmd_PlayerHeadObj`|`BattleAnimCmd_BattlerGFX_2Row`
`ANIM_OBJ_PLAYERFEETFOLLOW`|`ANIM_OBJ_ENEMYFEET_1ROW`
`ANIM_OBJ_ENEMYFEETFOLLOW`|`ANIM_OBJ_PLAYERHEAD_1ROW`
`ANIM_OBJ_PLAYERHEADFOLLOW`|`ANIM_OBJ_ENEMYFEET_2ROW`
`ANIM_OBJ_ENEMYHEADFOLLOW`|`ANIM_OBJ_PLAYERHEAD_2ROW`
`ANIM_BG_FEET_FOLLOW`|`ANIM_BG_BATTLER_OBJ_1ROW`
`ANIM_BG_HEAD_FOLLOW`|`ANIM_BG_BATTLER_OBJ_2ROW`
`BattleAnim_FollowEnemyFeet_0`|`BattleAnim_TargetBattlerObj_1Row`
`BattleAnim_FollowEnemyFeet_1`|`BattleAnim_UserBattlerObj_1Row`
`BattleAnim_FollowPlayerHead_0`|`BattleAnim_TargetBattlerObj_2Row`
`BattleAnim_FollowPlayerHead_1`|`BattleAnim_UserBattlerObj_2Row`
`LoadBattleAnimObj`|`LoadBattleAnimGFX`

Alternatively, we could use `1Row` and `2Row` as prefixes instead, or remove the underscore separating the suffix. I'm not entirely sure what fits best in the current style.
Also I'm not convinced "Battler" is the best word for this.

Rangi42:
I'm okay with `1Row` and `2Row` as suffixes. "Battler" also sounds fine; maybe "Fighter" or "Combatant" or "PicSprite", to avoid confusion with "BattleMon"?

Be sure to update docs/battle_anim_commands.md, and mention those caveats to the two commands.

Also, related to the earlier misc edits: the Belly Drum fix in docs/bugs_and_glitches.md needs `call BattleCommand_AttackUp2` to be surrounded by `push bc`/`pop bc` (see the #pokecrystal logs).

mid-kid:
I decided to go with just "Battler" in the end, since we're talking about participants in what we otherwise call a "Battle". It might just be that I've never seen the term used as such before so it sounds a bit weird, but I don't think the other terms really fit the bill either.
I've also renamed `UserBattlerObj` to simply `UserObj`, and did the same for Target. Using "Battler" here made no sense, and I feel removing it doesn't strip it from its meaning.

mid-kid:
This is now ready for merge. I doubt I'll make any further changes for now.

Rangi42:
I'll review it soon.

PR #582, opened 2018-12-25
Use local rgbds/ binaries if they exist

Rangi42:
Assuming that rgbds 0.4.0 will be [released soon](https://github.com/rednex/rgbds/pull/311#issuecomment-446012107), and that we [take advantage](https://github.com/pret/pokecrystal/wiki/Code-cleanup#take-advantage-of-up-to-date-rgbds-features) of its new features, this will make it easier for people to build up-to-date pokecrystal without globally installing 0.4.0 (which might fail for forks using older versions).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,10 +4,15 @@ else
 SHA1 := sha1sum
 endif
 
-RGBASM := rgbasm
-RGBFIX := rgbfix
-RGBGFX := rgbgfx
-RGBLINK := rgblink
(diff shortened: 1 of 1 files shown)


Rangi42:
Thank you, never mind this then. If/when we update rgbds again, those solutions can go in the FAQ.

mid-kid:
Yeah, mentioning it in the FAQ or the wiki definitely sounds like a good idea.

PR #583, opened 2018-12-25
allow setting RGBASM/etc in environment vars

melthelesbian:


aaaaaa123456789:
An overall `RGBDS` variable to set the path to all tools would also go a long way.

mid-kid:
What's wrong with modifying `$PATH` to set the path to all tools?

meltheadorable:
@mid-kid it's pretty common practice to allow setting a search-path for specific things as an environment variable for critical libraries or build tools that may have project-specific considerations, globally setting the path can have other undesirable implications, and manually setting it on every invocation is pretty inconvenient, i don't really see a problem with this compromise, as it doesn't introduce any specific checks for where to look, just allows an override.

Rangi42:
@melthelesbian Mind renaming `RGBDS_DIR` to just `RGBDS`? Might as well make it shorter to specify on the command line.

aaaaaa123456789:
@mid-kid As far as I know, that doesn't even work. Environment variables have to be set before the command, so you'd have to do `PATH=../rgbds:$PATH make`, which is already awkward. I'm not sure if the OS will consider relative paths in PATH; I've never tried it.

Makefile
```diff
@@ -4,7 +4,7 @@ else
 SHA1 := sha1sum
 endif
 
-RGBDS_DIR ?= 
+RGBDS ?= 
 RGBASM ?= $(RGBDS_DIR)rgbasm
```
Rangi42: The uses of `RGBDS_DIR` need changing too. :P Thanks!

melthelesbian:
Setting the path will also break wrapper scripts that rely on the original commands being the ones in the path, like the one i was using to test this worked properly -- and I can see situations where somebody may want to use light wrapper scripts like the one I was using to modify or otherwise process the commands.

mid-kid:
@aaaaaa123456789 The OS will consider relative paths and yes, that'll work.

mid-kid:
>globally setting the path can have other undesirable implications, and manually setting it on every invocation is pretty inconvenient

Both of these problems apply to the `RGBDS_DIR` solution as well.

I mean, I don't mind this solution, but I've never had any case where I've needed anything like this for this or any other project. Scripts that mess with using two different versions of the same tool should adapt the PATH themselves, instead of calling them with the adapted PATH.

melthelesbian:
> Both of these problems apply to the RGBDS_DIR solution as well.

@mid-kid globally setting `RGBDS_DIR` only has other implications for projects which use that env var, globally setting `PATH` has implications for everything on the system. My personal preference would actually be for a project-specific env var like `POKECRYSTAL_RGBDS` but this isn't a hill i'm prepared to die on and just about anything is better than manipulating the system path.

mid-kid:
>globally setting PATH has implications for everything on the system

That's the thing, though. It doesn't, as long as the directory added to the path doesn't have any other executable files than the rgbds commands.

melthelesbian:
@mid-kid there's good reasons you'd have the tools installed in a directory that does have other executables though, see also:

> Setting the path will also break wrapper scripts that rely on the original commands being the ones in the path like the one i was using to test this worked properly

mid-kid:
Assuming your usual PATH is: `~/bin:/usr/local/bin:...etc`, what's the problem with setting `PATH="/usr/local/bin:$PATH"` (or even better, `/path/to/rgbds-0.3.7`, in case you have other tools in ~/bin that can be overridden with tools in /usr/local) in your .direnv? Assuming .direnv can restore the original PATH appropriately, that is.

meltheadorable:
@mid-kid because i have other things in `~/bin` including more duplicates of system-wide tools that I use for scripting/etc and which would now be loading the wrong thing in that project directory if it was swapped out. Changing PATH is only safe if the path you're prepending _only_ has stuff RGBDS stuff in it.

melthelesbian:
@mid-kid because i have other things in ~/bin including more duplicates of system-wide tools that I use for scripting/etc and which would now be loading the wrong thing in that project directory if it was swapped out. Changing PATH is only safe if the path you're prepending only has stuff RGBDS stuff in it.

mid-kid:
>Changing PATH is only safe if the path you're prepending only has RGBDS stuff in it.

And I guess that's pretty much what this entire conversation boils down to. I've seen close to no practical cases where that isn't true. And if it really isn't, it isn't very hard to make it be true (and probably better in the long run, anyway). But I guess it boils down to preference ¯\_(ツ)_/¯

PR #585, opened 2018-12-30
Refactor engine/phone/

Rangi42:
Split engine/phone/phone_scripts.asm and engine/phone/phone_gossip.asm into individual files for each NPC, like data/phone/text.


Rangi42:
Besides the various `UnknownScript` and `UnknownText`, the phones' `*Script1` should be `*CalleeScript` and `*Script2` should be `*CallerScript`. (I'm not taking care of that in this PR.)

mid-kid:
I just went and double-checked a couple of things. I went and back-traced some of beth's "callee" texts:
`UnknownText_0x1b54d4` (beth_callee.asm) -> `PhoneScript_GreetPhone_Female` (generic_callee.asm) -> `BeverlyPhoneScript2` (beverly.asm)
`UnknownText_0x1b5424` (beth_callee.asm) -> `PhoneScript_AnswerPhone_Female` (generic_callee.asm) -> `BeverlyPhoneScript1` (beverly.asm)

This means that both `generic_callee` and `beth_callee` are misnomers.
I'd call it `generic_callee.asm` `generic_greetings.asm` or answers or replies, since it only deals in those.

`PhoneScript_HangupText_Male` (generic_caller.asm) -> `ParryBattleWithMe` (hangups.asm) -> `ParryPhoneScript1` (parry.asm).
It seems `generic_caller` is a misnomer as well. Furthermore, most hangup texts are in the `_caller` files, such as `UnknownText_0x65471` in `chad_caller.asm`.
I really have no clue what to do with those.

It seems like the text files you named `_caller` deal in both caller and greetings, while `_callee` files deal in callees and hangups. That said, I haven't checked more than a couple of cases. I really have no clue what GF meant with this.

Rangi42:
`UnknownText_0x1b54d4` (beth_callee.asm) is used by `PhoneScript_GreetPhone_Female_Nite.Beth` (generic_callee.asm), not `BeverlyPhoneScript2`. Likewise `UnknownText_0x1b5424` (beth_callee.asm) is used by `PhoneScript_AnswerPhone_Female_Day.Beth` (generic_callee.asm).

Also, yeah, some of the scripts+text are not clearly defined as being "caller" or "callee"; like `PhoneScript_HangupText_Male` is used by \*_gossip.asm, generic_callee.asm, generic_caller.asm, hangups.asm, and hangups2.asm. But since the "caller" files do deal in part with caller-exclusive data, and "callee" with callee (even though they also deal with greetings and hangups respectively), can they stay as-is? The individual files are at least separated now, which should make it easier later for someone to identify the `Unknown` labels and maybe rename the files if they discover a better description.

mid-kid:
Yeah, those unknowntexts are used by `PhoneScript_*_Female_*`, but those are in turn used by `PhoneScript_*_Female` which are in turn used by the beverly scripts. I was showing the reference graph in reverse, skipping one small part for brevity's sake.

But yeah, I agree with them staying like this for the time being. It's a step up from the previous situation. Thanks for doing this.
Should we open an issue for [the renames you mentioned](https://github.com/pret/pokecrystal/pull/585#issuecomment-450602795)?

Rangi42:
How about an issue for labels in general? It would be a good place to start tracking the different patterns of undocumented labels. I'll go ahead and make the PhoneScript1/2 change before merging this.

PR #592, opened 2019-01-06
Update Route39Farmhouse.asm

Lenmaz:
Added symbolic name to milk price, just like it's done in Route43Gate.

### DIFF
--- a/maps/Route39Farmhouse.asm
+++ b/maps/Route39Farmhouse.asm
@@ -1,3 +1,5 @@
+ROUTE39FARM_HOUSE_PRICE EQU 500
+
 	const_def 2 ; object constants
 	const ROUTE39FARMHOUSE_POKEFAN_M
 	const ROUTE39FARMHOUSE_POKEFAN_F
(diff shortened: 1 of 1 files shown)


Rangi42:
Thank you for your PR; however, I'm not sure if it's a necessary change.

The `ROUTE43GATE_TOLL` constant gets used in six different places, and two of those are actually `ROUTE43GATE_TOLL - 1` (so having a constant there shows that the 1000 and 999 quantities are related).

Here, there are just two uses of `500`, right next to each other. Many maps do that sort of thing: the Game Corner prize prices, the Dept. Store vending machine prices, the haircut prices, etc. IMO defining constants for all of them would be overkill.

(Also, the name would be `ROUTE39FARMHOUSE_PRICE` to match the other constants.)

Then again, maybe all those local prices *should* have constants?

Lenmaz:
First of all, thanks for correcting me the constant name. I need to re-read the style guide.

In my mind we should eventually define constant for all the local prices in order to make the code more understandable. I guess it's not the most primordial task right now. You can do as you wish, though.

Rangi42:
I'm leaving this open, so feel free to expand it into adding constants for all the local prices. Although if you do, specific names would help, especially since some maps have multiple prices. Like, `ROUTE39FARMHOUSE_MOOMOO_MILK_PRICE` (or `ROUTE39FARMHOUSE_MILK_PRICE` if you prefer something shorter, up to you).

PR #596, opened 2019-01-31
Remove the nested UNION within battle WRAM

Rangi42:


FredrIQ:
The alternate arrangement sounds much more sensible to me.

PR #601, opened 2019-02-16
Exciting adventures down battle animation street!

mid-kid:


constants/battle_anim_constants.asm
```diff
@@ -25,6 +25,7 @@
 	const BATTLEANIMSTRUCT_16
 	const BATTLEANIMSTRUCT_17
 BATTLEANIMSTRUCT_LENGTH EQU const_value
+NUM_ANIM_OBJECTS EQUS "(wActiveAnimObjectsEnd - wActiveAnimObjects) / BATTLEANIMSTRUCT_LENGTH"
```
Rangi42: `NUM_OBJECT_STRUCTS`, `NUM_SPRITE_ANIM_STRUCTS`, `MONS_PER_BOX`, `NUM_HOF_TEAMS`, and most other constants just hard-code a number. I'm not sure which approach is better, but can they be made consistent, one way or the other?

engine/battle_anims/core.asm
```diff
@@ -80,32 +80,36 @@ endr
...
+	cp delanim_command
 	jp z, .delete
+
 	push af
 	ld hl, wBattleAnimTempOAMFlags
-	ld a, [wBattleAnimTempAddSubFlags]
+	ld a, [wBattleAnimTempFrameOAMFlags]
 	xor [hl]
-	and $e0
+	and $e0 ; Keep priority, x flip and y flip flags.
```
Rangi42: `and X_FLIP | Y_FLIP | PRIORITY`

engine/battle_anims/helpers.asm
```diff
@@ -37,16 +37,16 @@ GetBattleAnimFrame:
...
-	and $3f
+	and %00111111
 	ld hl, BATTLEANIMSTRUCT_DURATION
 	add hl, bc
 	ld [hl], a
 	pop hl
 .okay
 	ld a, [hl]
-	and $c0
+	and Y_FLIP << 1 | X_FLIP << 1 ; The << 1 is compensated in the "frame" macro
```
Rangi42: Please update the `and $c0 / srl a` in `GetSpriteAnimFrame` to also use shifted constants + explain this

Rangi42:
Please add the old names to macros/legacy.asm.

constants/battle_anim_constants.asm
mid-kid: The downside of this approach is more overhead for the linker. I'm not sure which is better, so I don't know, but thanks for pointing it out.

Rangi42:
Since this is battle anim related, I won't make a separate PR for this:

I think `ld hl, $ffa0` in `BattleAnimFunction_25` should be `lb hl, -1, $a0`. It's being used to decrement the `BATTLEANIMSTRUCT_XOFFSET` and add $a0 to whatever `BATTLEANIMSTRUCT_10` is.

PR #602, opened 2019-02-16
Update to rgbds 0.3.8

Rangi42:


engine/menus/naming_screen.asm
```diff
@@ -1,8 +1,8 @@
 NAMINGSCREEN_CURSOR     EQU $7e
 
-NAMINGSCREEN_BORDER     EQUS "\"■\"" ; $60
-NAMINGSCREEN_MIDDLELINE EQUS "\"→\"" ; $eb
-NAMINGSCREEN_UNDERLINE  EQUS "\"<DOT>\"" ; $f2
+NAMINGSCREEN_BORDER     EQU "■" ; $60
```
mid-kid: I mean, it's nice that these happen to use the same tiles as other characters, but they have nothing in common. Shouldn't these just use the raw values?
(This applies to all other tile IDs as well)

Rangi42: It's important because they overwrite those text tiles while other text tiles still get used; so if you move any charmap IDs around, you can accidentally make these screens overwrite the wrong character.

PR #606, opened 2019-03-03
Correct some bugfixes, add some compatibility extensions

ISSOtm:
The original fix was bugged (example: erroneous `pop hl` if `.failed` was jumped to).
Replaced with a less invasive fix, keeping all side effects from the original function; it's less intuitive, but more performant - and the tricky bit has been commented.
Also the new fix is more friendly to the compatibility patch I also added, by request of ShinyDragonHunter.

### DIFF
--- a/docs/bugs_and_glitches.md
+++ b/docs/bugs_and_glitches.md
@@ -168,7 +168,7 @@ This bug existed for all battles in Gold and Silver, and was only fixed for sing
 
 ## Moves with a 100% secondary effect chance will not trigger it in 1/256 uses
 
-*Fixing this bug will break compatibility with standard Pokémon Crystal for link battles.*
+*Fixing this bug **may** break compatibility with standard Pokémon Crystal for link battles.*
 
 ([Video](https://www.youtube.com/watch?v=mHkyO5T5wZU&t=206))
 
@@ -194,7 +194,7 @@ This bug existed for all battles in Gold and Silver, and was only fixed for sing
  	ret
 ```
 
-If you wish to keep compatibility with standard Pokémon Crystal, you can disable the fix during link battles by also applying the following edit in the same place:
+**Compatibility preservation:** If you wish to keep compatibility with standard Pokémon Crystal, you can disable the fix during link battles by also applying the following edit in the same place:
 
 ```diff
 +	ld a, [wLinkMode]
--- a/docs/bugs_and_glitches.md
+++ b/docs/bugs_and_glitches.md
@@ -182,8 +182,8 @@ This bug existed for all battles in Gold and Silver, and was only fixed for sing
 +	; Thus chance will be subtracted from 0, guaranteeing a carry
 +	call c, BattleRandom
  	cp [hl]
-	pop hl
-	ret c
+ 	pop hl
+ 	ret c
 
  .failed
  	ld a, 1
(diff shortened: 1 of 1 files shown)


docs/bugs_and_glitches.md
```diff
@@ -175,24 +175,38 @@ This bug existed for all battles in Gold and Silver, and was only fixed for sing
...
- 	call BattleRandom
++	sub 100 percent
++	; If chance was 100%, RNG won't be called (carry not set)
++	; Thus chance will be subtracted from 0, guaranteeing a carry
++	call c, BattleRandom
  	cp [hl]
--	pop hl
--	ret c
-+	jr c, .ok
+	pop hl
```
mid-kid: Watch the indentation.

docs/bugs_and_glitches.md
```diff
@@ -175,24 +175,38 @@ This bug existed for all battles in Gold and Silver, and was only fixed for sing
...
  .failed
  	ld a, 1
  	ld [wEffectFailed], a
  	and a
-+.ok
-+	pop hl
  	ret
 ```
 
+If you wish to keep compatibility with standard Pokémon Crystal, you can disable the fix during link battles by also applying the following edit in the same place:
```
mid-kid: This is something that is applicable to at least half of the bug fixes in this document. If you wish to document this, do it in a generic way, such that we do not have to repeat similar instructions for all of the bugs.

ISSOtm: That has already been done (see the top of the current document), but the problem is that it's really complicated for some of these functions which rely on flag behavior. SDH, who requested a review of their code, had trouble with exactly that, hence wanting to push a fix to master.
Further, following this reasoning, the entire document could be summarized as "the bugs can be solved by editing the code to remove the bugs", which wouldn't make sense.

docs/bugs_and_glitches.md
ISSOtm: The indentation [is present](https://github.com/pret/pokecrystal/blob/add58fb00d73495655243c4cf13b730bbc1b3242/docs/bugs_and_glitches.md#moves-with-a-100-secondary-effect-chance-will-not-trigger-it-in-1256-uses), this looks like GitHub's preview is broken.
The tabs are also present in [my fork's version](https://github.com/ISSOtm/pokecrystal/blob/patch-1/docs/bugs_and_glitches.md#moves-with-a-100-secondary-effect-chance-will-not-trigger-it-in-1256-uses) of the file.

mid-kid: It has to have a space preceeding the tab, due to how `diff` files work, and to be consistent with the rest of this document.

ISSOtm:
As I said in the PR's description, the previous fix was not fine, as [jumping to `.failed`](https://github.com/pret/pokecrystal/blob/add58fb00d73495655243c4cf13b730bbc1b3242/engine/battle/effect_commands.asm#L1870) would cause a `pop hl`  when it hasn't been pushed (line below). This would cause a crash.
I didn't re-do the fix out of optimization, I did it to permit the link-only fix as well as fixing the crashing bug.

docs/bugs_and_glitches.md
ISSOtm: Ah, sorry, I wasn't aware of that. I just pushed a commit that fixes that issue.

Rangi42:
Was the "erroneous `pop hl` if `.failed` was jumped to" the only bug with this? If so, then this would be a simpler fix that avoids use of `call c`:

```
	ld a, [hl]
	cp 100 percent
	jr z, .ok

	call BattleRandom
	cp [hl]
	pop hl
	ret c

.failed
	ld a, 1
	ld [wEffectFailed], a
	and a
	ret

.ok
	pop hl
	ret
```

Diff:

```diff
+	ld a, [hl]
+	cp 100 percent
+	jr z, .ok
+
-	; BUG: 1/256 chance to fail even for a 100% effect chance,
-	; since carry is not set if BattleRandom == [hl] == 255
 	call BattleRandom
 	cp [hl]
 	pop hl
 	ret c
 
 .failed
 	ld a, 1
 	ld [wEffectFailed], a
 	and a
 	ret
+
+.ok
+	pop hl
+	ret
```

ISSOtm:
An even simpler fix would be to move the `push hl` before the `jr nz, .failed`. The problem becomes writing a link-compatible fix.
Also, no, this wouldn't work, because the function then returns with carry clear - you'd need a `scf` at `.ok`, and now this is becoming a tad too complex for my tastes. The fix I proposed is as simple as the comment it contains - takes a second of thinking, but straightforward.

Rangi42:
You're right, it makes sense as-is, and it's actually less code than my attempt.

PR #607, opened 2019-03-03
Add compatibility extension for Counter + Item fix

ISSOtm:
Requested by ShinyDragonHunter.


Rangi42:
The document header already mentions that `LINK_COLOSSEUM` checks are necessary for link compatibility. This particular check is very straightforward and does not need to be explicitly spelled out; otherwise every battle engine bug would have three lines like those.

Issue #603, opened 2019-02-17
Two documented bugs need fixed code


Rangi42:
- [Confusion damage is affected by type-boosting items and Explosion/Self-Destruct doubling](https://github.com/pret/pokecrystal/blob/master/docs/bugs_and_glitches.md#confusion-damage-is-affected-by-type-boosting-items-and-explosionself-destruct-doubling)
- [A Pokémon that fainted from Pursuit will have its old status condition when revived](https://github.com/pret/pokecrystal/blob/master/docs/bugs_and_glitches.md#a-pok%C3%A9mon-that-fainted-from-pursuit-will-have-its-old-status-condition-when-revived)

PR #610, opened 2019-03-03
Fix Confusion and Pursuit bugs in #603

ISSOtm:
Took a moment to debate what would be the proper fix, but we decided that also handling happiness was the correct thing to do.

### DIFF
--- a/docs/bugs_and_glitches.md
+++ b/docs/bugs_and_glitches.md
@@ -394,17 +394,17 @@ Add this to the end of each file:
  	ld [wCryTracks], a
  	ld a, [wBattleMonSpecies]
  	call PlayStereoCry
-+    ld a, [wCurBattleMon]
-+    push af
++	ld a, [wCurBattleMon]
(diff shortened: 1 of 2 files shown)


docs/bugs_and_glitches.md
```diff
@@ -329,7 +387,28 @@ Add this to the end of each file:
...
 
-*To do:* Identify specific code causing this bug and fix it.
+**Fix:** Edit `PursuitSwitch` in [engine/battle/core.asm](/engine/battle/core.asm)
+
+```diff
+ 	ld a, $f0
+ 	ld [wCryTracks], a
+ 	ld a, [wBattleMonSpecies]
+ 	call PlayStereoCry
++    ld a, [wCurBattleMon]
```
mid-kid: Please fix the indentation.

ISSOtm: Forgot to push f6e94df, my bad ^^'

PR #614, opened 2019-03-11
common_1 text documentation

Fontbane:
Should be almost all of it. The only things I left unknown seemed to be either mobile link stuff or debug text.
Probably did something wrong here.


iimarckus:
Please squash the two commits so the VS stuff doesn’t show up in the history.

Rangi42:
Can this be merged? Are the labels reviewed?

PR #618, opened 2019-03-19
Adding labels to placeholders (UnknownText, MovementData) in the maps folder

DeeDeeG:
Quick work-in-progress PR for feedback.

Let me know here or in the Discord what you think. Glad to finish this file and other ones I can figure out as I get the time.

### DIFF
--- a/maps/RadioTower1F.asm
+++ b/maps/RadioTower1F.asm
@@ -3,8 +3,8 @@
 	const RADIOTOWER1F_LASS
 	const RADIOTOWER1F_YOUNGSTER
 	const RADIOTOWER1F_ROCKET
-	const RADIOTOWER1F_GENTLEMAN
-	const RADIOTOWER1F_COOLTRAINER_F
+	const RADIOTOWER1F_LUCKYNUMBERMAN
+	const RADIOTOWER1F_CARD_WOMAN
 
 RadioTower1F_MapScripts:
 	db 0 ; scene scripts
@@ -42,7 +42,7 @@ RadioTower1FLuckyNumberManScript:
 	writetext RadioTower1FLuckyNumberManThisWeeksIdIsText
 	buttonsound
 	closetext
-	applymovement RADIOTOWER1F_GENTLEMAN, MovementData_0x5ce71
+	applymovement RADIOTOWER1F_LUCKYNUMBERMAN, MovementData_0x5ce71
 	opentext
 	writetext RadioTower1FLuckyNumberManCheckIfMatchText
 	buttonsound
@@ -53,7 +53,7 @@ RadioTower1FLuckyNumberManScript:
 	buttonsound
 	special CheckForLuckyNumberWinners
 	closetext
-	applymovement RADIOTOWER1F_GENTLEMAN, MovementData_0x5ce74
+	applymovement RADIOTOWER1F_LUCKYNUMBERMAN, MovementData_0x5ce74
 	opentext
 	ifequal 1, .FirstPlace
 	ifequal 2, .SecondPlace
--- a/maps/RadioTower1F.asm
+++ b/maps/RadioTower1F.asm
@@ -42,7 +42,7 @@ RadioTower1FLuckyNumberManScript:
 	writetext RadioTower1FLuckyNumberManThisWeeksIdIsText
 	buttonsound
 	closetext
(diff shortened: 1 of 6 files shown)


mid-kid:
This looks good. All the advice I can give really is to maybe avoid including unnecessary information in the label's name, to make it shorter. For example `RadioTower1FLuckyNumberManYouHaveAPerfectMatchText` could simply be `RadioTower1FLuckyNumberManMatchText`, and `RadioTower1FReceptionistNotOfferingToursText`  could be `RadioTower1FReceptionistNoToursText`.
There's no limit really to the length of the labels but there's a point where too much information is just not beneficial. Try to condense it into one or a couple of words to simply be able to distinguish the labels and guess at the meaning, if someone wants to read the actual text he can just scroll down.

DeeDeeG:
I was going for the info should have a subject/verb/object, since we're used to reading those a lot IRL and our brain is fast at parsing that kind of info. But it is long.

I probably wouldn't want to do "[...]MatchText", since there's three different levels of matches with the Lucky Number Show. (For 5 digits of ID (perfect match), 3 digits (medium match) and 2 digits (weak match). )

I could see paring it down to "[...]PerfectMatchText", which is a bit shorter than "[...]YouHaveAPerfectMatchText".

But Subject Verb Object is probably too much info/too long?

DeeDeeG:
(Commits aren't in their final state, planning to clean it up a bit when this is ready to merge.)

maps/RadioTower1F.asm
```diff
@@ -1,4 +1,4 @@
-	const_def 2 ; object constants
+RadioTower1FRadioCardWomanOfferQuizText	const_def 2 ; object constants
```
mid-kid: This doesn't look right.

DeeDeeG: There's tons of copy-pasting, I must have hit Ctrl+V when I didn't mean to.

DeeDeeG: Thanks.

mid-kid:
Ah, makes sense. Well, I think such a format would work too, and the point of labels is to be readable to humans, so it's good. Whether you go for Subject/Verb/Object or Verb/Object, however, please try to keep it consistent.

DeeDeeG:
I've been addressing new labels until the file is done, and I plan to circle back to everything in commit one and re-do it.

maps/RadioTower1F.asm
DeeDeeG: (Actually I think I deleted a newline)

DeeDeeG: Nope, this definitely doesn't belong at the top of the file... Right the first time.

DeeDeeG:
This file is totally done now. :raised_hands:

DeeDeeG:
`RadioTower2F.asm` is done now.

maps/Route35NationalParkGate.asm
```diff
@@ -134,65 +134,65 @@ Route35NationalParkGate_LeaveTheRestBehind:
...
 	ifequal TRUE, Route35NationalParkGate_FirstMonIsEgg
-	writetext UnknownText_0x6a4c6
+	writetext Route35NationalParkGateOfficer1AskToUseFirstMonText
 	yesorno
 	iffalse Route35NationalParkGate_DeclinedToLeaveMonsBehind
 	special ContestDropOffMons
 	iftrue Route35NationalParkGate_FirstMonIsFainted
 	setevent EVENT_LEFT_MONS_WITH_CONTEST_OFFICER
-	writetext UnknownText_0x6a537
+	writetext Route35NationalParkGateOfficer1WellHoldYourMonText
```
DeeDeeG: Maybe use `WeWillHoldYourMons` instead of `WellHoldYourMons`? (as mentioned on Discord)

Rangi42:
Can this be merged? Are the labels reviewed?

DeeDeeG:
I don't think anyone has gone through and reviewed these yet.

If there's one or a few that  are objectionable, I'd be glad to fix them, or else someone can get around to improving them in a future PR...

If you want any of the commits squashed together or reorganized, I can do that as well.

Rangi42:
Don't worry about squashing, that can be done all at once when it's merged.

PR #623, opened 2019-04-10
Create some battle tower SRAM labels

mid-kid:


engine/events/battle_tower/battle_tower.asm
```diff
@@ -197,7 +199,7 @@ _BattleTowerBattle:
 	call .do_dw
 	call DelayFrame
 	ld a, [wBattleTowerBattleEnded]
-	cp $1
+	cp 1
```
Rangi42: `cp TRUE`

engine/battle_anims/anim_commands.asm
```diff
@@ -162,7 +162,7 @@ BattleAnimRestoreHuds:
 	pop af
 	ldh [rSVBK], a
 
-	ld a, $1
+	ld a, 1
```
Rangi42: The various loads/reads of `hBGMapMode` should use constants eventually, so might as well leave them as hex until then.

engine/events/battle_tower/battle_tower.asm
mid-kid: wInBattleTowerBattle is a weird one, it's used as bitflags (through the `bit` instruction) in a lot of places, so it's most definitely not a boolean.

Rangi42: Even more so if it's a bitflag, then please leave related values as hex until they have `*_F` constants identified.

Issue #626, opened 2019-04-14
i18n builds


Rangi42:
With a system like pokeruby has to keep different languages' data separate, so code files don't get cluttered with `if`/`else`s.

So far the only separate builds are USA/Europe Revision 1.1 and Australian (which just censors the Game Corner text from 1.1), but others are possible:

```
889a06fc0bb863666865aa69def0adf97945ac2a *pokecrystal-es.gbc
accb584293ba056152f1fd908439b019017ff2fe *pokecrystal-de.gbc
c055992b16b7399c687647725cdd1f4f13a2f75c *pokecrystal-fr.gbc
6cee05e5b95beeae74b8365ad18ec4a07a8c4af8 *pokecrystal-it.gbc
```

There's also `95127b901bbce2407daf43cce9f45d4c27ef635d *pokecrystal-jp.gbc`, but that probably has enough different code to justify a separate project, like pokegold.

aaaaaa123456789:
This is probably a hopeless endeavor.

When Ruby and friends were translated, they just had to replace strings. They relied on the compiler to place them in ROM, so only the strings themselves changed.

This is not the case for generations 1 and 2. Data was put into banks manually. Translations often caused banks to get full, leading to text (and maps, and everything) being moved around; that's the reason why `text_far` exists at all, for instance.

If nothing was moved across banks for the translations, this is possible, but I very much doubt it. If stuff was moved around, this will quickly turn into either `if` jungle or a pile of _ad hoc_ pseudo-metadata files just to handle the bank allocation.

Rangi42:
Maybe so. I'm leaving it open just like #285 for now, as a maybe-impossible goal. Even just separate repos for each language would be better than nothing.

mid-kid:
I'll just leave my notes on this endeavour behind, here:

First of all, pokeruby's way of doing this is terribly wasteful. It uses `rsync` to copy the `data_de` directory on top of the regular directory. If we're going to do this, we should do the following:
* Have the makefile pick up the dependencies for the proper language, for example through the [VPATH](https://www.gnu.org/software/make/manual/html_node/General-Search.html) variable, and make sure all generated files go to a proper subdirectory.
* Make sure you can build all of the languages with `make compare`
* Abuse rgbasm's `-i` option to "overlay" included files. This should also be able to be used to include generated files for the proper language.
I'm not entirely too sure what I'd make the directory structure look like, but I think top-level `i18n/<lang>` directories would be the simplest and most effective.

[...]

PR #631, opened 2019-04-26
Sectionful refactor

mid-kid:
Right now, the current disassembly splits sections arbitrarily. This means that
we have an odd combination of a linker script and `INCLUDE`s in several files
to sort everything in the ROM. Some sections even cause the game to break when
they're moved around to different banks, such as `bank10` (which contains
`engine/pokemon/evolve.asm`) and `Evolutions and Attacks`.  
To make it easier for a regular user to move things around to different banks
without headaches, I decided this should be fixed.

To achieve this, I plan to do the following:
1. Split everything with the sections as generated
2. Try splitting sections that span over multiple files if something inbetween is completely independent/unrelated
3. Move sections around to different files if necessary/practical
4. Merge whatever needs to be merged to avoid having too many pointless sections inside a single file
5. Remove main.asm and generate all object files separately with only the linker script to specify order. Rename included files to `.inc` and have the Makefile scan for `.asm` files.

This Pull Request is mostly to show progress, explain what I'm doing, and
discuss several decisions relating to this endeavour.

Current state
-------------

Everything I wanted to do has been done. I'll be reviewing some of the changes made, and so should you.


mid-kid:
How the section splitting was done
---
First of all, I'm establishing a definition for what a "section" is: A section is a self-contained blob of code or data that can move around the ROM banks without anything breaking.  

Given the current state of the disassembly, I believed this problem could be solved in a semi-objective manner by writing some tools to analyze the code. So, I wrote a couple of tools to do the following:
1. Generate a list of what "functions" (all labels bar local labels, excluding wram/sram) are referenced by each "function", and whether they're referenced through BANK() anywhere in that same "function". (this was done by parsing the compiled object files)
2. Group all "functions" that reference eachother without BANK().
3. Merge any groups that overlap anywhere in the ROM

Doing this gave me actually surprisingly decent results, but it requires working out of false-positives and other quirks. As such, every single one of these steps has an additional step to fix them. I won't go too much into detail about this here but you can read my `fix_*.py` scripts in the [splitting-tools](https://github.com/mid-kid/pokecrystal/tree/splitting-tools/splitting) branch.

The biggest flaw of this system is its inability to identify unrelated sections that are right between two other sections in the same ROM bank that relate to eachother. This caused some rather huge chunks of several banks to be a single section in the initial pass I did of this. I've done my best to manually work away the biggest examples of this, but it's by no means perfect, and as such, some currently unidentified areas of the code might need to be split further, later down the line.

mid-kid:
Current assembler-related quirks
---
To make sure some sections are in the same bank as others (due to non-BANK()ed references), I use the BANK[] attribute with a static bank. This could be done better if [`BANK["section"]`](https://github.com/rednex/rgbds/issues/244) were to be implemented.  
Some sections are only split to use a different charmap in places. This is because the assembler creates a new, clean, section-local charmap when you use `charmap` in a section. This charmap is then discarded when entering a new one. This could be done better if [either](https://github.com/rednex/rgbds/issues/256) of [these](https://github.com/rednex/rgbds/issues/265) proposals were to be implemented.  
I don't expect either of these to be fixed at the time of merging this PR, but they're things that would be good to fix in the long run.

mid-kid:
Uncertainities
---
To be able to get out more sections, I've split `Unused_PlaceEnemyHPLevel` away from `CopyMonToTempMon` and `DrawEnemyHP`. However, since this function is unused, and I doubt it even works as intended, it seems rather useless to make sure it shares the same bank as the other two functions. This is mentioned in a comment, but should the assembly make sure this is the case, anyway?

Some files contain a bunch of tangentially related functions, these include most prominently `engine/events/specials.asm` and `engine/events/std_scripts.asm`.  
These functions are only related to eachother through being specials/scripts, or simply being around the same area in the ROM. These can be handled in two ways: Either we leave them split, which clearly signals the user that the different functions may be moved around, and aren't really related to eachother. This would for example allow us to include all of the dummyspecials in `engine/events/specials.asm`, and do more such related reorganization.

I've tried splitting map scripts and songs on a per-unit basis, since each map and song has a separate file, anyway. However, there's a small problem with maps and songs referencing eachother. This happens in for example the rival theme, or the different ruins of alph chambers. Thankfully, all of the instances where this happens are right after eachother in the ROM, so there's no need to deal with splitting into separate sections. Now, there's multiple ways to deal with this:
- Include both in the same file/section.
- Rename one to `.inc` and include it with the `INCLUDE` keyword into the other.
- Keep both as separate `.asm` files and sections but make sure both sections are in the same bank.

I'm leaning towards the second option but this might mess with map editors that rely on everything being `.asm`. Though all of the options might mess with editors in different capacities.

Rangi42:
For maps or songs that reference each other, I'd recommend option 3, "Keep both as separate .asm files and sections but make sure both sections are in the same bank." Since making sure that multiple sections share a bank is something that needs doing elsewhere any (currently with `BANK[$xx]`, ideally with `BANK["Section Name"]`). Including one inside another would have to arbitrarily pick which one gets included.

mid-kid:
>Including one inside another would have to arbitrarily pick which one gets included.

While I agree this is the case with maps, it's actually rather obvious which one is the "parent" track with music. But I like consistency, so doing the same for maps as for audio tracks is probably the best idea.

mid-kid:
The case for select_menu.asm
---
`engine/pokemon/mon_menu.asm` is only ever used by `engine/menus/start_menu.asm`, so it makes sense to include them into eachother.  
However, this isn't really the case for `engine/overworld/select_menu.asm`. This file is only ever used from the overworld code, and does never touch any of the start menu code, nor the other way around.  
The problem lies, in that it's the only user of `CantUseItem`. This function is right smack in the middle of `mon_menu.asm`. I have no idea if this is yet another artifact caused by our beloved localization team, but that's the way things are.  
The ideal solution here would be to split `CantUseItem` off into its own section, that would then be included in `select_menu.asm`. However, to do so, you'd also have to split up at least three references across the function: `StartMenu_Quit -> StartMenuYesNo`, `StartMenu_Pokemon -> PokemonActionSubmenu`, `TossItemFromPC -> PartyMonItemName`. This would generate a bunch of sections that would depend on being located in the same bank, anyway.  
This all means that there's no natural way to split this up, and we'd end up with a mess of `SECTION`s and `BANK[]` attributes I'd rather not have. Hence, I'm probably not going to split apart select_menu.asm anytime soon. Maybe when we get a proper `BANK["Section"]` attribute.

mid-kid:
Reverted.

mid-kid:
Alright, so for the section names, I'm currently doing the following:
- If there's only one section in the file, the section's name is the filename.
- If there's multiple, it follows the `"path/to/file.asm@name"` format, where name is:
  - Nothing if this can be considered the "main" section in the file (for example when the only other section is a dummy predef)
  - A label name if there's only one label in the section
  - The "main" label's name if there's multiple and there's a clear function everything revolves around
  - A "generic" name if no label's name fits. This can be part of other labels (for example `FindPartyMon` in the section with `FindPartyMonAboveLevel`, `FindPartyMonThatSpecies`, etc...), or something completely custom.

Big exception: anything in `mobile/`. I won't be able to name its sections properly before it's disassembled further. I expect anyone disassembling it to do so accordingly. The placeholder names will largely be `"mobile/file.asm@StartLabel - EndLabel"`

mid-kid:
Splitting the source into separate translation units (.o files)
---

First of all, let's start with why we'd do this. With the section splitting done, this just feels like a logical next step, because:
- We'll only need the linker script to order things in the ROM
- You can avoid namespace conflicts in different parts of the code/see what is called from different units, and what is local.
- It'll improve incremental build times.

However, as with anything, it doesn't come without tradeoffs. Building so many different units of code separately will increase full rebuild times. To mitigate this, I've had to make one other difficult decision: To only include the constants/macro files a certain unit needs. This was necessary because the biggest time-waster in rebuilding all of the separate modules was including everything with `constants.asm`. The rgbds PR https://github.com/rednex/rgbds/pull/337 also helps with improving the full rebuild times.  
To help people use the codebase in such a setup, `constants/maps_common.inc` and `constants/audio_common.inc` were created, and some include files include eachother where necessary, to save the user from having to know the inter-file dependencies by heart.

mid-kid:
The new makefile
---

In an effort of making https://github.com/pret/pokecrystal/issues/626 possible in a way that doesn't interfere with the development of ROM hacks (i.e. no superfluous `IF DEF()`s everywhere, can remove the translation code with a simple `rm -rf`), I've revamped the Makefile. Its features include:
- It builds everything in a build directory
- It allows building multiple different versions
- It allows "overlaying" the source directory with a version-specific directory

To do this, [this method](http://make.mad-scientist.net/papers/multi-architecture-builds/) is employed, in which the makefile will re-execute itself inside of a build directory, and use VPATH to locate the sources in different directories. This facilitates implementing an "overlay"-style system.

The biggest drawback of this system is the inability to reuse parts of the build from different versions. Since every version is built in a separate directory by a separate instance of `make`, there's no way for these to communicate in this setup. There's probably ways to achieve this, but they would most probably require manual makefile modifications, which is undesireable in this project. As such, for every version, _everything_ will be rebuilt, including graphics. I don't see this as a big issue, since most people will only want to build one of the versions.

The new Makefile works pretty much the same as the old one from the user-side, with the addition of the `version=` parameter, to specify what version to build a specific (set of) target(s) for (not needed when using the "version" targets `crystal11` or `crystal-au`). From the developer side, you might need to scroll a bit further down to reach the actual rules used for building the ROM, and depending on what version you're working on, you might have to check `version/<version>.mk`.

mid-kid:
And with that, this PR is done and open for review.

Rangi42:
Since "We'll only need the linker script to order things in the ROM" is an explicit goal of this PR, it would be nice if home.asm were eliminated the same way as main.asm.

The main difficulty with doing so is that ROM0 doesn't neatly break into unrelated pieces, and to organize it well, we'd have to pretty much inspect every home routine, decide which ones ought to go together, and group their sections together in single files.

However, I think that can wait for another PR, and for this one, it would be enough to turn each file in home/ into its own section, plus a few more for the remaining routines in home.asm.

gfx/tileset_palette_maps.inc
```diff
@@ -0,0 +1,115 @@
+tilepal: MACRO
+; used in gfx/tilesets/*_palette_map.asm
```
Rangi42: They're gfx/tilesets/\*_palette_map.inc now.

INSTALL.md
```diff
@@ -181,11 +181,11 @@ cd pokecrystal
 To build **pokecrystal.gbc**:
 
 ```bash
-make
+make -j4
```
Rangi42: How about `make -j\`nproc\``?

Rangi42: (Or `make -j$(nproc)`, either works.)

mid-kid: I was just following the precedent set by pokeruby, and this is a sane, easy to remember default. How about adding [this section](https://github.com/pret/pokeruby/blob/master/INSTALL.md#parallel-build) or at least linking to it, instead?

Rangi42: Yeah, I guess `-j4` is a good compromise between minimal-but-slow `make` and optimal-but-long `-j$(nproc)`. Please do add a section explaining the multiple and parallel builds like pokeruby has.

Rangi42:
@mid-kid Can you rebase this, and possibly eliminate home.asm like main.asm? Then I guess this will be ready to become the main branch, with the current one preserved as `classic` or something.

PikalaxALT:
Is there any real interest in refactoring wram.asm and/or sram.asm in the same way?

mid-kid:
@Rangi42 Sorry, I've been a bit busy lately. I will do that, and I also wanted to write a little bit of documentation for the new makefile, towards the end of the month.

@PikalaxALT Eventually, I hope so. Those two files (especially wram) are almost unreadable and unmanageable, and it'd be nice to see them be cleaned up like this in the future. However, I won't tackle this problem anytime soon, so if anyone else is up for it, please do.

gfx/tileset_palette_maps.inc
coltongit: This is fixed already

INSTALL.md
coltongit: This depends on the amount of processors your CPU has. Would be more useful to go the nproc route imo

coltongit:
Can this be rebased if at all possible?

Rangi42:
charmap.inc would make sense in constants/.

Rangi42:
Is engine/crystal.asm necessary? It's just grouping some .incs related only by their unique Crystal code; main_menu is mostly G/S content anyway; and a few files that come after it in the same bank are already standalone (engine/pokemon/search.asm, mobile/mobile_12_2.asm, engine/events/buena_menu.asm).

mid-kid:
BANK_CRYSTAL
==============

Bank $12 is a bit entangled, having been used mostly for crystal-related things. There's 5 files with a bunch of references across eachother:
- `mobile/mobile_22`
- `engine/menus/init_gender`
- `engine/gfx/crystal_layouts`
- `engine/menus/main_menu`
- `mobile/mobile_menu`

I've merged the first two, and the last two, and made 3 separate sections of these, to "liberate" the couple of files in-between. These files should be looked at in the future, especially `mobile_22`, and the rest among them, to see whether they can be split further.

Rangi42:
Issue: a depfile won't be updated if an included file updates.

So if an .asm file `INCLUDE`s an .inc file which `INCBIN`s a .2bpp file, and then you only change the .2bpp file, `make` won't update the ROM.

PikalaxALT:
Is this PR still active?

mid-kid:
Yes, I intend on merging parts of this PR upstream first, however, since that'll allow separate review of those changes.

mid-kid:
Closing for the time being, as mentioned in a previous comment, this should be split up.

PR #640, opened 2019-07-08
Adding names to pointers in common text

nephitejnf:
Just finished up going through most of the common text pointers giving out more descriptive names. I left many of the unused pointer be.


nephitejnf:
Forgot to push fixes

PR #641, opened 2019-07-08
Adding names to pointers in common text

nephitejnf:
Just finished up going through most of the common text pointers giving out more descriptive names. I left many of the unused pointer be.

Has fixes that weren't present in the first request.

### DIFF
--- a/data/text/common_1.asm
+++ b/data/text/common_1.asm
@@ -260,6 +260,8 @@ UnknownText_0x1bc44c::
 	done
 
 ; last Unused Mobile Adapter
+
+
 UnknownText_0x1bc45e::
 	text "Enter the"
 	line "amount."
--- a/data/text/common_1.asm
+++ b/data/text/common_1.asm
@@ -213,6 +213,7 @@ _OakTimeText14::
 	text ", is it?"
 	done
 
+; unused Mobile Text
 UnknownText_0x1bc384::
 	text "There is nothing"
 	line "connected."
@@ -258,6 +259,7 @@ UnknownText_0x1bc44c::
 	line "ID no."
 	done
 
+; last Unused Mobile Adapter
 UnknownText_0x1bc45e::
 	text "Enter the"
 	line "amount."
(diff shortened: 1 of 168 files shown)


PikalaxALT:
You could have just pushed the fixes to the original pr instead of opening a new one.

PikalaxALT:
Also your build failed. Please run `make compare` and achieve OK on all three targets before opening your pull request.

mid-kid:
Be sure to follow our code cleanup guidelines, specifically the part about text_far labels: https://github.com/pret/pokecrystal/wiki/Code-cleanup#text-label-styles-in-datatext

data/text/common_1.asm
```diff
@@ -258,6 +259,7 @@ UnknownText_0x1bc44c::
 	line "ID no."
 	done
 
+; last Unused Mobile Adapter
```
mid-kid: I'd put this as its own line, surrounded by two blank ones, after this text. `end of unused Mobile Text`. Surround the starting comment with blank lines as well, mostly for consistency with similar comments further down this file.

engine/events/overworld.asm
```diff
@@ -323,7 +323,7 @@ Script_UseFlash:
 	end
 
 UnknownText_0xc8f3:
-	text_far UnknownText_0x1c0609
+	text_far Text_UseFlash
```
mid-kid: Also make sure the text_far labels use the same name as their parent text script counterpart.

nephitejnf: Thanks for pointing me to the clean up, couldn't quite remember which document pointed that out, I'll get to work on cleaning it up properly. And my main focus on this part was going through the common text, but getting that out of the way is definitely in the plans.

mid-kid: Considering these are simply pointers to the common text it's probably easier to do it right now than have to clean it up later, but if you wish to leave this for a future PR, I guess that's fine.

nephitejnf: To be honest, most things outside of the common text asm I wasn't aware of, because I was using `sed` and `grep` heavily.

nephitejnf:
I think that is all the used pointers, now.

Rangi42:
Superseded by #646 .

PR #652, opened 2019-09-23
Clarify naming of SFX_GET_EGG_* pointers

rtandy:
In game, one receives eggs from the Aide and the Day-care Man, not the Day-care Lady. SFX `0x96` is used for both. Rename it to reflect it is not unique to the Day-care.

SFX `0x95` is an alias of the same sound and appears to be unused. Rename the pointer to clarify that.

### DIFF
--- a/audio/sfx.asm
+++ b/audio/sfx.asm
@@ -1060,15 +1060,13 @@ Sfx_3RdPlace_Ch7:
 
 	togglesfx
 
-Sfx_GetEggFromDayCareLady:
-Sfx_GetEggFromDayCareMan:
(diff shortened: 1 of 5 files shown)


rtandy:
gold edition: https://github.com/pret/pokegold/pull/12

Rangi42:
This will conflict with #650 once that's merged, but we can fix that. This change looks good.

dannye:
> This will conflict with #650 once that's merged, but we can fix that. This change looks good.

Easy fix on my part, just changed one line of python :grin:

PR #659, opened 2019-11-18
Fixed a charmap warning with the new RGBDS

coltongit:
Fixed the charmap warning issue as per https://github.com/aaaaaa123456789/pokecrystal16/commit/bb7b22e0dfddaa4def206982e5e151fa284a56ee


mobile/mobile_46.asm
```diff
@@ -1484,7 +1484,8 @@ Function118b9a:
 	and a
 	ret
 
-	charmap " ", $20 ; revert to ascii
+	pushc
+	setcharmap ascii ; revert to ascii
```
mid-kid: This use of setcharmap should end after the URLs.

mobile/mobile_46.asm
```diff
@@ -2927,6 +2928,7 @@ Unknown_1196b8:
 	db "Sat"
 	db "Sun"
 
+popc
```
mid-kid: You should wrap the weekdays here in a setcharmap instead of covering this entire area.

mid-kid:
It also seems the travis build has failed. Run `make compare` before submitting your PRs to make sure the ROM still matches, and run a binary diff to see what you accidentally changed.

mobile/mobile_46.asm
coltongit: Resolved in latest commit.

mobile/mobile_46.asm
coltongit: Resolved in latest commit.

PR #660, opened 2019-11-18
Make the map setup commands match the actual function names better

### LINKED ISSUE #655: Map/object/event cleanup
Points carried over from #639 :

- Map setup command constants vs. functions
- `map_warp_face` vs. `RestoreFacingAfterWarp` are misnomers
- `object_struct` constants vs. WRAM labels
- `*MapObject*` -> `*ObjectEvent*`

mid-kid:
These functions used as map setup commands are used in other places, too,
so I can't prefix them under the same. The names should match except I
won't repeat "map" in a map setup command name.

Partially fixes #655


Rangi42:
Leaving this open in case you address more of #655.

Rangi42:
Should I go ahead and merge this as-is?

mid-kid:
Yeah sorry I won't get around to more anytime soon.

PR #661, opened 2019-11-18
Fixed a charmap warning with the new RGBDS

mid-kid:
Supersedes #659


Rangi42:
Since this requires RGBDS 0.3.9, we should update INSTALL.md and FAQ.md to specify that version.

PR #664, opened 2019-11-28
Add and simplify bugfixes in bugs_and_glitches.md

coltongit:
Simplified the `BattleCheckTypeMatchup` fix as well as added a fix for the Full Heal not curing AI confusion. Also made fixes for virtually every glitch known to occur with Crystal, which also fixes issue #666 as a result.

### DIFF
--- a/docs/bugs_and_glitches.md
+++ b/docs/bugs_and_glitches.md
@@ -14,7 +14,7 @@ Some fixes are mentioned as breaking compatibility with link battles. This can b
 
 
 ## Contents
-
+- [Berserk Gene's confusion isn't permanent](#berserk-genes-confusion-isnt-permanent)
 - [Thick Club and Light Ball can make (Special) Attack wrap around above 1024](#thick-club-and-light-ball-can-make-special-attack-wrap-around-above-1024)
 - [Metal Powder can increase damage taken with boosted (Special) Defense](#metal-powder-can-increase-damage-taken-with-boosted-special-defense)
 - [Reflect and Light Screen can make (Special) Defense wrap around above 1024](#reflect-and-light-screen-can-make-special-defense-wrap-around-above-1024)
@@ -75,6 +75,14 @@ Some fixes are mentioned as breaking compatibility with link battles. This can b
 - [`ClearWRAM` only clears WRAM bank 1](#clearwram-only-clears-wram-bank-1)
 - [`BattleAnimCmd_ClearObjs` only clears the first 6⅔ objects](#battleanimcmd_clearobjs-only-clears-the-first-6-objects)
 
+
+## Berserk Gene's confusion isn't permanent
+
+*Fixing this bug will break compatibility with standard Pokémon Crystal for link battles.*
+
+([Video](https://youtube.com/watch?v=Pru3mohq20A))
+
+**Fix:** TBD
 
 ## Thick Club and Light Ball can make (Special) Attack wrap around above 1024
 
--- a/docs/bugs_and_glitches.md
+++ b/docs/bugs_and_glitches.md
@@ -767,7 +767,7 @@ This bug existed for all battles in Gold and Silver, and was only fixed for sing
 
 ([Video]())
 
-**Fix:** Edit `EnemyUsedFullRestore` and `AI_HealStatus` in [engine/battle/ai/items.asm](https://github.com/pret/pokecrystal/blob/master/engine/battle/ai/items.asm):
+**Fix:** Edit `EnemyUsedFullRestore`, `EnemyUsedFullHeal`, and `AI_HealStatus` in [engine/battle/ai/items.asm](https://github.com/pret/pokecrystal/blob/master/engine/battle/ai/items.asm):
 
 ```diff
 EnemyUsedFullRestore:
@@ -778,6 +778,17 @@ EnemyUsedFullRestore:
 -	res SUBSTATUS_CONFUSED, [hl]
 	xor a
 	ld [wEnemyConfuseCount], a
+```
+
+```diff
+EnemyUsedFullHeal:
+	call AIUsedItemSound
+	call AI_HealStatus
+	ld a, FULL_HEAL
++	ld [wCurEnemyItem], a
++	xor a
++	ld [wEnemyConfuseCount], a
+	jp PrintText_UsedItemOn_AND_AIUpdateHUD
 ```
 
 ```diff
--- a/docs/bugs_and_glitches.md
+++ b/docs/bugs_and_glitches.md
@@ -86,33 +86,22 @@ Some fixes are mentioned as breaking compatibility with link battles. This can b
 **Fix:** Edit `HandleBerserkGene` in [engine/battle/effect_commands.asm](https://github.com/pret/pokecrystal/blob/master/engine/battle/effect_commands.asm)
 
 ```diff
-	ld h, d
-    	ld l, e
-    	ld a, b
-    	call GetPartyLocation
-    	xor a
-    	ld [hl], a
-    	ld a, BATTLE_VARS_SUBSTATUS3
-    	call GetBattleVarAddr
-    	push af
-+    	push bc
-+    	ld bc, wEnemyConfuseCount
-+    	ldh a, [hBattleTurn]
-+    	and a
-+    	jr z, .got_confuse_count
-+    	ld bc, wPlayerConfuseCount
-+
-+.got_confuse_count
-    	set SUBSTATUS_CONFUSED, [hl]
-+    	; confused for 2-5 turns
-+    	call BattleRandom
-+    	and %11
-+    	inc a
-+    	inc a
-+    	ld [bc], a
-+    	pop bc
-    	ld a, BATTLE_VARS_MOVE_ANIM
-    	call GetBattleVarAddr
+     ld a, BATTLE_VARS_SUBSTATUS3
+     call GetBattleVarAddr
+     push af
+     set SUBSTATUS_CONFUSED, [hl]
++    ld a, [hBattleTurn]
++    and a
++    ld hl, wEnemyConfuseCount
++    jr z, .set_confuse_count
++    ld hl, wPlayerConfuseCount
++.set_confuse_count
++    call BattleRandom
++    and %11
++    add a, 2
++    ld [hl], a
+     ld a, BATTLE_VARS_MOVE_ANIM
+     call GetBattleVarAddr
 ```
 
 
--- a/docs/bugs_and_glitches.md
+++ b/docs/bugs_and_glitches.md
@@ -1853,8 +1853,9 @@ This supports up to six entries.
  	ld de, 3
  	ld hl, .pointers
  	call IsInArray
- 	jr nc, .nope
+- 	jr nc, .nope
  	pop bc
++	jr nc, .nope
 
  	inc hl
  	ld a, [hli]
@@ -1864,7 +1865,6 @@ This supports up to six entries.
 
  .nope
 -	; pop bc
-+	pop bc
  	xor a
  	ret
 ```
--- a/docs/bugs_and_glitches.md
+++ b/docs/bugs_and_glitches.md
@@ -696,7 +696,7 @@ This bug existed for all battles in Gold and Silver, and was only fixed for sing
 
 ([Video](https://www.youtube.com/watch?v=rGqu3d3pdok&t=322))
 
-**Fix:** Edit `AI_HealStatus` and `EnemyUsedFullRestore` in [engine/battle/ai/items.asm](https://github.com/pret/pokecrystal/blob/master/engine/battle/ai/items.asm):
+**Fix:** Edit `AI_HealStatus` in [engine/battle/ai/items.asm](https://github.com/pret/pokecrystal/blob/master/engine/battle/ai/items.asm):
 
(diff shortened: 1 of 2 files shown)


coltongit:
The fix itself has no video as of yet, however the video will be added before this gets merged.

docs/bugs_and_glitches.md
```diff
@@ -720,6 +728,45 @@ This bug existed for all battles in Gold and Silver, and was only fixed for sing
 ```
 
 
+## NPC use of Full Heal does not cure confusion status
+
+([Video]())
+
+**Fix:** Edit `AI_HealStatus` and `EnemyUsedFullRestore` in [engine/battle/ai/items.asm](https://github.com/pret/pokecrystal/blob/master/engine/battle/ai/items.asm):
```
Rangi42: Saying "`EnemyUsedFullRestore` and `AI_HealStatus`" would match their order in the file. Also, to keep the comments consistent, either remove the whole "Bug:" comment or change it to address both:

```
	; Bug: this should reset SUBSTATUS_NIGHTMARE and SUBSTATUS_CONFUSED too
	; Uncomment the lines below to fix
	; ld hl, wEnemySubStatus1
	; res SUBSTATUS_NIGHTMARE, [hl]
	; ld hl, wEnemySubStatus3
	; res SUBSTATUS_CONFUSED, [hl]
```

coltongit: Will do the latter to make it easier to apply the fix. Duly noted.

docs/bugs_and_glitches.md
```diff
@@ -14,7 +14,7 @@ Some fixes are mentioned as breaking compatibility with link battles. This can b
 
 
 ## Contents
-
+- [Berserk Gene's confusion isn't permanent](#berserk-genes-confusion-isnt-permanent)
```
Rangi42: Keep a blank line between the heading and the list. Also, if this is only considered a bug due to Stadium 2 changing it, the bug should be that it lasts for 256 turns and not the standard duration.

Also, this seems like a more unambiguous bug: "If a confused Pokémon is switched out and replaced by a Pokémon holding a Berserk Gene, the holder will be confused for the remaining duration of the first Pokémon's confusion, rather than for 256 turns."

docs/bugs_and_glitches.md
coltongit: Applied in the latest commit.

docs/bugs_and_glitches.md
coltongit: Fixed in the latest commit.

docs/bugs_and_glitches.md
```diff
@@ -708,12 +717,57 @@ This bug existed for all battles in Gold and Silver, and was only fixed for sing
...
+```diff
+ AI_HealStatus:
+ 	ld a, [wCurOTMon]
+ 	ld hl, wOTPartyMon1Status
+ 	ld bc, PARTYMON_STRUCT_LENGTH
+ 	call AddNTimes
+ 	xor a
+ 	ld [hl], a
+ 	ld [wEnemyMonStatus], a
+	; Bug: this should reset SUBSTATUS_NIGHTMARE
```
Rangi42: `-` before the bug comment.

docs/bugs_and_glitches.md
```diff
@@ -708,12 +717,57 @@ This bug existed for all battles in Gold and Silver, and was only fixed for sing
...
+```
+
+
+## NPC use of Full Heal does not cure confusion status
+
+([Video]())
+
+**Fix:** Edit `EnemyUsedFullRestore` and `AI_HealStatus` in [engine/battle/ai/items.asm](https://github.com/pret/pokecrystal/blob/master/engine/battle/ai/items.asm):
+
+
```
Rangi42: Extra newline here

Rangi42:
Looks pretty good, thanks! Leaving it open for now until the video link and TBD fix.

docs/bugs_and_glitches.md
coltongit: Fixed.

docs/bugs_and_glitches.md
coltongit: Fixed.

coltongit:
You're welcome. Surely this will help get everything sorted concerning all the bugs/glitches

engine/battle/ai/items.asm
```diff
@@ -725,10 +725,14 @@ AI_HealStatus:
 	xor a
 	ld [hl], a
 	ld [wEnemyMonStatus], a
-	; Bug: this should reset SUBSTATUS_NIGHTMARE too
-	; Uncomment the lines below to fix
+	; Bug: this should reset SUBSTATUS_NIGHTMARE
+	; Uncomment the former 2 lines below to fix
+	; This should also reset SUBSTATUS_CONFUSED
+	; Uncomment the latter 2 lines below to fix
```
ISSOtm: Why not just put each comment block before each code block?

coltongit: I can see your point there. It'd make more sense for the beginner the way you're thinking of.

ISSOtm: LGTM

docs/bugs_and_glitches.md
```diff
@@ -75,6 +77,45 @@ Some fixes are mentioned as breaking compatibility with link battles. This can b
 - [`BattleAnimCmd_ClearObjs` only clears the first 6⅔ objects](#battleanimcmd_clearobjs-only-clears-the-first-6-objects)
 
 
+## Berserk Gene's confusion lasts for 256 turns
+
+*Fixing this bug will break compatibility with standard Pokémon Crystal for link battles.*
+
+([Video](https://youtube.com/watch?v=Pru3mohq20A))
+
+**Fix:** Edit `HandleBerserkGene` in [engine/battle/effect_commands.asm](https://github.com/pret/pokecrystal/blob/master/engine/battle/effect_commands.asm)
```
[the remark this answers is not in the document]
coltongit: Already fixed.

engine/battle/ai/items.asm
coltongit: Fixed.

docs/bugs_and_glitches.md
```diff
@@ -708,12 +736,69 @@ This bug existed for all battles in Gold and Silver, and was only fixed for sing
...
+	; res SUBSTATUS_CONFUSED, [hl]
+ 	ld hl, wEnemySubStatus5
+ 	res SUBSTATUS_TOXIC, [hl]
+ 	ret
+```
+
+
+## NPC use of Full Heal does not cure confusion status
+
+([Video]())
```
Rangi42: If there's no video available, just remove this.

engine/battle/ai/items.asm
```diff
@@ -725,10 +725,14 @@ AI_HealStatus:
 	xor a
 	ld [hl], a
 	ld [wEnemyMonStatus], a
-	; Bug: this should reset SUBSTATUS_NIGHTMARE too
-	; Uncomment the lines below to fix
+	; Bug: this should reset SUBSTATUS_NIGHTMARE
+	; Uncomment the 2 lines below to fix
 	; ld hl, wEnemySubStatus1
 	; res SUBSTATUS_NIGHTMARE, [hl]
+	; This should also reset SUBSTATUS_CONFUSED
```
Rangi42: "Bug: This should reset SUBSTATUS_CONFUSED"

docs/bugs_and_glitches.md
coltongit: I'm in the process of having one made for this bug. It won't be too long before it finds it's way here.

Rangi42:
Bulbapedia claims "If a confused Pokémon is switched out and replaced by a Pokémon holding a Berserk Gene, the holder will be confused for the remaining duration of the first Pokémon's confusion, rather than for 256 turns." It appears that your fix also resolves that bug; so can you rename it to "Berserk Gene's confusion lasts for 256 turns or the previous Pokémon's confusion count", and note that "This fix will make it last for the standard 2−5 turns"?

engine/battle/ai/items.asm
coltongit: Fixed already.

coltongit:
Resolved.

coltongit:
Figured I'd try to add a fix for the save glitches, as these usually tend to cause issues when left unchecked.

docs/bugs_and_glitches.md
```diff
@@ -391,6 +392,104 @@ Then edit four routines in [engine/battle/effect_commands.asm](https://github.co
 ```
 
 
+## Saves can be corrupted in such a way as to glitch the game
```
Rangi42: How about "Turning off the power during a save can corrupt the game"?

ISSOtm: Well, that's not a problem that can be corrected, since the save game write is not and cannot be atomic. How about "Saves corrupted [by power-off mid-saving] are not handled"?

coltongit: Resolved.

docs/bugs_and_glitches.md
coltongit: Added a TBA field for this for now. This will have to hold out until a video for it actually surfaces.

docs/bugs_and_glitches.md
```diff
@@ -361,6 +457,108 @@ Then edit four routines in [engine/battle/effect_commands.asm](https://github.co
...
+	call CloseSRAM
+	ret
+
++InvalidateSave:
++	ld a, BANK(sCheckValue1) ; aka BANK(sCheckValue2)
++	call GetSRAMBank
++	xor a
++	ld [sCheckValue1], a
++	ld [sCheckValue2], a
++	call CloseSRAM
```
ISSOtm: Why was the tail call removed?

docs/bugs_and_glitches.md
```diff
@@ -361,6 +457,108 @@ Then edit four routines in [engine/battle/effect_commands.asm](https://github.co
...
+	call CloseSRAM
+	ret
+
++InvalidateBackupSave:
++	ld a, BANK(sBackupCheckValue1) ; aka BANK(sBackupCheckValue2)
++	call GetSRAMBank
++	xor a
++	ld [sBackupCheckValue1], a
++	ld [sBackupCheckValue2], a
++	call CloseSRAM
```
ISSOtm: (Same here)

docs/bugs_and_glitches.md
coltongit: This was due to the fact that the original code had just the `call CloseSRAM`+`ret` instead of the tail call. It doesn't make a difference as far as functionality is concerned.

docs/bugs_and_glitches.md
coltongit: This was due to the fact that the original code had just the `call CloseSRAM`+`ret` instead of the tail call. It doesn't make a difference as far as functionality is concerned.

docs/bugs_and_glitches.md
coltongit: Due to the fact that the original code had it this way and that the document uses fixes that work for the original code, which didn't have the tail call in the first place.

docs/bugs_and_glitches.md
coltongit: Due to the fact that the original code had it this way and that the document uses fixes that work for the original code, which didn't have the tail call in the first place.

docs/bugs_and_glitches.md
ISSOtm: Since this is 100% original code, I'm not sure the point makes sense. Plus, as many users seem to be struggling against lack of space, it sounds like a better idea not to worsen the problem.

coltongit: I've checked again and it seems that you were right to begin with. Will resolve shortly.

docs/bugs_and_glitches.md
coltongit: Never mind. I looked at the source again and it had the tail call to begin with. Resolved.

docs/bugs_and_glitches.md
```diff
@@ -80,13 +81,61 @@ Some fixes are mentioned as breaking compatibility with link battles. This can b
 - [`BattleAnimCmd_ClearObjs` only clears the first 6⅔ objects](#battleanimcmd_clearobjs-only-clears-the-first-6-objects)
 
 
+## The S.S. Aqua and the ports have you white out to your cabin or either the Vermilion or Olivine Pokémon Centers even if you saved somewhere else
```
Rangi42: How is this a bug or glitch? The `blackoutmod` is intentional. I expect it's either for convenience so you can quickly get back on the ship if you faint on it, or to prevent sequence-breaking or getting stuck somehow on the first trip to Kanto.

coltongit: HGSS doesn't have this, but I agree with what you're saying here. This one may be more suited to the wiki than anything else, so it's probably a good idea to revert this one.

coltongit: Moved to the wiki.

docs/bugs_and_glitches.md
coltongit: This has been marked as resolved due to the file now having the video portion removed.

Rangi42: Correct. It's not essential to merging.

PR #665, opened 2019-12-03
Improve "Beat Up may fail to raise Substitute" fix

ISSOtm:
Improve the standard fix, and introduce a lightweight patch to preserve link compatibility

### DIFF
--- a/docs/bugs_and_glitches.md
+++ b/docs/bugs_and_glitches.md
@@ -579,9 +579,29 @@ This bug prevents Substitute from being raised if Beat Up was blocked by Protect
  	jr z, .multihit
  	cp EFFECT_POISON_MULTI_HIT
  	jr z, .multihit
+-	jp EndMoveEffect
 +	cp EFFECT_BEAT_UP
-+	jr z, .multihit
++	jp nz, EndMoveEffect
+
+ .multihit
+ 	call BattleCommand_RaiseSub
  	jp EndMoveEffect
+```
+
+**Compatibility preservation:** If you wish to keep compatibility with standard Pokémon Crystal, you can disable the fix during link battles by also applying the following edit in the same place:
+
+```diff
+ 	cp EFFECT_MULTI_HIT
+ 	jr z, .multihit
+ 	cp EFFECT_DOUBLE_HIT
+ 	jr z, .multihit
+ 	cp EFFECT_POISON_MULTI_HIT
+ 	jr z, .multihit
+ 	cp EFFECT_BEAT_UP
+ 	jp nz, EndMoveEffect
++	ld a, [wLinkMode]
++	cp LINK_COLOSSEUM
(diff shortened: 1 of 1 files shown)


mid-kid:
Sorry, but two things:

We don't document compatibility-preserving fixes when it's just "if multiplayer excert regular behavior", since that'd bloat the file up a ton. Our only exception to this right now is the "beat up with only one pokemon" fix, which documents how to at least fix how substitute appears on screen.

We try to write fixes in the same way that GF would've done it if they had found the bug, since trying to apply our own, more optimal style is a slippery slope that'll cause more bloat in the file than good. Further, a case can be made for this style, as you can more easily tell someone to add two extra lines to that function in multiple fixes separately, without having to explain the difference between `jr z,` and `jr nz,` to the naïve user.

I recall explaining this before but I can't really find the issue or PR :P

Leaving open for @Rangi42's opinion, feel free to close.

ISSOtm:
> since trying to apply our own, more optimal style is a slippery slope that'll cause more bloat in the file than good

Considering the patch is modifying the original code, and that it's thus a modification, it doesn't make sense to strictly stick to a coding style that we can only speculate about anyways, does it?

mid-kid:
> we can only speculate

I don't know about you, but I think it's pretty obvious how this function would've been fixed if they had noticed the bug, especially considering there's a [lot](https://github.com/pret/pokecrystal/blob/master/engine/battle/effect_commands.asm#L1632) of [similar](https://github.com/pret/pokecrystal/blob/master/engine/battle/core.asm#L5204) [code](https://github.com/pret/pokecrystal/blob/master/engine/battle/effect_commands.asm#L1326), and the way this function is written kinda gives away it's another one of the "this is an array but in code" they do.

Generally we try to stick to the way the surrounding code is written, exactly to avoid arguing about it...

Rangi42:
There are times when a micro-optimization is fine in bugfixes—like if you're adding a new routine, may as well tail call with `jp` instead of `call`+`ret`—but here there's a clear pattern of repeated `cp EFFECT_*` `jr z, .multihit` ended by `jp EndMoveEffect`, and the simplest fix is to add two lines continuing that pattern.

I also don't feel like the compatibility preservation fix is necessary. It's a standard application of the check described at the top, "only apply if the value at `[wLinkMode]` is not `LINK_COLOSSEUM`". The only exceptions are the 1/256 effect change bug, which has an elegant fix involving `scf` that may not be obvious without explanation; and a partial fix for one of Beat Up's bugs that only works cosmetically.

ISSOtm:
I was reading code in battle/core.asm yesterday, and found some code that defied expectations on how it was written (unlike usually, it had a `ret c` in the middle of it, and amusingly that was the unoptimized thing to do versus a `jr` to the end of the function). So I would argue Game Freak's style is indeed speculative.

PR #670, opened 2020-01-05
Identify "branch" labels in music files

Rangi42:


Rangi42:
@mid-kid no need to review every single file; they all follow the same pattern of `.mainloop`, `.loopN`, `.subN`, and sometimes `.subNloopN` labels. If that's OK then it can be merged.

mid-kid:
Sounds good to me. Anything's better than `_branch_xxxxxx` everywhere.

mid-kid:
...and of course now I realize this merge will mess with @dannye's work. I hope it's not too bad :x

coltongit:
Seems to be minor. Wouldn't take him too much longer to fix that.

dannye:
> Anything's better than _branch_xxxxxx everywhere.

I actually strongly object to this PR.
The audio/music/ files should never be modified by hand. They are auto-dumped by a script, so that whenever there is a format change you only need to edit the script and re-run it. When that happens, all hand-made changes are lost.

coltongit:
> 
> 
> > Anything's better than _branch_xxxxxx everywhere.
> 
> I actually strongly object to this PR.
> The audio/music/ files should never be modified by hand. They are auto-dumped by a script, so that whenever there is a format change you only need to edit the script and re-run it. When that happens, all hand-made changes are lost.
> 
> Please consider reverting this PR.

Then couldn't the script be edited, re-ran, and match up with your PR? This could be worked around.

Rangi42:
The music was originally dumped using a script, but so were map events. As far as I know, we haven't been maintaining either of the original dump scripts to work with the updated music and event commands. Also, this was a sufficiently mechanical update that a script could at least approximate it: use GlobalLabels for just the 1-4 channels, and .local ones for anything else. Calling them .local1, .local2, etc would be sufficient, or with a bit more complexity, notice which ones are used for `callchannel`, `loopchannel N`, or `loopchannel 0`, and name them `.subN`, `.loopN`, and `.mainloop` respectively. (This manual PR went further, adding redundant `.loopN` labels when a sub or mainloop is also looped, but that wouldn't be necessary.)

mid-kid:
@dannye 
We can revert this if need be, sorry for getting it out so quickly.

In my opinion, however, there are many, many files in the repository that have been initially dumped with such tools and were later modified to fit our needs, by adding comments, renaming macros, and whatever else. You could argue this is the case for the better part of the repository, even disregarding the assembly. In the end, though, this source is meant to be read by humans, so human-made adaptations are inevitable, both upstream and downstream. I think adhering to the laws of such a tool is a bit restrictive, even though the changes here aren't really that astonishing.

dannye:
> As far as I know, we haven't been maintaining either of the original dump scripts

I have always maintained the audio dump scripts, and I created a new cleaner one just for this PR (haven't pushed it to pret yet).

> In the end, though, this source is meant to be read by humans, so human-made adaptations are inevitable

I get what you're saying but I don't really agree with respect to audio/music/. Our hand-made changes can go in audio/engine.asm and macros/ etc.
But the audio/music/ files are 10's of thousands of lines. When we choose to change a macro name or macro argument format, I want to re-run the script 100% of the time.

mid-kid:
> But the audio/music/ files are 10's of thousands of lines. When we choose to change a macro name or macro argument format, I want to re-run the script 100% of the time.

We deal with this problem with a lot of the renames, considering the size of this repository. Usually, I write a throwaway script or `sed` command to go through whatever changes it can automatically. However, I don't think renames of this scale happen often enough. It's nice to keep it around to still be able to dump songs from the binaries, though.

PR #673, opened 2020-01-07
Add 0x1500 control code ACE fix

coltongit:
A fix for the 0x1500 control code ACE bug, which is further documented in https://glitchcity.info/wiki/0x1500_control_code_arbitrary_code_execution


Rangi42:
"Calling functions with values outside their expected domain causes misbehavior" is not a bug unless they can be so called in normal gameplay, and even then the fix would probably be to not create such invalid input in the first place. Step one of this glitch is "corrupt a Lapras to have character $15 in its name"; there's no reason for any code to be robust against this, much less a random unused-in-English-Crystal mobile routine.

Issue #666, opened 2019-12-08
Bug: Perish Song + Spikes = 0 HP Pokémon still battling


Rangi42:
Pokémon A faints to Perish Song; you then switch to Pokémon B, who faints to Spikes. Except it doesn't faint, it just reaches 0 HP. It can take its next move before actually fainting, which even allows it to Recover and avoid fainting at all.

mid-kid:
Cursed issue

coltongit:
Does this occur in the original ROM?

Rangi42:
Yes, I'm reporting it here so we can design a minimal fix and document it.

mid-kid:
It'd help a lot to have a save file and/or save state (preferably for BGB) to test, since this is a thing that requires a bit of setup.

coltongit:
This would preferably require video for the bugs/glitches document. I'd be happy to implement it in my PR.

coltongit:
A fitting name for this glitch: Perish Song+Spikes zombie Pokemon glitch

ISSOtm:
I'm willing to debug this this afternoon. Is there a save state reproducing the bug available somewhere?

coltongit:
Someone on Discord has a save state that was posted earlier today.

entrpntr:
Just to clarify the bug, it is the interaction between the faint checks in HandleBetweenTurnEffects and SpikesDamage (which only subtracts HP on a Pokémon's entrance, and doesn't handle fainting conditions).

For instance, say you have a full party at 1 HP and spikes up, and Pokémon A faints to Sandstorm damage in "HandleWeather". Pokémon B will faint after taking spikes damage from the check after "HandleWrap", Pokémon C will faint from the check after "HandlePerishSong", but Pokémon D will survive on 0 HP.

coltongit:
A fix for this issue is in PR #664.

PR #678, opened 2020-01-27
Optimize the Perish Song+Spikes glitch fix

coltongit:
This particular bugfix can be optimized by one line, which has been done here in this PR. References the glitch in issue #666.


FredrIQ:
This breaks the fix, causing an infinite loop.

mid-kid:
Please test your fixes.

ISSOtm:
The `call` *was* optimizable to `jp`, but the `ret` shouldn't have been removed.

coltongit:
> 
> 
> The `call` _was_ optimizable to `jp`, but the `ret` shouldn't have been removed.

Done so, but I guess that this will need a new PR now.

Issue #676, opened 2020-01-23
Miscellaneous fixes, part... 4?


aaaaaa123456789:
At [line 131 in `home/time.asm`](https://github.com/pret/pokecrystal/blob/master/home/time.asm#L131), tabs are used to horizontally align text with the line immediately below. As much as Python has attempted to convince people that tabs are four spaces, they are of course not, and thus this line is clearly not aligned to the one below it, as GitHub itself (which uses proper tabs) will show.

coltongit:
So you'd rather:
`; day     hr    min    sec`
instead of:
`;				  day     hr    min    sec`
The former would still make sense but in less tabs

Rangi42:
You can't assume tabs are a certain width; even in the browser that's configurable via `tab-size` in user CSS. So tabs should only be for indenting opcodes, not aligning lines vertically.

IMO the whole line is redundant given the self-explanatory WRAM names. And in my default Chrome configuration, they don't even line up:

![image](https://user-images.githubusercontent.com/35663410/72991444-05e30280-3dc0-11ea-964c-88e21eb16bfa.png)

Rangi42:
The `; S` `; M` `; H` `; DL` comments on further lines in this routine are also unnecessary.

Rangi42:
The four "padding" bytes in base data used to be two pointers to the front and back pics. This was the case in pokered and pokegold-spaceworld, and it explains why `GetBaseData` in home/mon_data.asm loads `UnknownEggPic` into `wBasePadding` for `EGG`.

PR #680, opened 2020-02-01
Miscellaneous fixes for #676

Rangi42:


data/pokemon/base_stats/abra.asm
```diff
@@ -12,7 +12,7 @@
 	db 20 ; step cycles to hatch
 	db 5 ; unknown 2
 	INCBIN "gfx/pokemon/abra/front.dimensions"
-	db 0, 0, 0, 0 ; padding
+	dw 0, 0 ; former front/back pics
```
mid-kid: This reads weird to me. Maybe say `; unused (front/backpics in generation 1)` or some form that mentions _when_ they were used.

home/mon_data.asm
```diff
@@ -31,24 +31,24 @@ GetBaseData::
 	jr .end
 
 .egg
-; ????
 	ld de, UnknownEggPic
```
mid-kid: Wasn't this also a gen1 thing?

data/pokemon/base_stats/abra.asm
coltongit: That seems to make far more sense description-wise.

home/mon_data.asm
Rangi42: Gen 1 wouldn't have had to special-case Egg.

mid-kid: Right shit yeah :sweat_smile:

data/pokemon/base_stats/abra.asm
Rangi42: Sure. I'm using "beta" rather than "Gen 1" though, IMO that's more applicable to pokecrystal since spaceworld still used those pointers.

PR #684, opened 2020-02-14
Define TileCollisionTable as a standard 'db' series

Rangi42:
There's no reason for this particular series to hide the `db` in an `EQUS` constant.


Rangi42:
Trivia: the `TALK` value is never checked for. `GetTileCollision` is the only routine that reads `TileCollisionTable`, and it masks out the high nybble before returning.

PR #685, opened 2020-02-14
Moshi moshi? minoru enhansumentsu desu

mid-kid:


home/names.asm
```diff
@@ -125,7 +125,7 @@ GetPokemonName::
 	ld e, a
 	ld h, 0
 	ld l, a
-	add hl, hl ; hl = hl * 4
+	add hl, hl ; hl = hl * 2
 	add hl, hl ; hl = hl * 4
 	add hl, de ; hl = (hl*4) + de
```
Rangi42: `; hl = (hl*4) + de = hl*5

home/names.asm
```diff
@@ -125,7 +125,7 @@ GetPokemonName::
 	ld e, a
 	ld h, 0
 	ld l, a
-	add hl, hl ; hl = hl * 4
+	add hl, hl ; hl = hl * 2
 	add hl, hl ; hl = hl * 4
```
Rangi42: `; hl = (hl*4) + de = hl*5`

home/names.asm
```diff
@@ -125,7 +125,7 @@ GetPokemonName::
...
 	add hl, de ; hl = (hl*4) + de
 	add hl, hl ; hl = (5*hl) + (5*hl)
```
Rangi42: `; hl = (hl*5) + (hl*5) = hl*10

home/names.asm
Rangi42: `; hl = (hl*4) + de = hl*5`

home/names.asm
Rangi42: I'm not sure these comments need to be here at all; "`; Each name is ten characters`" plus knowing what `add pp, qq` does tells you it's offsetting `hl` by `[wNamedObjectIndexBuffer] * 10`.

mid-kid: Or just one comment at the top with `hl = hl * 10`. I think it's at least slightly useful to annotate this one as it'll differ between translations.

Issue #686, opened 2020-02-21
Improvements thanks to upcoming rgbds versions


mid-kid:
This issue thread will keep track of different optimizations and improvements we can make when newer versions of rgbds are released.

- [ ] (RGBDS >0.3.9) Remove RST macros from macros/rst.asm in favor of using their labels

Rangi42:
- [ ] `ds $byte, $count` can replace `rept $count / db $byte / endr` in gfx/tileset_palette_maps.asm's `; unused` data

ISSOtm:
>     * [ ]  `ds $0150 - $0104` can become `ds $0150 - @` in home/header.asm

Well, considering hacks could choose to put anything there but 0 should be the default value for the header (as RGBFIX just patches over instead of overwriting the whole region), that'd be one of the places where `ds $150 - @, 0` would make sense.

ISSOtm:
2-arg `ds` added in rednex/rgbds@4cc24f436947890f41a3e16392a3c7ec7e8949ba.

ISSOtm:
https://github.com/rednex/rgbds/releases/tag/v0.4.0

Rangi42:
User-defined functions are under development, which will allow replacing the coord macros:

```
def coord(x, y) = coord(x, y, wTilemap)
def coord(x, y, origin) = y * SCREEN_WIDTH + x + origin

def bgcoord(x, y) = bgcoord(x, y, vBGMap0)
def bgcoord(x, y, origin) = y * BG_MAP_WIDTH + x + origin
```

Old:

```
	hlcoord 1, 2
	debgcoord 3, 4, wAttrmap
	dwcoord 5, 6
	ldcoord_a 7, 8
	lda_coord 9, 10
```

New:

```
	ld hl, coord(1, 2)
	ld de, bgcoord(3, 4, wAttrmap)
	dw coord(5, 6)
	ld [coord(7, 8)], a
	ld a, [coord(9, 10)]
```

If a ternary operator is added later, then `maskbits n` can become `ld a, bitmask(n)`:

```
def bitmask(n) = n > 0 ? bitmask(n, 1) : 0
def bitmask(n, x) = n <= x+1 ? x : bitmask(n, x*2+1)
```

Rangi42:
[...]

PR #687, opened 2020-02-21
Small home/ reorganization

mid-kid:
Time to move everything out of home.asm:
- InexplicablyEmptyFunction was moved to home/map.asm
- The wDebugFlags functions and xor_a brothers were moved to
  home/flag.asm because they're all flag-related.
- ret_2f3e was moved into home/region.asm
- The register alias sisters were moved to a new file called home/call_regs.asm
- IsInArray and SkipNames were joined by AddNTimes from home/math.asm
  into home/array.asm, as they're all used to index arrays.
- CallPointerAt was moved into home/print_text.asm because given the
  contents of that file it doesn't feel very out of place (that file
  isn't very aptly named...)
- CountSetBits was moved into home/pokedex_flags.asm because it's unique
  use is counting the amount of seen/caught mon in the podedex. GetWeekday
  was pulled into this by proximity.

Other changes were also made:
- PushLYOverrides was moved from home/sprite_anims.asm to
  home/battle.asm, because it's almost exclusively used for battle
  animations, with the lone exception being the Magnet Train.
- home/copy.asm was renamed to home/gfx.asm, as it's all gfx-related
- home/copy2.asm was renamed to home/copy.asm, now it's the only file
  called copy.
- SetHPPal and GetHPPal were moved from home/hp_pals.asm to
  home/tilemap.asm, as they're attrmap related, like many functions in
  that file are.
- home/rtc.asm was renamed to home/time_palettes.asm, as it had very
  little to do with the RTC at all, all RTC functions being in home/time.asm
- home/handshake.asm was renamed to home/printer.asm.
- home/mon_data_2.asm was renamed to home/mon_party.asm.


mid-kid:
Here's also a few suggestions on what else could be done, that I'm not too sure about:

Merge home/mon_stats.asm home/cry.asm home/print_level.asm home/mon_data.asm into home/pokemon.asm
These could be considered very related, and home/mon_stats.asm makes little sense, dealing with graphics as much as stats.
Also home/mon_data.asm deals with the base data as much as nicknames.

Merge home/mon_party.asm into home/battle.asm.
Even though it's used in *some* things other than battles, battle also deals a lot with getting data from the party.

I have no clue what to do with home/menu.asm, home/menu_window.asm and home/menu2.asm.
- home/menu.asm: Mostly joypad functions
- home/menu_window.asm: Aptly named, mostly "window stack"-related functions.
- home/menu2.asm: Actually loading menus and their data, a lot of these create a new window stack entry, and joypad functions.
I honestly would consider merging them, as I can't see any way to split them properly.

home/print_text.asm has things like CopyDataUntil and the name makes little sense but idk what to call it

Rangi42:
Thank you for clearly explaining your changes. This all looks good.

I also agree with your suggested merges:

- Merge mon_stats.asm, cry.asm, print_level.asm, and mon_data.asm into pokemon.asm
- Merge mon_party.asm into battle.asm
- Merge menu.asm, menu_window.asm, and menu2.asm into menu.asm

Four out of five functions in print_text.asm are related to text printing, so IMO it's fine. Other home files have one-off unrelated functions too, e.g. `FarCopyRadioText` in battle.asm.

This can be merged after those files are combined.

Rangi42:
So `NULL` and `Start` are still defined in home.asm. I think `NULL` would make sense merged into home/rst.asm; rst0 will always be at $0000, and following #686, the label can be used directly anyway. Also home/header.asm could be its own file, with a comment below `Start` explaining that the cartridge header data gets written there by rgblink (some other projects use a `ROM_HEADER` macro from gbhw.inc in the source code itself).

```
INCLUDE "constants.asm"


SECTION "NULL", ROM0

INCLUDE "home/rst.asm"
INCLUDE "home/interrupts.asm"


SECTION "Header", ROM0

INCLUDE "home/header.asm"


SECTION "Home", ROM0

INCLUDE "home/init.asm"
...
```

Rangi42:
Also `JoypadInt` could be renamed to `Joypad` like all the other interrupt labels. `Joypad` (the one called from `VBlank`) could then be renamed to `ReadJoypad` as a parallel to `GetJoypad`.

mid-kid:
Currently it's all in home.asm because that's (imo) where it makes most sense. In #631 I moved it all to home/init.asm. Maybe a home/header.asm would make more sense but home/init.asm also works imo. If we do that I'd also move the RSTs and interrupts into it (they're essentially part of the "header").
I wouldn't put `NULL` in home/rst.asm, as that's a really odd place imo (it's not an RST, despite the location).

mid-kid:
As for `JoypadInt`, I don't feel strongly towards anything but it'll always be confusing for there to be two joypad functions. `ReadJoypad` and `GetJoypad` are imo nonsensical twins. Maybe `UpdateJoypad` for the vblank one but idk.

Rangi42:
IMO the rsts, interrupts, and `Start` are special because they all come before the actual cart header data and the ROM code proper. Plus each rst, interrupt, and `Start` needs its own `SECTION` since their addresses are all fixed. I'd be okay with a single home/header.asm that has all 13 `SECTION`s in it, if you prefer that to three files where home/header.asm would just be for `Start`; IMO either one is better than `Start` being the only code left in home.asm.

I'm not sure if `NULL` is better as a label in the first place. True, that makes it viewable in the debugger, but is that useful? Looking at pokecrystal in BGB I see cases like "`ld hl,NULL / add hl,bc`" or "`ld de,NULL / call GetAnyMapField`"; in both cases it's misleading because the $0000 is meant as the number 0, not a pointer to start-of-ROM. (Compare: "`ld de,$0003 / call GetAnyMapField`".)

I don't feel strongly about `Joypad` either, so may as well leave it alone. It might be worth revisiting, since `Joypad` and `GetJoypad` are still a pretty confusing pair.

mid-kid:
I guess there isn't a specific reason for NULL to be a label but honestly don't see much reason to make it a constant either.
Also, if anything, once #686 is implemented we'll be seeing a lot of `ld hl, Reset`. Idk what the priority rules of the debugger/assembler are.

As for the header, I'll create the shared home/header.asm with everything that is accessed by the hardware or bootrom.

`Joypad`, `UpdateJoypad` and `GetJoypad` it is then.

Rangi42:
I'd make `NULL` a constant because now that most code is moved out of home.asm, it needs its own whole `SECTION` just to exist (you're right that rst.asm/header.asm would be nonsensical for it). And from C I'd expect something like `#define NULL ((void *)0)`, so `NULL EQU 0` in some constants file is slightly more obvious than a label `org`ed to the zero address. misc_constants.asm could work, since `FALSE` and `TRUE` are there too.

mid-kid:
That makes sense. I guess it doesn't hurt either. I'll just do that then.

mid-kid:
Apparently this is the reason it's a label:
```
ERROR: home.asm(51) -> home/names.asm(5) -> macros/data.asm::dba(71) -> macros/data.asm::dba::REPT~1(72) -> macros/data.asm::dbw(13):
    BANK argument must be a relocatable identifier
```

`NULL` is often used in `dba` or other macros where BANK() is in use.

home/header.asm
```diff
@@ -0,0 +1,66 @@
...
+
+SECTION "Header", ROM0
+
+Start::
+	nop
+	jp _Start
+
+; The cartridge header data is filled in by rgbfix.
+; This makes sure it doesn't get used.
+	ds $0150 - $0104
```
Rangi42: One more nitpick before this can merge: please make this `ds $0150 - CartridgeHeader` with a new `CartridgeHeader` label right above this, or maybe `ds $0150 - .StartEnd` if you prefer.

PR #688, opened 2020-02-23
Edit some comments about bad code

Rangi42:


mid-kid:
Honestly, it should be considered if these comments should even be there. There's a lot of similar inoptimal stuff like `call; ret` all over the place. We have a document pointing out the different optimization tricks you can apply and honestly there doesn't really have to be an optimization guide in the code itself.

ISSOtm:
Still useful to document bad code for people who are interested in the code itself but not in hacking it.

mid-kid:
Is it bad code though? We don't really know what the intention of these pieces of code is. Maybe some things were commented? Maybe the intention was different? I don't think the disassembly should comment more on this sort of stuff other than "we don't know what it does".

ISSOtm:
Well, the finished product is still what it is, so shouldn't that be commented? If we start to speculate on what the source code was, we will probably go nowhere.

Rangi42:
@mid-kid I agree in at least some cases. We shouldn't even try to comment on every suboptimal bit; it's subjective and there are too many. On the other hand, some of the "useless"/"pointless" comments are for weird code that doesn't follow a typical suboptimal pattern: `Unreferenced_Function92bbe`, `ItemSwitch_ConvertItemFormatToDW`, `and %11111100`, etc. Since those already have comments, I'd rather keep them and just not try to find more bits to comment. And get rid of the comments on standard things like `jp` instead of fallthrough.

Rangi42:
FWIW, the "stupid interns" comments confused me at first because I didn't understand what the code was doing + that it could be simpler.

Should any particular comments be removed outright, or can this be merged?

mid-kid:
@ISSOtm 
I just mean that commenting a "fix" for these is about as backwards as commenting other optimization fixes.

@Rangi42 
`Unreferenced_Function92bbe` is a similar situation to `GetForestTreeFrame`. Yes, given the current state of the code it _can_ be optimized bit but so can many, many other things both small and function-wide. Maybe they wanted the array to possibly be non-sequential? We don't even know what it was supposed to be, and it's unused to boot.
Idem for `ItemSwitch_ConvertItemFormatToDW` (also why aren't the array entries constants?).

Except for the `add` joypad values (which are obviously and verifiably written with no other intention) most of it falls into the "we don't really know why this is like that" category, and imo, so are many many things. Code improvements (of working code) shouldn't be in the disassembly imo, for the same reason we don't document them in docs/bugs_and_glitches.md. I'd just document what they do, which is what is already being done in places like `GetForestTreeFrame`.

I guess if you really want to comment improvements these changes look fine.

Rangi42:
I'm not attached to these particular optimizations, so may as well be consistent and just remove this sort of "here's some bad code" comment.

mid-kid:
If it were up to me I would just remove the added comment in `GetForestTreeFrame`, change the stats screen comments to `unknown purpose; increments a by 1`, change the one in `ItemSwitch_ConvertItemFormatToDW` to `Returns menu item format in bc` or `Returns wMenuData_ScrollingMenuItemFormat in bc` or no comment at all, and remove the comment in `Unreferenced_Function92bbe`. I would also remove the one in init_gender.

The one in mobile_menu is good, the one in map is good too, so is the one in phone.

I don't understand the one in menu_40 (hl is being loaded into de right?)

engine/pokemon/stats_screen.asm
```diff
@@ -60,11 +60,10 @@ StatsScreenInit_gotaddress:
 StatsScreenMain:
 	xor a
 	ld [wJumptableIndex], a
-	; stupid interns
 	ld [wcf64], a
 	ld a, [wcf64]
 	and %11111100
-	or 1
+	or 1 ; ???
```
mid-kid: Place the question marks at the start of the block, before `ld [wcf64], a`. We know what this line is doing (setting some 2-bit value in the byte) but not why.

PR #689, opened 2020-02-25
Fix Mac OSX install instructions

mid-kid:
Using homebrew since non-standard dependencies are required to install rgbds anyway.

### DIFF
--- a/engine/rtc/rtc.asm
+++ b/engine/rtc/rtc.asm
@@ -136,15 +136,14 @@ Function140ae:
 .time_overflow
 	farcall ClearDailyTimers
 	farcall Function170923
-; mobile
-	ld a, 5 ; MBC30 bank used by JP Crystal; inaccessible by MBC3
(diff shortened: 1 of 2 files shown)


engine/rtc/rtc.asm
```diff
@@ -136,15 +136,14 @@ Function140ae:
...
+	ld a, [s5_aa8c]
 	inc a
-	ld [$aa8c], a ; address of MBC30 bank
-	ld a, [$b2fa] ; address of MBC30 bank
+	ld [s5_aa8c], a
+	ld a, [s5_b2fa]
 	inc a
-	ld [$b2fa], a ; address of MBC30 bank
+	ld [s5_b2fa], a
 	call CloseSRAM
```
Asday: Were these changes _meant_ to be in this PR?

mid-kid: Good catch, they weren't, but it's often hard to make proper commits for small, inconsequential changes like this so I tend to leave them laying around for the next commit.

PR #699, opened 2020-03-21
Resolve some miscellaneous issues

Rangi42:


engine/battle_anims/anim_commands.asm
```diff
@@ -237,14 +237,14 @@ BattleAnim_ClearOAM:
...
 	jr z, .delete
 
-	; Instead of deleting the sprites, make them all use palette 0 (monochrome)
+	; Instead of deleting the sprites, make them all use PAL_BATTLE_OB_ENEMY
 	ld hl, wVirtualOAMSprite00Attributes
 	ld c, NUM_SPRITE_OAM_STRUCTS
 .loop
 	ld a, [hl]
-	and $ff ^ (PALETTE_MASK | VRAM_BANK_1)
+	and $ff ^ (PALETTE_MASK | VRAM_BANK_1) ; PAL_BATTLE_OB_ENEMY (0)
```
mid-kid: The `(0)` is redundant, imo, unless the index is more relevant than the constant name?

aaaaaa123456789: The `(0)` makes it clear why the `and` works by itself, without a comparison.

mid-kid: You're ANDing away all of the palette bits, so it seems rather obvious that the result is 0. IIRC this isn't the only place where we do this, and we only comment the constant. Unless there's something I'm missing?

aaaaaa123456789: Without the `(0)`, it seems like a random comment for a constant not mentioned at all in the expression. (The usual form is `and a ;CONST_FOO`, which is a lot clearer, but that's not what's happening here.) Maybe it's an old name for some of the constants involved? Maybe it's an alias? It's hard to tell what's going on by reading just the name of the constant there.

PR #702, opened 2020-03-30
Assorted labels in mobile related code

ariscop:
Not entirely sure Mobile46_SetErrorCode the a good name, but its good enough to show which branches are error handlers


mobile/mobile_45.asm
```diff
@@ -5026,27 +5026,27 @@ endr
 	jp nz, .asm_115c64
 	ret
 
-Function115cfd:
-	cp $2b
+Function115cfd: ; Base64
+	cp $2b ; "+"
```
mid-kid: Use the ascii charmap

mobile/mobile_46.asm
```diff
@@ -956,7 +956,7 @@ Function11878d:
 .asm_118803
 	ld a, $d3
 
-Function118805:
+Mobile46_SetErrorCode:
```
mid-kid: Just `Mobile_`

lib/mobile/main.asm
```diff
@@ -8983,23 +8983,23 @@ Unknown_113b7e:
...
-	db $05, $e9, $e3, $a9, $f8, $a3, $ef, $fc, $d9, $02, $6f, $67, $8a, $4c, $2a, $8d
-	db $42, $39, $fa, $ff, $81, $f6, $71, $87, $22, $61, $9d, $6d, $0c, $38, $e5, $fd
-	db $44, $ea, $be, $a4, $a9, $cf, $de, $4b, $60, $4b, $bb, $f6, $70, $bc, $bf, $be
-	db $c6, $7e, $9b, $28, $fa, $27, $a1, $ea, $85, $30, $ef, $d4, $05, $1d, $88, $04
-	db $39, $d0, $d4, $d9, $e5, $99, $db, $e6, $f8, $7c, $a2, $1f, $65, $56, $ac, $c4
-	db $44, $22, $29, $f4, $97, $ff, $2a, $43, $a7, $23, $94, $ab, $39, $a0, $93, $fc
-	db $c3, $59, $5b, $65, $92, $cc, $0c, $8f, $7d, $f4, $ef, $ff, $d1, $5d, $84, $85
-	db $4f, $7e, $a8, $6f, $e0, $e6, $2c, $fe, $14, $43, $01, $a3, $a1, $11, $08, $4e
-	db $82, $7e, $53, $f7, $35, $f2, $3a, $bd, $bb, $d2, $d7, $2a, $91, $d3, $86, $eb
+Unknown_113b8e: ; MD5 k table
```
Rangi42: Don't bother with a comment, just change the label to `MD5_K_Table`.

mobile/mobile_46.asm
ariscop: I used Mobile46 since there'll be equivalent functions in other bank

mid-kid: Alright then, though it'd be better if you could name them something else. Are they identical or are they serving for different purposes or deal with different data/codes?
If not this name is ok.

mobile/mobile_46.asm
```diff
@@ -3712,27 +3712,27 @@ Function119b6b:
 	call CopyBytes
 	jp BattleTowerRoomMenu_IncrementJumptable
 
-Function119c3e:
-	cp $2b
+Function119c3e: ; Base64
+	cp $2b ; "+"
```
mid-kid: Still not using ascii here.

ariscop: My bad, thought i forgot one

mobile/mobile_46.asm
ariscop: If its a problem i can just undo that change, the only reason i named it is so at call sites its apparent that branch is handling an error

Having another look i actually cant find an equivalent function, its unique to bank 46 and other banks just have its code inlined, but given that it Is unique to that bank a generic name like "Mobile_" could be misleading

mobile/mobile_45.asm
```diff
@@ -5026,27 +5024,27 @@ endr
 	jp nz, .asm_115c64
 	ret
 
-Function115cfd:
-	cp $2b
+Function115cfd: ; Base64
```
Rangi42: Is this encoding or decoding Base64? Please change the `Function115cfd` label to clarify.

mobile/mobile_45.asm
```diff
@@ -559,18 +557,18 @@ Function114412:
 	inc bc
 	inc e
 	call z, Function1144c8
-	cp $d ; CR
+	cp $d ; "\r"
```
Rangi42: (I've updated issue #686 to remind us to use `"\r"` here when rgbds 0.4.0 supports it.)

mobile/mobile_46.asm
Rangi42: Since the value it's setting is `wMobileErrorCodeBuffer`, how about `SetMobileErrorCode`? It's fine if the function is unique to this bank, there's no implication that it has to be usable in every bank (and if it were, it could always be `farcall`ed, not that any hacks are editing the mobile system anyway.)

mobile/mobile_46.asm
```diff
@@ -3712,29 +3718,34 @@ Function119b6b:
 	call CopyBytes
 	jp BattleTowerRoomMenu_IncrementJumptable
 
-Function119c3e:
-	cp $2b
+pushc
+setcharmap ascii
+
+Function119c3e: ; Base64
```
Rangi42: Likewise for this function; Base64 encode or decode?

PR #704, opened 2020-04-02
(gfx.c) Fix several out-of-bound reads in; fix a memory leak

LIJI32:
These caused several crashes on `rgbds`'s CI tests, which compile `pokecrystal` for sanity.

### DIFF
--- a/tools/gfx.c
+++ b/tools/gfx.c
@@ -102,9 +102,13 @@ void trim_whitespace(struct Graphic *graphic) {
 void remove_whitespace(struct Graphic *graphic) {
 	int tile_size = Options.depth * 8;
 	if (Options.interleave) tile_size *= 2;
+    
+    // Make sure we have a whole number of tiles, round down if required
+    graphic->size &= tile_size - 1;
+    
 	int i = 0;
 	for (int j = 0; i < graphic->size && j < graphic->size; i += tile_size, j += tile_size) {
-		while (is_whitespace(&graphic->data[j], tile_size)) {
+		while (j < graphic->size && is_whitespace(&graphic->data[j], tile_size)) {
 			j += tile_size;
 		}
 		if (j >= graphic->size) {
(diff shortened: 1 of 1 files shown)


tools/gfx.c
```diff
@@ -102,9 +102,13 @@ void trim_whitespace(struct Graphic *graphic) {
 void remove_whitespace(struct Graphic *graphic) {
 	int tile_size = Options.depth * 8;
 	if (Options.interleave) tile_size *= 2;
+    
+    // Make sure we have a whole number of tiles, round down if required
+    graphic->size &= tile_size - 1;
```
mid-kid, line 107: Mind your indents

tools/gfx.c
```diff
@@ -155,7 +163,7 @@ void remove_duplicates(struct Graphic *graphic) {
 }
 
 bool flip_exists(uint8_t *tile, uint8_t *tiles, int tile_size, int num_tiles, bool xflip, bool yflip) {
-	uint8_t *flip = calloc(tile_size, 1);
+	uint8_t *flip = alloca(tile_size);
```
mid-kid, line 166: Just use the array syntax.

LIJI32: Wasn't sure if the codebase was targeting C99 or not, I'll do that.

tools/gfx.c
```diff
@@ -164,6 +164,7 @@ void remove_duplicates(struct Graphic *graphic) {
 
 bool flip_exists(uint8_t *tile, uint8_t *tiles, int tile_size, int num_tiles, bool xflip, bool yflip) {
 	uint8_t flip[tile_size];
```
mid-kid, line 166:
```suggestion
	uint8_t flip[tile_size] = {0};
```

LIJI32: VLAs can't be initialized this way.

Issue #700, opened 2020-03-24
War for the Planet of the Miscellaneous Fixes


entrpntr:
(presumably the correct sequel to #694)

There is some variability in whether DAYCARE or DAY_CARE is used in various constant definitions. (Personally, I think DAYCARE looks better, and it is the more used of the two.)
```
$ grep -nr "const.*DAYCARE\|DAYCARE.*EQU" . | wc -l
      28
$ grep -nr "const.*DAY_CARE\|DAY_CARE.*EQU" . | wc -l
      15
```

Rangi42:
There is a hyphen in "DAY-CARE", but either is OK.

Lhivorde:
The macro `menu_coords` is currently defined in /macros/data.asm, but it would make more sense for it to be defined in /macros/coords.asm.

Lhivorde:
In the function `InitScrollingMenuCursor` (in /engine/menus/scrolling_menu.asm) theres a subfunction:
```
.asm_24763
	ret
```
Is there any reason not to rename it to `.ret` or something similar?

mid-kid:
There's many like these. Feel free to point them out and suggest names and/or make a pull request.

Rangi42:
There's never a reason for `.asm_XXXX` labels, we just haven't fixed them all yet. `python3 utils/unnamed.py pokecrystal.sym` will list them all.

Rangi42:
In `PokegearPhoneContactSubmenu`, `ld bc, hBGMapAddress + 1` should be `bccoord -2, -2, 0`.

Rangi42:
The `add_tm` macros can be simplified:

```
add_tm: MACRO
if !DEF(TM01)
TM01 EQU const_value
    enum_start 1
endc
    const TM_\1
    enum \1_TMNUM
ENDM
```

Rangi42:
`hSecondsBackup` is written three times but never read: twice `[hSecondsBackup] = [rLY]` in `_ResetWRAM`, once `[hSecondsBackup] = [hSeconds]` at the end of `VBlank0`. Maybe rename it `hUnusedBackup`.

Rangi42:
Most coordinate-related macros take X and then Y, but `dsprite` is the exception:

```
dsprite: MACRO
; y tile, y pxl, x tile, x pxl, vtile offset, attributes
	db (\1 * 8) % $100 + \2, (\3 * 8) % $100 + \4, \5, \6
ENDM
```

[...]

PR #705, opened 2020-04-04
Upgrade to rgbds 0.4.0, and do miscellaneous fixes

Rangi42:
- fail for rgbds < 0.4.0
- `rst <Label>`
- `ds <count>, <byte>`
- `shift <N>`
- ASCII "\r"
- Sorted .sym file


Rangi42:
This does not yet resolve the `DAYCARE`/`DAY_CARE` inconsistency from #700 because I'm not sure which to go with. `DAYCARE` is more common, but in-game it's spelled "DAY-CARE".

This does not yet use `LOAD`/`ENDL` from rgbds 0.4.0 for `WriteOAMDMACodeToHRAM.PushOAM` and `hTransferVirtualOAM` because that fails to build the `jr` correctly.

Rangi42:
Re: #700, I think the `DAY_CARE`/`DAYCARE` constants are already fine. Most use "`DAY_CARE`" to match in-game "DAY-CARE": there's `ENGINE_DAY_CARE_*`, `EVENT_DAY_CARE_*`, the `DAY_CARE` map, `SPRITE_DAY_CARE_MON_1/2`, `ROUTE34_DAY_CARE_MON_1/2`, etc.

There are three kinds of constant that use "`DAYCARE`": `DAYCAREMAN/DAYCARELADY_*_F`, `DAYCARETEXT_*`, and the `object_event` constants `DAYCARE_GRAMPS/GRANNY`. All of those are following the pattern where spaces get taken out of names for brevity. For example, `NEWBARKTOWN_SILVER`, `BATTLETOWERTEXT_INTRO`, or `LUCKYNUMBERSHOW_GAME_OVER_F`.

Also this doesn't need to wait for `LOAD`/`ENDL` to work properly, since `WriteOAMDMACodeToHRAM.PushOAM` is fine without them (the hard-coded `hTransferVirtualOAM:: ds 10` size is alright for now). So this should be ready to merge.

mid-kid:
Except for the `*TEXT` constants, I think object event constants and event flags constants should match their respective map's names.

Rangi42:
`EVENT_DAY_CARE_*` do match `map_const DAY_CARE`. Do you mean that all the `object_const_def`s should be updated? Such as `AZALEAMART_COOLTRAINER_M` → `AZALEA_MART_COOLTRAINER_M`?

mid-kid:
In all honesty - yes. It's kind of arbitrary to make a difference with something that's related to a very exact map. But if it's a lot, it can be left for a different PR.

Rangi42:
If we get file-local defines (https://github.com/rednex/rgbds/issues/342), I'd like them to be even shorter: no need for the map name in every constant, with or without underscores, since they're only ever used locally in the map script.

mid-kid:
Agreed, it's something I wanted to consider as a result of #631, not only with defines, but also with the labels. The only two that need to be prefixed are the two that are exported, `_MapScripts` and `_MapEvents`.

FAQ.md
```diff
@@ -38,11 +39,11 @@ You need to install `gcc`. If you're using Cygwin, re-run its setup, and at "Sel
 
 ### "ERROR: `UNION` already defined"
 
-Download [**rgbds 0.3.9**][rgbds] or newer. Older versions will not work.
+Download [**rgbds 0.4.0**][rgbds] or newer. Older versions will not work.
 
 ### "Expression must be 8-bit"
```
mid-kid, line 44: Add one with `ERROR: Macro not defined`. It turns up with the ascii thing.

INSTALL.md
```diff
@@ -157,7 +157,7 @@ If `rgbds` is not available, you'll also need these:
 To install **rgbds**:
 
 ```bash
-git clone -b v0.3.9 --depth=1 https://github.com/rednex/rgbds
+git clone -b v0.4.0 --depth=1 https://github.com/rednex/rgbds
 sudo make -C rgbds CFLAGS=-O2 install
```
mid-kid, line 161: Remove `CFLAGS` from each of these. `-O3` is the default now.

constants.asm
```diff
@@ -1,3 +1,7 @@
+if __RGBDS_MINOR__ < 4
```
mid-kid, line 1:
```suggestion
if __RGBDS_MAJOR__ <= 0 && __RGBDS_MINOR__ < 4
```

Rangi42:
Shall I change this to just `-Weverything` and merge?

mid-kid:
Sure, we can change it at any point. Be sure to include it in RGBASMFLAGS though.

PR #707, opened 2020-04-06
Update the header structure

Rangi42:
$0100–$0150 is "the cartridge header"; but only $0100–$0103 is the entry point code, $0104–$0150 is reserved to be patched by rgbfix.

People can add more code in all of the rst and interrupt sections, particularly `"joypad"`, which has 157 free bytes. This prevents them from accidentally doing the same after `Start`.

Also if they change the padding byte, like to $ff, rgbfix will still fill the header with $00s.

### DIFF
--- a/home/header.asm
+++ b/home/header.asm
@@ -56,14 +56,15 @@ SECTION "joypad", ROM0[$0060]
 	jp Joypad
 
 
-; Game Boy cartridge header
-
 SECTION "Header", ROM0[$0100]
 
 Start::
+; Nintendo requires all Game Boy ROMs to begin with a nop ($00) and a jp ($C3)
+; to the starting address.
 	nop
 	jp _Start
 
-; The cartridge header data is filled in by rgbfix.
-; This makes sure it doesn't get used.
-	ds $0150 - @
+; The Game Boy cartridge header data is patched over by rgbfix.
+; This makes sure it doesn't get used for anything else.
+
+	ds $0150 - @, $00


mid-kid:
What is the functional difference between this and what we had?

Rangi42:
A user can see this:

```
SECTION "rst8", ROM0[$0008]
FarCall::
	jp FarCall_hl

SECTION "rst10", ROM0[$0010]
```

and realize that `"rst8"` has 5 unused bytes for whatever code/data they want. But this:

```
SECTION "Header", ROM0[$0100]

Start::
	nop
	jp _Start

	ds $0150 - @, $00
```

does not have any room to add more code. There are conceptually two sections, the 4-byte entry point at $0100 and 76-byte rgbfix header at $0104. (Likewise, we don't have just one `SECTION "Header", ROM0[$0000]`; there's one for each thing that needs to start at a certain address.)

mid-kid:
I honestly don't think they're two separate things, but you're right in saying that any code written there will be overwritten by `rgbfix`. I feel like you'd be able to clear this up in the comment, however. Or how does one of them newfangled `assert`s sound for this?

Rangi42:
Okay, no separate `SECTION`; I'll just add a comment that the standard entry point is `nop / jp $0150`. (@aaaaaa123456789 do you know more about why that's the case? Prism no longer has that pattern and it's doing fine.) Also keeping the `, $00` since $ff padding is common.

aaaaaa123456789:
Nintendo required all ROMs to begin with a `nop` and a `jp`.  
(In fact, the official manual requires the ROM to begin with `00 C3 LL HH` where `$HHLL` is the entry point — it doesn't even mention that those are actual CPU instructions!)

PR #708, opened 2020-04-07
Remove unused nonmatching files for electro_ball.png and sgb_border.bin

Rangi42:
electro_ball.png needed to preserve one whitespace tile which tools/gfx would normally remove. Preserving a list of tile indexes has been added as a general feature to tools/gfx.

sgb_border.bin needed to remove the 20x18 center area of $00 bytes. This is achievable with the standard tr utility.

The output electro_ball.2bpp is identical to the previous matching one:

image

image

Likewise, the output sgb_border.nonull.bin matches, and the new sgb_border.bin is a simple 32x28 rectangle:

image


.gitignore
```diff
@@ -45,6 +45,7 @@ pokecrystal.txt
 *.1bpp
 *.lz
 *.animated.tilemap
+*.nonull.bin
```
mid-kid, line 48: Delete this file in the `Makefile`'s clean rule.

mid-kid: Also what about `.sgb.tilemap` instead?

PR #710, opened 2020-04-13
Proposal: Check the rgbasm version by using a separate unit

mid-kid:
This solves the problem of using build flags that aren't available in earlier versions of rgbasm, while still producing a clear error.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -50,7 +50,7 @@ crystal11:  pokecrystal11.gbc
 crystal-au: pokecrystal-au.gbc
 
 clean:
-	rm -f $(roms) $(crystal_obj) $(crystal11_obj) $(crystal_au_obj) $(roms:.gbc=.map) $(roms:.gbc=.sym)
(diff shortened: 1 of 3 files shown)


mid-kid:
RGBASM evaluates all symbols before parsing the expression, so `if !DEF(X) || X == Y` won't work, but the rest seems like a good idea.

Rangi42:
Oh, so we'd need

```
if !DEF(__RGBDS_MAJOR__) || !DEF(__RGBDS_MINOR__) || !DEF(__RGBDS_PATCH__)
	fail "pokecrystal requires rgbds {MAJOR}.{MINOR}.{PATCH} or newer."
elif (__RGBDS_MAJOR__ < MAJOR) || \
	(__RGBDS_MAJOR__ == MAJOR && __RGBDS_MINOR__ < MINOR) || \
	(__RGBDS_MAJOR__ == MAJOR && __RGBDS_MINOR__ == MINOR && __RGBDS_PATCH__ < PATCH)
	fail "pokecrystal requires rgbds {MAJOR}.{MINOR}.{PATCH} or newer."
endc
```

PR #712, opened 2020-04-17
Miscellaneous fixes

Rangi42:
The sectionful refactor was renaming pokecrystal.link anyway, and now the only files with "pokecrystal" in the name are build products.


engine/movie/credits.asm
```diff
@@ -85,7 +85,7 @@ Credits::
 	xor a
 	ldh [hBGMapMode], a
 	ld [wCreditsPos], a
```
mid-kid, line 87:
```suggestion
	ld [wCreditsPos + 0], a
```
I don't remember if we settled to do this for arrays, but if we did please change it.

Rangi42: I don't think we settled on that, or at least, the other two spots which load to `wCreditsPos` have `[wCreditsPos]` and `[wCreditsPos + 1]`.

Issue #709, opened 2020-04-08
Miscellany~


Rangi42:
- [x] macros/scripts/trade_anims.asm refers to engine/trade/animation.asm; it should say engine/movies/trade_animation.asm
- [x] Haircut brothers' probability comments are wrong in data/events/happiness_probabilities.asm
- [x] Document the home/map.asm:`ReadObjectEvents` bug
- [x] Correct `8 - a` comments to `-8 - a` in engine/gfx/sprites.asm
- [x] Identify `hFFBF`/`hFFC0`/`hFFC1`/`hFFC2` (see Polished Crystal)
- [x] `$7F` in `InitGenderScreen` should be `" "`
- [x] Pass `-p 0x00` to rgblink as well as rgbfix
- [x] Various 26s in engine/events/print_unown.asm should be `NUM_UNOWN`
- [x] Options menu needs a `NUM_OPTIONS` constant; and `ld c, $10` should be `SCREEN_HEIGHT - 2` (likewise for the textbox size 18x16)
- [x] `wCreditsUnusedCD21` is not unused, it's high byte of `dw` `wCreditsPos` (see https://github.com/TheFakeMateo/rgbs/issues/5)
- [x] PlayersHouse2F's `.SetSpawn` callback does not set a spawn point; it should be called `.SetUpTileDecorations`.

entrpntr:
As noted on the discord, rgbds v0.3.9 users of pokecrystal get the following error as of 9076d63:  `Expected constant expression: '__RGBDS_MINOR__' is not constant at assembly time.`

ISSOtm proposed this addition to make the error message less vague: `IF !DEF(__RGBDS_MINOR__) FAIL "Cannot detect RGBDS version; if using 0.3.9 please upgrade to 0.4.0" ENDC`

mid-kid:
Just add `!DEF(__RGBDS_MINOR__)` to the expression.

entrpntr:
Very minor, but since `hSecondsBackup` was renamed to `hUnusedBackup`, would it make sense to do something similar for `$FFA1` as well? It's only written to [here](https://github.com/pret/pokecrystal/blob/ed44d1cdb7e2f389be56a0158f8ce857e02a76b3/engine/overworld/events.asm#L816) and never read from, as far as I can tell. This makes `hMenuReturn` look like it is 2 bytes in that code, when it's only 1.

No idea on what the rename would be; maybe `hUnusedMenuReturn2`? Kinda long, and not sure it's an improvement.

[...]

PR #714, opened 2020-04-23
Miscellaneous updates and corrections

Rangi42:
The link battle−related changes are in regard to @entrpntr's comments on #709. They're far from clarifying what all the code is doing, but do introduce some named constants instead of repeating magic numbers, and get rid of the unnecessary `sLinkBattleStats` label.

### DIFF
--- a/audio/sfx.asm
+++ b/audio/sfx.asm
@@ -3327,7 +3327,7 @@ Sfx_DexFanfare170199_Ch8:
 	note B_, 16
 	sound_ret
 
-	db $fd
+	db sound_loop_cmd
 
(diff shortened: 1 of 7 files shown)


audio/sfx.asm
```diff
@@ -3327,7 +3327,7 @@ Sfx_DexFanfare170199_Ch8:
 	note B_, 16
 	sound_ret
 
-	db $fd
+	db sound_loop_cmd
```
mid-kid, line 3330:
```suggestion
	db sound_loop_cmd ; unused
```

PR #717, opened 2020-05-05
Miscellaneous fixes

Rangi42:
Addresses the haircut brothers' probability comments in #709


Rangi42:
Regarding the happiness probabilities:

```
older = $00 - $4b = 30%
      = $4c - $cb = 50%
      = $cc - $ff = 20%

young = $00 - $99 = 60%
      = $9a - $e5 = 30%
      = $e6 - $ff = 10%

daisy = $00 - $fe = 99.6%
```

data/pokemon/unown_words.asm
```diff
@@ -1,7 +1,7 @@
 unownword: MACRO
 x = 1
 rept STRLEN(\1)
-	db STRSUB(\1, x, 1) - $40
+	db STRSUB(\1, x, 1) - "A" + $40
```
mid-kid, line 4: What's the reasoning behind this?

Rangi42: The Unown font is loaded at tile $40. This macro turns charmap "A" into the first font tile, "B" into the second, etc. If the Unown font were loaded at e.g. tile $60, this would become `- "A" + $60`.

mid-kid: Ah, that makes sense. Could we add a `UNOWN_CHARMAP_START` constant so the loading routines can be aware of it, too?
Also, I wonder if it'd be better to use the new "multiple charmap" feature, instead.

Rangi42: I think a charmap would be overkill, since it would require 26 entries and only 26 words would even use it; plus it wouldn't apply to the `vTiles2 tile $40` code without a separate push/pop charmap just for that.

PR #724, opened 2020-05-22
Match LZ compressed files

### LINKED ISSUE #285: reproduce the compressor
the main sticking point is the way lookbacks are detected. merely finding the most efficient ones fails to match what the compressor does. this suggests that there is some time-saving trick at play. are there any obvious hacks that other compressors have used?

it could also be focusing on other metrics, but at a glance, it does not appear to care about decompression speed

aaaaaa123456789:
This PR submits an updated version of lzcomp (taken from [aaaaaa123456789/lzcomp](https://github.com/aaaaaa123456789/lzcomp)) that implements the compression variants needed to match the compressed files in the repo, as well as the makefile rules that will enable that match. The existing *.lz.\<hash\> files and the corresponding hash-checking rule from the makefile are deleted, since they are no longer necessary.

Resolves #285.


Rangi42:
Wow! Congratulations @aaaaaa123456789 ! I'm going to review the compression code for a while before merging this; it looks pretty dense.

gfx/lz.mk
```diff
@@ -0,0 +1,67 @@
+# Delete these rules if you don't care about matching.
+
+%.lz: LZFLAGS = --method 80
```
mid-kid, line 3: If you're using the `--align` argument everywhere, use it here, too. Or just remove it.

aaaaaa123456789: I don't think all rules use `--align`, but we can add an `--align 0` argument to the ones that don't, if that's what you want. Is it?

mid-kid: Idk what `--align 0` would do, woudn't `--align 1` just work?

aaaaaa123456789: Much like RGBASM's own alignment directives, `--align` here specifies the number of low bits that are clear — i.e., it aligns to 1 << n.

Rangi42:
It would be great if the usage info listed the compressor names and their valid method ranges.

aaaaaa123456789:
Done and force-pushed.

Rangi42:
> I can't really be bothered to look at the code itself much. As long as it works it's fine, I figure. The rest looks good.

I've read over it and it basically makes sense as an extension of lzcomp. The refactoring really helps with that. I'd personally use #defined constants more, but eh, it's ax6's code and it works.

Anything left before squash+merge?

PR #726, opened 2020-05-29
Rename ScrollingMenu_UpdateDisplay.string_2485f

Lhivorde:
Noticed an unnamed string in /engine/menus/scrolling_menu.asm, and a little investigating found that there isn't a consistent name for this sort of thing throughout the project (nor do current labels follow `STYLE.md`).  Are `.CancelString:` and `Prefix_CancelString:` adequate names?

### DIFF
--- a/engine/link/link.asm
+++ b/engine/link/link.asm
@@ -1571,10 +1571,10 @@ Unreferenced_Function28b42:
 	ld bc, SCREEN_WIDTH - 2
 	call ByteFill
 	hlcoord 2, 16
-	ld de, .cancel_string
+	ld de, .CancelString
 	jp PlaceString
 
-.cancel_string:
+.CancelString:
 	db "CANCEL@"
 
 Function28b68:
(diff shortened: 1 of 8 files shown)


engine/link/link.asm
```diff
@@ -1571,10 +1571,10 @@ Unreferenced_Function28b42:
 	ld bc, SCREEN_WIDTH - 2
 	call ByteFill
 	hlcoord 2, 16
-	ld de, .Cancel
+	ld de, .cancel_string
 	jp PlaceString
 
-.Cancel:
+.cancel_string:
```
mid-kid, line 1577: As per https://github.com/pret/pokecrystal/blob/master/STYLE.md
This should be `CancelString`

Lhivorde: Yup, I definitely misread the style guide.  I've fixed that and the other similar issues in the other edits.

mid-kid:
Please squash your commits, you don't need one commit per file, as this is pretty much one change across a bunch of files.

Lhivorde:
I'm using Github's website interface, and I think it defaulted to making every change its own commit.  I'm not finding a way to squash them together with the GUI, and while Google results are giving me lots of command line tutorials, I'm leery of using them when I don't really understand what they are saying or if they address the issue I think they do.  My apologies for making this more difficult than it needed to be.

PR #729, opened 2020-06-10
Treat std scripts like specials and predefs

Rangi42:
`callstd pc` becomes `callstd PCScript`, etc. Legacy support for the old constants.


Rangi42:
@entrpntr This is another change set which should be reflected in pokegold.

PR #740, opened 2020-07-08
Remove enum; add const_skip and const_next

Rangi42:
Corresponds to https://github.com/pret/pokegold/pull/55


constants/deco_constants.asm
```diff
@@ -54,16 +54,18 @@ const_value = 1
 	const SET_UP_ORNAMENT
 	const PUT_AWAY_ORNAMENT
 
+__deco_value__ = 0
+
 deco: MACRO
 	const DECO_\1
-	enum DECOFLAG_\1
+DECOFLAG_\1 EQU __deco_value__
```
mid-kid, line 61: Can't this just become `DECOFLAG_\1 EQU const_value - 1`?

constants/event_flags.asm
```diff
@@ -201,13 +202,9 @@
...
 	const EVENT_FOREST_IS_RESTLESS
 	const EVENT_ANSWERED_DRAGON_MASTER_QUIZ_WRONG
-; Unused
-	const EVENT_0C2
-	const EVENT_0C3
-	const EVENT_0C4
-	const EVENT_0C5
-	const EVENT_0C6
-	const EVENT_0C7
+; Unused: $0C2-$0C7
```
mid-kid, line 205: Is this useful at all? Maybe just document the size of the gap instead of the exact values?

constants/event_flags.asm
```diff
@@ -653,7 +313,7 @@
 	const EVENT_TODD_ASKED_FOR_PHONE_NUMBER
 	const EVENT_27E ; unused
 	const EVENT_GINA_ASKED_FOR_PHONE_NUMBER
-	const EVENT_280 ; 280 ; unused
+	const EVENT_280 ; unused
```
mid-kid, line 316: since we're using const_skip everywhere already... why not here?

constants/map_constants.asm
```diff
@@ -1,14 +1,15 @@
 newgroup: MACRO
 const_value = const_value + 1
```
mid-kid, line 2: honestly if you're going to re-implement `const` here, just give it a more appropriate name.

constants/trainer_constants.asm
```diff
@@ -1,5 +1,8 @@
+__trainer_class__ = 0
+
 trainerclass: MACRO
-	enum \1
+\1 EQU __trainer_class__
+__trainer_class__ = __trainer_class__ + 1
 const_value = 1
```
mid-kid, line 6: use const_def?

constants/type_constants.asm
```diff
@@ -19,15 +19,7 @@ PHYSICAL EQU const_value
...
-	const TYPE_10
-	const TYPE_11
-	const TYPE_12
-	const TYPE_13
-	const TYPE_14
-	const TYPE_15
-	const TYPE_16
-	const TYPE_17
-	const TYPE_18
+	const_skip 9
```
mid-kid, line 22: Super minor nitpick by why not `const_next 19` here? It's not like the data for any of these types is defined anywhere, and kinda clear they just decided to start at 20 for the special types.

macros/scripts/audio.asm
```diff
@@ -41,17 +41,17 @@ noise_note: MACRO
...
+FIRST_MUSIC_CMD EQU const_value
 
-	enum octave_cmd ; $d0
+	const octave_cmd ; $d0
 octave: MACRO
 	db octave_cmd | 8 - (\1) ; octave
 ENDM
 
-__enumdir__ = 1
+	const_next $d8
```
mid-kid, line 52: more like `const_skip 8`, the spacing is needed, not optional.

macros/scripts/battle_commands.asm
```diff
@@ -181,6 +181,6 @@ ENDM
 	command startloop               ; ae
 	command curl                    ; af
 
-	enum_start $fe
-	command endturn                 ; fe
+	const_def $ff, -1
```
mid-kid, line 184: `const_def -1, -1`

macros/scripts/gfx_anims.asm
```diff
@@ -12,42 +12,42 @@ endc
 	db x
 ENDM
 
-	enum_start $fc
+	const_def $ff, -1
```
mid-kid, line 15: const_def -1, -1
We use -1 to delimit lists and whatnot.

macros/scripts/movement.asm
```diff
@@ -117,104 +117,104 @@ else
 endc
 ENDM
 
-__enum__ = __enum__ + 8
+const_value = const_value + 8
```
mid-kid, line 120: const_skip 8

wram.asm
```diff
@@ -445,58 +445,52 @@ wAttackMissed::
...
 ; bit
-; 7 in love
-; 6 rollout
-; 5 endure
-; 4 perish song
-; 3 identified
-; 2 protect
-; 1 curse
 ; 0 nightmare
+; 1 curse
```
mid-kid, line 449: Maybe this should be handled in a different PR, but why document the bits here when they're documented in `constants/` and associated with the proper label in a comment? Maybe this is one of those places where `EXPORT` is a good idea?

constants/deco_constants.asm
Rangi42: No, because amid the `deco` uses are `const` uses: `const BEDS`, `const CARPETS`, `const PLANTS`, etc. So the bed `DECOFLAG`s are 2 less than the bed `DECO`s; the carpet `DECOFLAG`s are 3 less than the carpet `DECO`s; etc.

constants/event_flags.asm
Rangi42: Either is fine with me. I guess the size is more intuitive; "you can add this many events within the gap".

constants/map_constants.asm
Rangi42: `newgroup` increments `const_value` which is used for the `GROUP` constants, and resets `__map_value__` to 1 for the subsequent `MAP` constants via `map_const`.

Are you implying it should be `group_const`? But (a) that suggests it's defining a constant, but we don't have group constants independent of the maps (e.g. it's not `group_const OLIVINE`, `group_const DUNGEONS`, etc); and (b) the name "`group_const`" wouldn't suggest anything about resetting the `map_const` counter.

constants/trainer_constants.asm
Rangi42: `const_value` is for the individual trainer party constants, `__trainer_class__` is for the class constants. They have to run up in parallel.

wram.asm
Rangi42: Yeah, I'd rather just delete those redundant comments in wram.asm. The `SUBSTATUS_*` constant definitions already comment that they apply to `wPlayerSubStatus1`, `wPlayerSubStatus2`, etc, and have descriptive names. Maybe this wram.asm comment could just be "`; takes a SUBSTATUS_*` value`".

(Also, definitely save that for a separate PR, because I think the constants should also be `SUBSTATUS1_*`, `SUBSTATUS2_*`, etc, to make it clear which go where.)

constants/pokemon_constants.asm
```diff
@@ -272,9 +272,8 @@ JOHTO_POKEMON EQU const_value
 	const HO_OH      ; fa
 	const CELEBI     ; fb
 NUM_POKEMON EQU const_value - 1
-	const MON_FC     ; fc
+	const_skip       ; fc
 	const EGG        ; fd
-	const MON_FE     ; fe
```
mid-kid, line 277: Doesn't this one have palette slots/entries in other tables? If so, keep it with the name.

Rangi42: Some tables stop at Celebi; some have data for 252, Egg, 254, and also 255 (which is not a valid `MON_*` ID). How about putting back `MON_FE` and adding `CANCEL`?

mid-kid: CANCEL is a universal menu-related constant, not a pokemon. While I think it's a worthwhile addition, it shouldn't be here. I guess the extra entries aren't necessary if some tables stop at different places, anyway.

Issue #742, opened 2020-07-11
Automatically insert correct map event counts with macros


Rangi42:
It's easy to forget to update the `db` counts for `warp_event`s, `object_event`s, etc, which will build correctly but can cause obscure bugs. An automatic solution—ideally one that still just requires one line at the top of each event block—would be nice.

This one-line example will work in the next rgbds release, though it does not in 0.4.0:

```
def_warp_events: MACRO
	db NUM_WARP_EVENTS
NUM_WARP_EVENTS = 0
ENDM

warp_event: MACRO
NUM_WARP_EVENTS = NUM_WARP_EVENTS + 1
	...
ENDM

	def_warp_events
	warp_event  6,  7, NEW_BARK_TOWN, 2
	warp_event  7,  7, NEW_BARK_TOWN, 2
	warp_event  9,  0, PLAYERS_HOUSE_2F, 1
```

This two-line solution works now:

```
def_warp_events: MACRO
    db (.End_Warp_Events - .Warp_Events) / WARP_EVENT_SIZE
.Warp_Events
ENDM

end_warp_events: MACRO
.End_Warp_Events
ENDM

	def_warp_events
	warp_event  6,  7, NEW_BARK_TOWN, 2
	warp_event  7,  7, NEW_BARK_TOWN, 2
	warp_event  9,  0, PLAYERS_HOUSE_2F, 1
	end_warp_events
```

Rangi42:
So far this isn't working. I modified each event macro to have its own counter, like so:

```
def_callbacks: MACRO
	db _NUM_CALLBACKS
_NUM_CALLBACKS = 0
ENDM

callback: MACRO
;\1: type: a MAPCALLBACK_* constant
;\2: script pointer
_NUM_CALLBACKS = _NUM_CALLBACKS + 1
	dbw \1, \2
ENDM
```

and replaced all the `db N ; callbacks` with `def_callbacks`, etc. (I also had to put `_NUM_OBJECT_EVENTS = 0` before the dummy `object_event` for `PlayerObjectTemplate`.)

This builds, but does not match.

Rangi42:
Only some of the event counts are incorrect, but I'm not sure if the correct ones are just a coincidence.

This could be due to how multiple map scripts are included into sections of one scripts.asm file, even though each individual script has its own file. Possibly solvable by https://github.com/rednex/rgbds/issues/342

Rangi42:
[...]

PR #745, opened 2020-07-21
Use LOAD/ENDL for OAM DMA code in HRAM (requires rgbds 0.4.1)

Rangi42:
This takes care of one point from #686. Some label and comment changes are copied from pokered.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ os:
 install:
   - |-
     ( cd
-        git clone -b v0.4.0 --depth=1 https://github.com/rednex/rgbds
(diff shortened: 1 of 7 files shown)


Rangi42:
Also @mid-kid once we support 0.4.1, I'll do a PR for #742 with the one-line solution, if that sounds like a good idea. (It's easily backwards-compatible since `db #` would still work instead of `def_warp_events` or whatever the macro name ends up being.) (Any name suggestions for the six event types?)

mid-kid:
> Any name suggestions for the six event types?

Whatever is in the comment after the current `db`, prefixed by `def_`.

PR #751, opened 2020-07-24
Generate the Stadium 2 checksum data with pfero's tool instead of raw INCBINs

### LINKED ISSUE #744: Automatically generate the "stadium" data
Gold, Silver, and Crystal all have a binary blob of data at the very end of the ROM, apparently related to Stadium 2. In G/S it's 520 bytes; in Crystal it's 544. The format is as follows:

1. A 24-byte prefix only in Crystal:
   1. ASCII "`base`"
   2. Two bytes `01 00`
   3. An 18-byte sequence that's identical among the three final ROMs (1.0, 1.1, and AU) and among the two debug ROMs (1.0 and 1.1) but differs between the final and debug ROMs (it's unclear what it means; maybe it's related to Mobile Stadium)
2. An 8-byte header:
   1. ASCII "`N64PS3`" (presumably for "Nintendo 64 Pokémon Stadium 3", since Stadium G/S was the third Stadium release in Japan)
   2. A two-byte version marker that's unique to each ROM (possibly a checksum of the entire ROM, or of all the subsequent Stadium data?)
3. 512 bytes of actual Stadium data. This looks like it corresponds to the banks of the ROM prior to having the Stadium data patched over it. There are 128 ROM banks, so 4 Stadium bytes per bank. They appear to be local checksums: the byte `FE` corresponds to empty ROM space that's entirely `00`s.

For completeness' sake, it would be nice to reverse-engineer the checksum algorithm and write a tool to run between `rgblink` and `rgbfix` to patch the Stadium data onto the ROM.

Note that these checksums do not appear to be verified by Stadium 2; the hack Crystal Clear removed this binary data, but is still compatible with Stadium 2.

Rangi42:
Resolves #744.

@mid-kid Any critiques of the C style are welcome. Once this is approved I'll add it to pokegold as well.

(I separated the global checksum calculation from the half-bank checksums, even though it's a bit less efficient, because it makes the individual calculations simpler and summing 2 million bytes is trivial now.)


main.asm
```diff
@@ -724,20 +724,12 @@ INCLUDE "engine/events/battle_tower/load_trainer.asm"
 INCLUDE "engine/events/odd_egg.asm"
 
 
-SECTION "Mobile Stadium 2", ROMX
+SECTION "Stadium 2 Checksums", ROMX
```
mid-kid, line 727: While you're at it, hardcode the address and bank here too

tools/stadium.c
```diff
@@ -0,0 +1,170 @@
...
+			break;
+		default:
+			usage();
+			exit(1);
+			break;
+		}
+	}
+}
+
+#define SET2(file, off, v) do { \
```
mid-kid, line 70: `SET_U16BE`? `WRITE_U16BE`?

tools/stadium.c
```diff
@@ -0,0 +1,164 @@
...
+
+	int NUMCHECKS = filesize / CHECKSIZE;
+	int DATASIZE = HEADERSIZE + NUMCHECKS * 2; // 2 bytes per checksum
+	int ORIGIN = filesize - DATASIZE; // Stadium data goes at the end of the file
+
+	// Clear global checksum
+	SET2(file, GLOBALOFF, 0x0000);
+
+	// Clear Stadium data (this should be free space anyway)
+	for (int i = 0; i < DATASIZE; i++) {
```
mid-kid: `memset(file + ORIGIN, 0, DATASIZE);`

tools/stadium.c
```diff
@@ -0,0 +1,164 @@
...
+		file[ORIGIN + i] = 0x00;
+	}
+
+	// Fill Stadium header values
+	for (int i = 0; i < HEADERSIZE; i++) {
```
mid-kid: `memcpy(file + ORIGIN, header, HEADERSIZE);`

tools/stadium.c
```diff
@@ -0,0 +1,158 @@
...
+
+	// Calculate the CRC of the half-bank checksums
+	uint16_t crc = 0xFEFE;
+	for (int i = ORIGIN + HEADERSIZE; i < ORIGIN + DATASIZE; i++) {
+		crc = (crc >> 8) ^ crc_table[(crc & 0xFF) ^ file[i]];
+	}
+	SET2(file, ORIGIN + HEADERSIZE - 2, crc);
+
+	// Calculate the global checksum
+	uint16_t globalsum = 0x0000;
```
mid-kid, line 129: I would honestly do this during the halfbank checksum calculation. By making it a separate variable you can still make the intent clear enough.

tools/stadium.c
```diff
@@ -0,0 +1,158 @@
...
+		for (int j = 0; j < CHECKSIZE; j++) {
+			checksum += file[i * CHECKSIZE + j];
+		}
+		SET2(file, ORIGIN + HEADERSIZE + i * 2, checksum);
+	}
+
+	// Initialize the CRC table
+	uint16_t crc_table[256] = {0};
+	for (int i = 0; i < 256; i++) {
+		uint16_t c = (uint16_t)i;
```
mid-kid, line 112: Unnecessary cast.

tools/stadium.c
```diff
@@ -0,0 +1,154 @@
...
+		uint16_t t = 0x0000;
```
mid-kid, line 113: Either call it `crc` or `rem` (remainder), as that's pretty much what's happening here, you're precalculating the crc for a set of values to match the size of the values you're reading later. I guess it could be incorporated into the same loop so the table doesn't have to be built, but here's hoping a smart compiler will realize it's static.

tools/stadium.c
```diff
@@ -0,0 +1,154 @@
...
+		for (int y = 0; y < 8; y++) {
+			t = (t >> 1) ^ ((t ^ c) & 1 ? 0xC387 : 0);
```
mid-kid, line 115: Make a constant for `CRC_POLY` and `CRC_INIT`.

tools/stadium.c
```diff
@@ -0,0 +1,154 @@
...
+	for (int i = 0; i < NUMCHECKS; i++) {
+		uint16_t checksum = 0xFEFE;
+		for (int j = 0; j < CHECKSIZE; j++) {
+			checksum += file[i * CHECKSIZE + j];
+		}
+		SET2(file, ORIGIN + HEADERSIZE + i * 2, checksum);
+	}
+
+	// Initialize the CRC table
+	uint16_t crc_table[256] = {0};
```
mid-kid, line 110: No need to initialize it here, it's being initialized in the loop anyway.

tools/stadium.c
Rangi42: The problem is having to account for the final checksum and CRC separately. Here it's clearly just a sum of every byte in the file, no relation to the Stadium checksums or CRC.

tools/stadium.c
Rangi42: I went with `rem` since `crc` below is the actual single CRC.

mid-kid:
One last thing, the `base` area shouldn't be cleared if the `--base` option isn't specified. You can set `ORIGIN` to the start of the N64PS3 header and have `HEADERSIZE` only account for the N64PS3+checksum.

Rangi42:
The tool is updated to handle the no-`base` case appropriately.

PR #754, opened 2020-07-24
Use def_* macros for automatic map event counts

### LINKED ISSUE #742: Automatically insert correct map event counts with macros

Rangi42:
Resolves #742

### DIFF
--- a/docs/map_event_scripts.md
+++ b/docs/map_event_scripts.md
@@ -24,7 +24,7 @@
 ## Object constants
 
 <pre>
-	object_const_def ; object_event constants
+	object_const_def
 	const <i>MAPNAME</i>_<i>OBJECTNAME</i>
 </pre>
 
@@ -39,7 +39,7 @@
 ### Scene scripts
 
 <pre>
-	db <i>N</i> ; scene scripts
+	def_scene_scripts
 	scene_script <i>script</i>
 </pre>
 
@@ -47,7 +47,7 @@
 ### Callbacks
 
 <pre>
-	db <i>N</i> ; callbacks
+	def_callbacks
 	callback <i>type</i>, <i>script</i>
 </pre>
 
@@ -105,7 +105,7 @@
 ### Warp events
 
 <pre>
-	db <i>N</i> ; warp events
+	def_warp_events
 	warp_event <i>x</i>, <i>y</i>, <i>map</i>, <i>warp_id</i>
 </pre>
 
@@ -113,7 +113,7 @@
 ### Coord events
 
 <pre>
-	db <i>N</i> ; coord events
+	def_coord_events
 	coord_event <i>x</i>, <i>y</i>, <i>scene_id</i>, <i>script</i>
 </pre>
 
@@ -121,7 +121,7 @@
 ### BG events
 
 <pre>
-	db <i>N</i> ; bg events
+	def_bg_events
 	bg_event <i>x</i>, <i>y</i>, <i>type</i>, <i>script</i>
 </pre>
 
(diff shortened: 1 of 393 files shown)


aaaaaa123456789:
I have mixed feelings about this PR. It will certainly eliminate a common pain point, but it also looks like a giant hack that will break in five minutes.

Rangi42:
@aaaaaa123456789 How so? It has the same failure modes as the former `db`s, where you might omit a `def_*`  line or define an event before its `def_*` line (although I doubt either of those errors is common), but makes impossible the more likely failure of getting the number wrong.

aaaaaa123456789:
It's the lazy evaluation of redefined constants what makes me uneasy.

Rangi42:
Well, they're not quite redefined; the "`\@`" creates a new constant for each map (`_NUM_SCENE_SCRIPTS_0`, `_NUM_SCENE_SCRIPTS_1`, `_NUM_CALLBACKS_0`, etc). The strings like `_NUM_SCENE_SCRIPTS` and `_NUM_CALLBACKS` do get purged and redefined for each map, but that's a standard rgbds feature, essential for things like [rgbds-structs](https://github.com/ISSOtm/rgbds-structs/blob/master/structs.asm), so its behavior isn't going to change unexpectedly.

aaaaaa123456789:
I guess that `PURGE` isn't about to break...

ISSOtm:
I just reviewed this, and this is absolutely defined, and perhaps the cleanest implementation of this. The `equs` is stored in-place, and in fact the `\@` is expanded when the string is read anyways, so even if evaluation was to become lazy, this would still work.

Note that the way the current lexer works does make this prone to breaking (macro args expansions have to be explicitly handled everywhere), but the one I'm currently writing should stabilize this. (And keep the behavior.)

Rangi42:
> macro args expansions have to be explicitly handled everywhere

Can you explain how this might break due to that? I tested omitting a `def_` macro and putting an event macro before its heading `def_` macro, and in both cases, it just assembled without errors, same as the current `db` solution. (Of course the data would be wrong, but that's expected. The macros could have error checking for those kind of mistakes, but it's probably not worth the extra complexity.)

ISSOtm:
I'm saying that if a change were to be made to the current lexer that forgot to handle macro args, they'd be essentially deferred, breaking this. My actual point was that the new lexer will be much more reliable as far as this functionality goes.

PR #760, opened 2020-08-18
Enable compatibility with upcoming RGBASM version

ISSOtm:
This is additionally required, because an `elif`'s condition is evaluated even when it's about to be skipped over, and this `"\2"` will become an error.

### DIFF
--- a/home/text.asm
+++ b/home/text.asm
@@ -192,11 +192,13 @@ if ISCONST(\2)
 	jr nz, ._\@
 	ld a, \2
 ._\@:
-elif STRSUB("\2", 1, 1) == "."
-; Locals can use a short jump
-	jr z, \2
 else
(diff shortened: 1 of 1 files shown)


Rangi42:
I'm surprised that `elif`, or a subsequent `else`, would *evaluate* the condition (not just parse it) even when not taken. Python, C, Java, and I'd expect most other languages don't do that with their `elif`/`elsif`/`else if`.

aaaaaa123456789:
This is bad, but I feel we're in the wrong issue tracker.

Rangi42:
> I could inject a lexer hack to return a different token and skip the evaluation, but...  
> Well, expression evaluation needs to be deferred anyways  
> Especially if we want user-defined functions

Merging this for now; short-circuit `if`/`elif` can be tracked in #686.

PR #761, opened 2020-08-19
Linkcable and the Mystery of Mew

mid-kid:
Document weird code involving link trades of mew and celebi

This piece of code is frankly rather confusing, as it seemingly doesn't
serve much of a purpose. Documenting it here for future reference.

### DIFF
--- a/engine/link/link.asm
+++ b/engine/link/link.asm
@@ -1882,24 +1882,31 @@ LinkTrade:
 	call LoadTradeScreenBorder
 	call SetTradeRoomBGPals
 	farcall Link_WaitBGMap
-	ld b, $1
+
+; Check if either of the Pokémon sent was a Mew or Celebi, and send a different
+; byte depending on that. Presumably this would've been some prevention against
+; illicit trade machines, but it doesn't seem like a very effective one.
+; Removing this code breaks link compatibility with the vanilla gen2 games, but
+; has otherwise no consequence.
+	ld b, 1
 	pop af
 	ld c, a
 	cp MEW
-	jr z, .loop
+	jr z, .send_checkbyte
 	ld a, [wCurPartySpecies]
 	cp MEW
-	jr z, .loop
-	ld b, $2
+	jr z, .send_checkbyte
+	ld b, 2
 	ld a, c
 	cp CELEBI
-	jr z, .loop
+	jr z, .send_checkbyte
 	ld a, [wCurPartySpecies]
 	cp CELEBI
-	jr z, .loop
-	ld b, $0
+	jr z, .send_checkbyte
 
-.loop
+; Send the byte in a loop until the desired byte has been received.
+	ld b, 0
+.send_checkbyte
 	ld a, b
 	ld [wPlayerLinkAction], a
 	push bc
@@ -1913,7 +1920,7 @@ LinkTrade:
 	jr z, .save
 	ld a, [wOtherPlayerLinkAction]
 	cp b
-	jr nz, .loop
+	jr nz, .send_checkbyte
 
 .save
 	farcall SaveAfterLinkTrade


engine/link/link.asm
```diff
@@ -1882,23 +1882,29 @@ LinkTrade:
 	call LoadTradeScreenBorder
 	call SetTradeRoomBGPals
 	farcall Link_WaitBGMap
-	ld b, $1
+
+; Check if either of the pokemon sent was a Mew or Celebi, and send a different
+; byte depending on that. Presumably this would've been some prevention against
+; illicit trade machines, but it doesn't seem like a very effective one.
```
Rangi42, line 1888: "Presumably this could've been some prevention against illicit trade machines, but it doesn't actually do anything." (It doesn't do anything, right? You can trade Mew or Celebi like any other Pokémon.)

engine/link/link.asm
```diff
@@ -1882,23 +1882,29 @@ LinkTrade:
 	call LoadTradeScreenBorder
 	call SetTradeRoomBGPals
 	farcall Link_WaitBGMap
-	ld b, $1
+
+; Check if either of the pokemon sent was a Mew or Celebi, and send a different
```
Rangi42, line 1886: **P**okémon

engine/link/link.asm
```diff
@@ -1882,23 +1882,29 @@ LinkTrade:
...
+; byte depending on that. Presumably this would've been some prevention against
+; illicit trade machines, but it doesn't seem like a very effective one.
+; Removing this code breaks link compatibility with the vanilla games, but has
+; otherwise no consequence.
+	ld b, 1
 	pop af
 	ld c, a
 	cp MEW
 	jr z, .loop
```
Rangi42, line 1895: The `.loop` makes this slightly confusing since no looping is going on. How about a `.set_b` label at the end of this block, coinciding with the `.loop` below?

mid-kid: Looping is going on, though. If the value being sent by the other device doesn't match the `b` decided here, it keeps looping.

engine/link/link.asm
mid-kid: You can trade these pokemon between devices that send the right byte here, which all gen2 games do, and it's ignored when linking with gen1 games.
My best guess is that the intention was to trip up people that made software or hardware to trade with gen2 games, that wouldn't have sent the right byte here if they didn't specifically test this scenario. To that effect it certainly does _something_. It's just not necessary to perform a trade, as this specific trade will still work between two games that remove this code.

Rangi42: Okay, then how about another comment on the `cp b` below explaining that the two games' sent bytes have to match?

engine/link/link.asm
Rangi42: What I meant is the `jr z, .loop` lines in this block are not going backwards to re-iterate a loop; they're going forwards to the beginning of a loop below. Or rather, to the end of the block that sets `b`, and right after it is the beginning of the block that loops and compares the received value with `b`.

Rangi42: May as well also comment above the `cp LINK_TIMECAPSULE` noting that Gen 1 games don't send such a byte for Mew.

mid-kid: > What I meant is the jr z, .loop lines in this block are not going backwards to re-iterate a loop; they're going forwards to the beginning of a loop below.

This is a rather common pattern you'll see across the codebase, except with names other than `.loop`. Would `.send_checkbyte` work?

> May as well also comment above the cp LINK_TIMECAPSULE noting that Gen 1 games don't send such a byte for Mew.

"Removing this code breaks link compatibility with **other gen2 games**"?

Rangi42: Sure, it's a long function and some of those `.asm_XXXX` labels are probably loops too, so `.send_checkbyte` is more explicit.

engine/link/link.asm
mid-kid: I've clarified this in a comment right before the loop. Other than that this should be obvious from the context (set byte to send, exchange, check received byte).

autofire372:
Just stopping by to say that is maybe the best PR name ever.

PR #762, opened 2020-08-22
Better document splash screen

mid-kid:
Moved assets to the right directory, split engine/movie/intro.asm, and documented the code surrounding the splash screen.


Rangi42:
This is still marked as a draft; is there anything left to do, or can it be merged (along with similar updates to pokegold and pokered)?

mid-kid:
IIRC I was still working on something, but I can't exactly recall what. I'll create a pokegold PR and undraft this.

engine/movie/splash.asm
```diff
@@ -149,17 +149,16 @@ GameFreakPresents_PlaceGameFreak:
...
 	call CopyBytes
 	call GameFreakPresents_NextScene
 	ld de, SFX_GAME_FREAK_PRESENTS
 	call PlaySFX
 	ret
 
-.GAME_FREAK:
-	;  G  A  M  E   _  F  R  E  A  K
+.game_freak
 	db 0, 1, 2, 3, 13, 4, 5, 3, 1, 6
```
Rangi42, line 162: These are tile IDs are are usually written in $hex.

engine/gfx/sprite_anims.asm
```diff
@@ -220,7 +220,7 @@ DoAnimFrame:
 	callfar ComposeMail_AnimateCursor
 	ret
 
-.GameFreakLogo:
+.GameFreakLogo
```
Rangi42, line 223: Can we consistently have colons for the labels that correspond to `SPRITE_ANIM_SEQ_*` constants? Right now some do, some don't.

mid-kid: I just looked at the surrounding stuff and nothing had it, so removed it. Will just add to everything.

mid-kid: This will be left for a different PR, where global labels will be created, since there's many locals in this file.

PR #768, opened 2020-08-29
Group bugs and glitches by category

Rangi42:
Categories were inspired by the ones from [GCL for Red/Blue](https://forums.glitchcity.info/index.php?topic=7682.0). (Reminder that pokered has no bug documentation yet.)

Please comment if any categorizations should be changed. In particular, the two cosmetic HP bar animation bugs are grouped in "Link battle engine" since they're visible to both linked players, but should arguably be in "Graphics, audio, and text".

### DIFF
--- a/docs/bugs_and_glitches.md
+++ b/docs/bugs_and_glitches.md
@@ -10,84 +10,93 @@ Fixes are written in the `diff` format. If you've used Git before, this should l
 +add green + lines
 ```
 
-Some fixes are mentioned as breaking compatibility with link battles. This can be avoided by writing more complicated fixes that only apply if the value at `[wLinkMode]` is not `LINK_COLOSSEUM`. That's how Crystal itself fixed two bugs in Gold and Silver regarding the moves [Reflect and Light Screen](#reflect-and-light-screen-can-make-special-defense-wrap-around-above-1024) and [Present](#present-damage-is-incorrect-in-link-battles).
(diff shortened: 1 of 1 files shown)


Rangi42:
(Okay, no more force pushes. :P)

mid-kid:
Instead of "link battle engine" I'd say "multi-player battle engine" to contrast with the single player engine.
Also instead of "overworld sprites and tiles" just say "overworld engine". That fits all of the issues listed there better.

I'm not sure adding the disclaimer for link compatibility breakage under the link battle engine header is a good idea, especially when no other headers have any text, and you have to scroll past the table of contents to see it, it's incredibly easy to miss. Instead, modify the third paragraph to mention it, something like:

```
Fixes under the "multi-player battle engine" category will break compatibility with standard Pokémon Gold/Silver/Crystal for link battles, unless otherwise noted. This can be avoided by writing more complicated fixes that only apply if the value at [wLinkMode] is not LINK_COLOSSEUM. That's how Crystal itself fixed two bugs in Gold and Silver regarding the moves Reflect and Light Screen and Present.
```

Rangi42:
Thanks, those changes have been made.

docs/bugs_and_glitches.md
```diff
@@ -10,84 +10,93 @@ Fixes are written in the `diff` format. If you've used Git before, this should l
 +add green + lines
 ```
 
-Some fixes are mentioned as breaking compatibility with link battles. This can be avoided by writing more complicated fixes that only apply if the value at `[wLinkMode]` is not `LINK_COLOSSEUM`. That's how Crystal itself fixed two bugs in Gold and Silver regarding the moves [Reflect and Light Screen](#reflect-and-light-screen-can-make-special-defense-wrap-around-above-1024) and [Present](#present-damage-is-incorrect-in-link-battles).
+Fixes in the "[multi-player battle engine](#multi-player-battle-engine)" category will break compatibility with standard Pokémon Gold/Silver/Crystal for link battles, unless otherwise noted. This can be avoided by writing more complicated fixes that only apply if the value at `[wLinkMode]` is not `LINK_COLOSSEUM`. That's how Crystal itself fixed two bugs in Gold and Silver regarding the moves [Reflect and Light Screen](#reflect-and-light-screen-can-make-special-defense-wrap-around-above-1024) and [Present](#present-damage-is-incorrect-in-link-battles).
```
mid-kid, line 13: No need for quotes if it's a hyperlink.

PR #771, opened 2020-09-27
Identify battle animation functions

7Soul:
I'm new to git stuff so let me know if this is a good start and I'll work on the rest


Rangi42:
Thanks for starting this! The label names could be more like the ones that were already identified. For example, we have "`ThrowFromPlayerToEnemy`", so "`DiagonalMoveStop`" could be "`MoveFromPlayerToEnemy`", "`TranslateSmallRadius`" could be "`MoveInCircle`" or "`MoveSmallCircle`" or something, etc. Your comments with examples are a good idea, and they seem to have more descriptive language than the labels so far. For instance, "`DiagonalSineMove`" could be "`MoveWaveToEnemy`", or "`MoveInWave`", or something like that. (If the whole "FromPlayerToEnemy" convention is too verbose, it could just be "ToEnemy" for all of them.)

7Soul:
Furthermore, "PlayerToEnemy" isn't quite right since the direction can be flipped, so something like "UserToOpponent" might be better

Rangi42:
We have the charmaps `"<USER>"` and `"<TARGET>"`, so "`UserToTarget`" would be great.

mid-kid:
For clarity, please mark this PR as a draft, and unmark it when you're happy with merging it.

7Soul:
As soon as I discover how to do it, I will!

7Soul:
For naming functions only used by one move, obviously I'll name it after the move like `BattleAnimFunction_WaterGun`, and when it's a general object it's named after it like `BattleAnimFunction_Horn`. But some functions are used specifically for 2 moves but the objects are different, such as the function used for Thief and Payday (one is a coin and one is a bag) so I named the function `BattleAnimFunction_ThiefPayday`. Other ones like it so far include `GrowthSwordsDance`, `StrengthSeismicToss`, `SmokeFlameWheel` and `PresentSmokescreen`

7Soul:
Just a heads up that it's finished. Once this is merged I can get to other animation stuff

engine/battle_anims/functions.asm
```diff
@@ -3995,7 +4112,9 @@ Functionce672:
 	ld [hl], a
 	ret
 
-BattleAnimFunction_4B:
+BattleAnimFunction_Unused1:
```
Rangi42, line 4115: We're [transitioning](https://github.com/pret/pokecrystal/wiki/Code-cleanup#unusedunreferencedstubs) from `Unused` in symbols to `; unused` as comments. How about `BattleAnimFunction_AnimObjB0`?

engine/battle_anims/functions.asm
```diff
@@ -3791,24 +3896,25 @@ Functionce577:
...
+	jr step_circle
 
-.asm_ce58b
+.done
 	call DeinitBattleAnimation
 	ret
 
-asm_ce58f:
-	call Functionce6f1
+step_circle:
```
Rangi42, line 3912: This sort of label should be local.

engine/battle_anims/functions.asm
```diff
@@ -3912,7 +4024,7 @@ Functionce618:
 	call DeinitBattleAnimation
 	ret
 
-asm_ce61c:
+move_up:
```
Rangi42, line 4027: This sort of label should be local (`.move_up`).

Rangi42 (changes requested):
Overall, this looks very good! I've made a few comments to change; and also, if you can rename the `BATTLEANIMFUNC_*` constants to match the `BattleAnimFunction_*` labels, that would be helpful (and then the table of `dw`s wouldn't need the commented constants).

engine/battle_anims/functions.asm
7Soul: With this one and .step_circle they're both called from a previous function

> error: main.asm(438) -> engine/battle_anims/functions.asm(3887): Unknown symbol "Functionce564.step_circle"
error: main.asm(438) -> engine/battle_anims/functions.asm(3887): jr target out of reach (expected -129 < -7027 < 128)

7Soul:
I went ahead and renamed the last battle anim struct constant that wasn't named `BATTLEANIMSTRUCT_02` and `wBattleAnimTempField02` which are used in one case where an animation has to be fixed because of how the engine moves objects in the Y coordinate when the user is swapped so they're named `BATTLEANIMSTRUCT_FIX_Y` and `wBattleAnimTempFixY`

engine/battle_anims/functions.asm
Rangi42: I see. `Functionce60a` is between `Functionce5f9` and `move_up`, so `move_up` can't be local to `Functionce5f9`. The underlying issue is that *all* of those should be local labels: the various `call BattleAnim_AnonJumptable`s toward the top of the file have local pointers `.zero`, `.one`, `.two`, etc, but further down they're just `dw FunctionXXXX`. Ideally they would all follow the local `.zero`, etc pattern, when possible. (It won't always be; for example `BattleAnimFunction_PoisonGas` has `dw BattleAnimFunction_SpiralDescent` instead of `dw .one`, and some pieces might be reused by multiple battle anims; but as much as possible.)

7Soul: Done with those and some other unnamed local labels, except for a couple under `BattleAnimFunction_Gust.GustWobble` which I can't understand what they mean

7Soul:
Renamed BATTLEANIMFUNC_* constants

Rangi42:
The only unidentified labels left are `Functioncd557` and a couple of local `.asm_*`s in `.GustWobble`, which is fine, not everything has to be identified at once. I'll take a look at them after merging.

Just one more change request: please make the anonymous jumptables have unique entries that correspond to the index value. For example, `BattleAnimFunction_RazorLeaf` has nine entries from `.zero` to `.eight`, but `.four`, `.five`, `.six`, and `.seven` are all for the same code. This way you can look at a block and know what values the jumptable index might have when reading it. Whereas some of the updated functions are like `BattleAnimFunction_ThunderWave`, which has a duplicate `.zero` entry and the entry named `.two` corresponds to an index of `3`.

Rangi42:
(Three still have non-sequential labels: `BattleAnimFunction_ThunderWave`, `BattleAnimFunction_Gust`, and `BattleAnimFunction_Wrap`.)

mid-kid:
I've been too busy to properly review this, but just wanted to drop by to thank you for the effort. These files really needed some love.

PR #772, opened 2020-10-04
Identify anim object constants

7Soul:
Named the ANIM_OBJ constants

Some special notes:
`ANIM_OBJ_SONICBOOM_JP_UNUSED` Uses the same FRAMESET and a similar function to `ANIM_OBJ_SONICBOOM_JP` but this one isn't actually used
`ANIM_OBJ_CAKE_UNUSED` Is the only object that points to the cake sprite from GFX_OBJECTS, but isn't used
`ANIM_OBJ_B0` isn't used and has no obvious purpose. It appears to be a red version of Mist but with its own movement function

Also renamed one animation:
`BattleAnimSub_ShakeEnemy` to `BattleAnimSub_EyeBeams` since this animation branch is only responsible to creating the objects for animations such as Leer and Glare. The shaking is done by a BG effect prior to calling this animation


constants/battle_anim_constants.asm
```diff
@@ -26,34 +26,34 @@ BATTLEANIM_BASE_TILE EQU 7 * 7  ; Maximum size of a pokemon picture
...
-	const ANIM_OBJ_07
-	const ANIM_OBJ_08
+	const ANIM_OBJ_HIT_BIG_YFIX
+	const ANIM_OBJ_HIT_YFIX
+	const ANIM_OBJ_HIT_SMALL_YFIX
+	const ANIM_OBJ_HIT_BIG
+	const ANIM_OBJ_HIT
+	const ANIM_OBJ_HIT_SMALL
+	const ANIM_OBJ_PUNCH
+	const ANIM_OBJ_FOOT
```
Rangi42, line 36: IMO do `ANIM_OBJ_KICK` to match `ANIM_OBJ_PUNCH`.

7Soul: Done

PR #773, opened 2020-10-04
 Identify battle bg effects functions and constants

7Soul:
Named battle bg effects functions and constants

Most functions can take either $0 or $1 in the `BG_EFFECT_STRUCT_BATTLE_TURN` parameter of the `anim_bgeffect` macro, which dictates which side of the animation it affects ($0 being the target of the effect). Even though some animations only ever use one or the other, the ability to change is there. In those cases the term "Mon" was used in the function name, otherwise "Player" or "Enemy" is used

Because of this, `ANIM_BG_WOBBLE_MON` and `ANIM_BG_WAVE_DEFORM_USER` were renamed. There is a different function that took the place of `ANIM_BG_WOBBLE_MON`

Functions related to palette:
The term "Light" was used when the palette deals with the lightest shade of the pokémon and not white
Similarly the term "Dark" was used then the palette deals with the darkest shade but not black

A few functions related to Rollout, `Functionc88a5` and `Functionc88a5`, weren't named as I don't quite understand what makes the Rollout animation special

Also fixed a typo from my previous merge


constants/battle_anim_constants.asm
```diff
@@ -712,9 +712,9 @@ BATTLEANIM_BASE_TILE EQU 7 * 7  ; Maximum size of a pokemon picture
 	const ANIM_BG_WHITE_HUES
 	const ANIM_BG_BLACK_HUES
 	const ANIM_BG_ALTERNATE_HUES
-	const ANIM_BG_06
-	const ANIM_BG_07
-	const ANIM_BG_08
+	const ANIM_BG_CYCLE_OBPALS_GRAY_AND_YELLOW
+	const ANIM_BG_CYCLE_MID_OBPALS_GRAY_AND_YELLOW
+	const ANIM_BG_CYCLE_WBGPALS_INVERTED
```
Rangi42, line 717: `ANIM_BG_CYCLE_BGPALS_INVERTED`

engine/battle_anims/bg_effects.asm
```diff
@@ -2367,6 +2395,9 @@ BattleBGEffect_GetNthDMGPal:
 	ret
 
 BGEffect_RapidCyclePals:
+; Last index in DE:
+;                   $fe signals a loop
```
Rangi42, line 2399: No need for the deep indentation; `; $f...` or `; - $f...` is fine

Rangi42:
If the `$1` and `$0` appear directly in places in the code, then constants would be appropriate for them; something like `BGEFFECTSTRUCT_BATTLETURN_TARGET EQU 0` and `BGEFFECTSTRUCT_BATTLETURN_USER EQU 1`.

Issue #706, opened 2020-04-04
Add assertions throughout the code to provide guarantees (not just comments)


Rangi42:
Some code is written assuming that a constant has a certain value; or is divisible by some quantity; or that one struct field is a certain offset from another; or that two maps are in the same group; or two graphics in the same bank; and so on. These can be documented with `assert` better than with comments.

mid-kid:
The main issue I have with this is `xor a` for `ld a, 0` and `and a` for `cp 0`, where a certain constant is assumed to be 0. This is currently commented and I think asserts would hurt readability of these a lot.
As for divisibility, I'm pretty sure you're talking about alignment, and that can be better enforced by the linker using the `ALIGN` parameter like we're already doing. for const/enums we always just set the value for the next item, or change the increment, which I think is appropriate enough.
Similarly, graphics being in the same bank should be enforced in the same way that two functions being in the same bank is: Putting them in the same section. If this is not possible, do what #631 does and make a constant and hardcode the BANK[] of their respective sections, as a stopgap for https://github.com/rednex/rgbds/issues/244.

I believe `assert` should be a last-resort solution for things that can't be expressed in any "natural" way: Some function expecting two maps to be in the same group instead of checking their groups separately being the prime example.
Just be careful and avoid using asserts for things like `hli` where `CONST+1` must equal `CONST2` (if you change a structure you should look for all uses of it, if anything, the disassembly should make these sorts of structures more explicit instead of just loose wram labels), or to check an array is the right size. It's assumed that if you modify a structure or an array length, things will break, and comments are enough for easy grepability and readability.

Kroc:
If asserts can be placed in macros then to some extent, the asserts can be automatic in places where the assembly code makes assumptions about order / value. Trust me, you _do not_ want to be debugging an issue where an order or value assumption changes, it can be completely invisible in the code.

aaaaaa123456789:
I think obvious "none" constants don't need any assert. It should be obvious that `MUSIC_NONE` equals 0, and anyone who changes something that basic should be very aware of the effects anyway.

But there are many cases where a constant is 0 because it's simply the first field of a struct or the first element in an enum, and those might be worth asserting.

Rangi42:
[...]

PR #776, opened 2020-10-26
Comment, remove, or revise many unreferenced labels

Rangi42:
These were found via `unref.py`: <https://pastebin.com/airBsQbQ> There are probably more that it didn't catch.

Some WRAM labels are not directly referenced but refer to used space, like individual struct fields; those shouldn't be commented.

This syncs with https://github.com/pret/pokegold/pull/64.


entrpntr:
Had a bunch more, but most appear to be addressed by the latest commit. I think this is all I see left (in 3 broad categories):

1. Should be commented as unreferenced? (all from `engine/gfx/pic_animation.asm`; not familiar with this part of pokecrystal, but they seem to be so?)
    - `AnimateMon_Menu`
    - `AnimateMon_Trade`
    - `AnimateMon_Evolve`
    - `AnimateMon_Hatch`
    - `AnimateMon_HOF`

2. Could be converted to comments:
    - `BattleTowerTrainer[1-70]DataTable`
    - `BattleTowerMons[1-10]`

3. Miscellaneous (applies to both pokecrystal/pokegold):
    - Could combine `SGBBorderMap`+`SGBBorderPalettes` under 1 label? (like with `ChrisCard`+`ChrisPic` -> `ChrisPicAndTrainerCardGFX` in pokegold)
    - Could add assert of some sort before `cp TrophyIDs - DecorationIDs` in `engine/link/mystery_gift.asm`? (like with `BoxNameInputUpper - NameInputUpper`, since `TrophyIDs` only purpose seems to be in offset calculation?)

Rangi42:
`AnimateMon_*` are all referenced by the `pokeanim` macro. I've applied the rest of your comments. Thanks!

Rangi42:
Since this already started using `assert`, I went ahead and added more. There may be other places where assertions are appropriate, but this covers the basics that #706 listed.

PR #781, opened 2020-11-05
audio/engine.asm and home/audio.asm restyling

aljohnston112:
I would like some feedback on my restyling before continuing. Are there things I'm doing wrong or could be done another way?

### DIFF
--- a/audio/engine.asm
+++ b/audio/engine.asm
@@ -3,43 +3,56 @@
 ; Interfaces are in bank 0.
 
 ; Notable functions:
-; 	FadeMusic
-; 	PlayStereoSFX
+  ; FadeMusic
+  ; PlayStereoSFX
 
-_InitSound::
 ; restart sound operation
 ; clear all relevant hardware registers & wram
+_InitSound::
 	push hl
 	push de
 	push bc
 	push af
 	call MusicOff
-	ld hl, rNR50 ; channel control registers
+	; channel control registers
+	ld hl, rNR50 
 	xor a
-	ld [hli], a ; rNR50 ; volume/vin
-	ld [hli], a ; rNR51 ; sfx channels
-	ld a, $80 ; all channels on
-	ld [hli], a ; rNR52 ; music channels
-
-	ld hl, rNR10 ; sound channel registers
+	; rNR50 ; volume/vin
+	ld [hli], a 
+	; rNR51 ; sfx channels
+	ld [hli], a 
+	; all channels on
+	ld a, $80 
+	; rNR52 ; music channels
+	ld [hli], a 
+  ; sound channel registers
+	ld hl, rNR10
 	ld e, NUM_MUSIC_CHANS
+	
 .clearsound
-;   sound channel   1      2      3      4
+	; sound channels  
+	;   1      2      3      4
 	xor a
-	ld [hli], a ; rNR10, rNR20, rNR30, rNR40 ; sweep = 0
-
-	ld [hli], a ; rNR11, rNR21, rNR31, rNR41 ; length/wavepattern = 0
+	; rNR10, rNR20, rNR30, rNR40 ; sweep = 0
+	ld [hli], a
+	; rNR11, rNR21, rNR31, rNR41 ; length/wavepattern = 0
+	ld [hli], a
 	ld a, $8
-	ld [hli], a ; rNR12, rNR22, rNR32, rNR42 ; envelope = 0
+	; rNR12, rNR22, rNR32, rNR42 ; envelope = 0
+	ld [hli], a 
 	xor a
-	ld [hli], a ; rNR13, rNR23, rNR33, rNR43 ; frequency lo = 0
+	; rNR13, rNR23, rNR33, rNR43 ; frequency lo = 0
+	ld [hli], a 
 	ld a, $80
-	ld [hli], a ; rNR14, rNR24, rNR34, rNR44 ; restart sound (freq hi = 0)
+	; rNR14, rNR24, rNR34, rNR44 ; restart sound (freq hi = 0)
+	ld [hli], a 
+	; clear the rest of the channels
 	dec e
 	jr nz, .clearsound
 
 	ld hl, wAudio
 	ld de, wAudioEnd - wAudio
+	
 .clearaudio
 	xor a
 	ld [hli], a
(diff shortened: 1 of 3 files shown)


audio/engine.asm
```diff
@@ -1197,12 +1216,15 @@ ParseMusic:
...
 .sound_ret
-; $ff is reached in music data
 	ld hl, CHANNEL_FLAGS1
 	add hl, bc
-	bit SOUND_SUBROUTINE, [hl] ; in a subroutine?
-	jr nz, .readcommand ; execute
+	; in a subroutine?
+	bit SOUND_SUBROUTINE, [hl] 
+	; execute
+	jr nz, .readcommand
```
aaaaaa123456789, line 1227: If these comments weren't redundant, which they are, they should remain where they are. But the main problem is that the comments aren't needed in the first place — it should be obvious that `bit SOUND_SUBROUTINE, [hl]` is checking whether the engine is in a subroutine. These comments might've been necessary back when these were unnamed constants or simple numbers, but right now, that instruction is self-documenting, and so is the jump.

audio/engine.asm
```diff
@@ -1356,6 +1379,7 @@ ParseMusicCommand:
 	ld d, 0
 	; seek command pointer
 	ld hl, MusicCommands
+	; each command takes up two bytes
```
aaaaaa123456789, line 1382: This is pretty much implied by the fact that the command offset is doubled before being added to the pointer. Labelling all such indexing operations with a comment would create lots of noise; it's a very common pattern.

audio/engine.asm
```diff
@@ -1446,9 +1470,14 @@ Music_Ret:
...
 
-Music_Call:
 ; call music stream (subroutine)
-; parameters: ll hh ; pointer to subroutine
+; input: 
+	; bc = start of current channel
+; output: 
+	; CHANNEL_LAST_MUSIC_ADDRESS = CHANNEL_MUSIC_ADDRESS
+	; CHANNEL_MUSIC_ADDRESS = pointer to subroutine (ll hh)
+	; Sets the SOUND_SUBROUTINE flag in CHANNEL_FLAGS1
```
aaaaaa123456789, line 1479: `CHANNEL_LAST_MUSIC_ADDRESS` and `CHANNEL_MUSIC_ADDRESS` are constants, and thus, they cannot ever be outputs in the first place. In fact, all inputs and outputs must be in a register or a memory location, and those are neither. Are you sure you're understanding the function in the first place?

Also, this excess verbosity in function comments isn't helpful; we're not trying to run doxygen on the codebase. The files are already massive as it is; keep your comments succint and to the point.

audio/engine.asm
```diff
@@ -1479,9 +1508,13 @@ Music_Call:
...
 	ret
 
-Music_Jump:
 ; jump
-; parameters: ll hh ; pointer
+; input: 
+	; bc = start of current channel
+; output: 
+	; CHANNEL_MUSIC_ADDRESS = pointer to 
+	; music address to jump to (ll hh)
```
aaaaaa123456789, line 1516: Again, this excess verbosity isn't helpful. Keep it all in one line. (And again, `CHANNEL_MUSIC_ADDRESS` cannot be an output. Are you sure you're understanding what's going on?)

The important part here, after all, is that the two bytes that follow the jump command itself in the channel stream represent the jump target address, in little-endian. The original comment already conveyed that information. The rest can easily be left out.

audio/engine.asm
```diff
@@ -2101,10 +2134,13 @@ Music_NewSong:
...
 	ret
 
-GetMusicByte:
 ; returns byte from current address in a
 ; advances to next byte in music data
-; input: bc = start of current channel
+; input: 
+	; bc = start of current channel
+; output: 
+	; a = music byte
```
aaaaaa123456789, line 2142: Once again, cut down the verbosity. The function comment already said that the function returned in `a`, so there's no point in saying that twice.

audio/engine.asm
```diff
@@ -2117,19 +2153,18 @@ GetMusicByte:
 	ld hl, CHANNEL_MUSIC_BANK
 	add hl, bc
 	ld a, [hl]
-	; get byte
-	call _LoadMusicByte ; load data into wCurMusicByte
-	inc de ; advance to next byte for next time this is called
+	; load byte into wCurMusicByte
+	call _LoadMusicByte
+	; advance to next byte for next time this is called and
```
aaaaaa123456789, line 2158: By moving this comment outside of the line where it was, you have made the flow more obscure. Before, it was clear that `inc de` was simply incrementing the pointer for the next time, and that the following code was updating the channel data address. Now that's merged into a single comment, still broken down into two lines for some unexplainable reason.

home/audio.asm
```diff
@@ -48,8 +48,12 @@ UpdateSound::
 	pop hl
 	ret
 
+; input: 
+	; a = channel music bank
+	; de = channel music address
+; output:
+	; wCurMusicByte = [a:de]
```
aaaaaa123456789, line 55: This is all easily implied by the single comment that was already there. What's the benefit in this over-verbosity?

aaaaaa123456789 (changes requested):
Honestly, most of these changes seem like they'd be doing a net damage to the codebase. I see a lot of changes "just because Visual Studio says so" (comments above function definitions instead of inside, pointlessly verbose calling convention descriptions, and so on) and many changes made by blindly applying rules without considering why things were done the way they were. It's a shame, because `audio/engine.asm` is a very old file that could definitely use some love, but this isn't helping.

To maintainers: maybe the style guide needs an update? There are many places in which things deviate from that guide, and while they probably seem very intuitive for us, we're not the ones checking the guide in the first place. If this is the kind of thing that the style guide leads to, I'd rather see the file wholly deleted.

Rangi42:
Thank you for your interest in improving pokecrystal! However, this PR is not a great start to that.

I see ax6 is going over some particular lines, but my overall comments:

- IDE-specific files like `.vs/*` don't belong in `.gitignore`
- Don't indent comments by two spaces
- No need to be absolutely consistent about whether comments go above vs below a label, or above vs next to an opcode
- Some comments like "`store next byte in a`" for `call GetMusicByte`, or "`load byte into wCurMusicByte`" for `call _LoadMusicByte`, are just noise (note that some files disassembled early on, like the audio engine and much of pokered, have excessive comments already, since fewer labels and constants existed at the time)

Rangi42:
> To maintainers: maybe the style guide needs an update? There are many places in which things deviate from that guide, and while they probably seem very intuitive for us, we're not the ones checking the guide in the first place.

Yes, it should be revised. For example, short comments next to short code lines are okay; and there's no obvious best choice for comments going above vs below a function label, or comments above "paragraphs" of code being indented vs not, so those are  probably better off left to individual judgment than trying to standardize the whole codebase.

PR #783, opened 2020-11-09
Identified two magic numbers and cleaned up some comments

aljohnston112:


audio/engine.asm
```diff
@@ -2671,13 +2662,11 @@ PlayStereoSFX::
...
-; prep channel for use
-; input:
-; 	de:
-	; get pointer to current channel
+; input: de = audio pointer
+; sets bc to current channel pointer
 	call LoadMusicByte
 	inc de
-	and $7 ; bit 0-2 (current channel)
+	and NUM_MUSIC_CHANS ; bit 0-2 (current channel)
```
Rangi42, line 2669: This breaks `make compare`. It should be `maskbits NUM_CHANNELS`. Also, make the same change to the `and $7` in `UpdateChannels`.

Rangi42 (changes requested):
Your comment changes look alright; just fix the `make compare` failure. Also please make the same PR for pokegold.

audio/engine.asm
aljohnston112: Why would it break make compare?

aljohnston112: I didn't know to check make compare. Got it fixed.

audio/engine.asm
```diff
@@ -212,7 +212,7 @@ _UpdateSound::
 UpdateChannels:
 	ld hl, .ChannelFnPtrs
 	ld a, [wCurChannel]
-	and $7
+	maskbits NUM_MUSIC_CHANS ; bit 0-2 (current channel)
```
Rangi42: Since `NUM_MUSIC_CHANS` has a meaningful name, the comment is no longer necessary to explain the meaning of the `$7` value; likewise for the other line.

PR #792, opened 2020-12-15
Document fix for StatUpHit effects when opponent is KO'd

nbasu02:
As a kid I always wondered if you could increase stats when you KO'd a pokemon with ancientpower, but I was never sure until I was testing around an attacking move that would increase attack 100% of the time.  This fixed the issue.

I'm not sure if this is considered a bug or is just worth a page here https://github.com/pret/pokecrystal/wiki/Tutorials, but in my opinion it feels like it would've been pasted from the StatDownHit moves, which obviously don't need to decrease stats if the opponent is KO'd.

### DIFF
--- a/docs/bugs_and_glitches.md
+++ b/docs/bugs_and_glitches.md
@@ -767,53 +767,44 @@ This bug existed for all battles in Gold and Silver, and was only fixed for sing
 
 ### Moves that do damage and increase your stats do not increase stats after a KO
 
-The `checkfaint` routine skips the stats increasing effects if the opponent is KO'd, unlike in modern pokemon games.  Note that this can lead to stats being increased at the end of battle, but will not have any negative effects.
+`BattleCommand_CheckFaint` "ends the move effect if the opponent faints", and these moves attempt to raise the user's stats *after* `checkfaint`. Note that fixing this can lead to stats being increased at the end of battle, but will not have any negative effects.
 
-**Fix:** Make the following changes in [data/moves/effects.asm](https://github.com/pret/pokecrystal/blob/master/data/moves/effects.asm)
+**Fix:** Edit [data/moves/effects.asm](https://github.com/pret/pokecrystal/blob/master/data/moves/effects.asm):
 
 ```diff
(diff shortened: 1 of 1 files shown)


docs/bugs_and_glitches.md
```diff
@@ -2261,3 +2261,80 @@ This does not have a simple and accurate fix. It would involve redesigning parts
...
 ```
+
+### Moves that do damage and increase your stats do not increase stats after a KO
+
+The `checkfaint` routine skips the stats increasing effects if the opponent is KO'd, unlike in modern pokemon games.  Note that this can lead to stats being increased at the end of battle, but will not have any negative effects.
+
+**Fix:** Make the following changes in [data/moves/effects.asm](https://github.com/pret/pokecrystal/blob/master/data/moves/effects.asm)
+
+```diff
+ DefenseUpHit:
```
nbasu02, line 2272: I posted the full effects here but let me know if I should truncate

nbasu02:
Wait I need to put this into the right section...

docs/bugs_and_glitches.md
```diff
@@ -764,6 +765,82 @@ This bug existed for all battles in Gold and Silver, and was only fixed for sing
...
 ```
 
+### Moves that do damage and increase your stats do not increase stats after a KO
+
+The `checkfaint` routine skips the stats increasing effects if the opponent is KO'd, unlike in modern pokemon games.  Note that this can lead to stats being increased at the end of battle, but will not have any negative effects.
+
+**Fix:** Make the following changes in [data/moves/effects.asm](https://github.com/pret/pokecrystal/blob/master/data/moves/effects.asm)
+
+```diff
+ DefenseUpHit:
```
nbasu02, line 775: I posted the full effects here but let me know if I should truncate

aaaaaa123456789:
Putting it in the bugs document is correct IMO. You might want to truncate the effects list a bit, particularly because there's three of them and they are very long.

nbasu02:
@aaaaaa123456789 done

Issue #739, opened 2020-07-06
Consistent struct definitions, possibly using the RS commands


Rangi42:
We're currently not using this feature of rgbds. It's similar to our current `const` and `enum` macros, with the counter `_RS` instead of `const_value` or `__enum__`.

- `rsreset` sets `_RS = 0`
- `rsset N` sets `_RS = N`
- `FOO rb N` sets `FOO = _RS` and does `_RS += N`
- `FOO rw N` sets `FOO = _RS` and does `_RS += N * 2`

Their intended use is for defining struct offsets. We have no consistent pattern for those:

- a `const_def`: the most common; e.g. `ITEMATTR_*` in [constants/item_data_constants.asm](https://github.com/pret/pokecrystal/blob/master/constants/item_data_constants.asm)
- a sequence of `EQU`s: easier to have non-sequential values; e.g. `NPCTRADE_*` in [constants/npc_trade_constants.asm](https://github.com/pret/pokecrystal/blob/master/constants/npc_trade_constants.asm)
- label differences from a canonical struct usage: avoids redundantly defining a struct macro and offset constants; e.g. `BASE_*` in [constants/pokemon_data_constants.asm](https://github.com/pret/pokecrystal/blob/master/constants/pokemon_data_constants.asm)

@ISSOtm also points out that using the rs commands is faster to build than using `const` or `enum` macros.

I think that at least the `const_def` and `EQU`s patterns could be replaced by usage of `_RS`, and possibly the label offsets as well.

The above examples rewritten:

```
	rsreset
ITEMATTR_PRICE       rw 1
ITEMATTR_EFFECT      rb 1
ITEMATTR_PARAM       rb 1
ITEMATTR_PERMISSIONS rb 1
ITEMATTR_POCKET      rb 1
ITEMATTR_HELP        rb 1
ITEMATTR_STRUCT_LENGTH EQU _RS
```

```
	rsreset
NPCTRADE_DIALOG  rb 1
NPCTRADE_GIVEMON rb 1
NPCTRADE_GETMON  rb 1
NPCTRADE_NICK    rb MON_NAME_LENGTH
NPCTRADE_DVS     rw 1
NPCTRADE_ITEM    rb 1
NPCTRADE_OT_ID   rw 1
NPCTRADE_OT_NAME rb NAME_LENGTH
NPCTRADE_GENDER  rb 1
NPCTRADE_PADDING rb 1
```

```
	rsreset
BASE_DEX_NO rb 1
[...]

PR #806, opened 2021-03-04
Verify data table sizes with table_width and assert_table_length macros

Rangi42:
This was discussed in #706

It also uncovered some off-by-one issues with defining some constants.

A few structs now use rsreset/_RS to define their offset constants, as discussed in #739


constants/gfx_constants.asm
```diff
@@ -6,7 +6,7 @@ NUM_PAL_COLORS EQU 4
 PAL_COLOR_SIZE EQU 2
 PALETTE_SIZE EQU NUM_PAL_COLORS * PAL_COLOR_SIZE
 
-PALRGB_WHITE EQUS "palred 31 + palgreen 31 + palblue 31" ; $7fff
+PALRGB_WHITE EQU (palred 31 + palgreen 31 + palblue 31) ; $7fff
```
aaaaaa123456789, line 9: No need for parens here, as it's just a value (this is as bad as `return (a * b);` in C code).

Rangi42:
Once this is finalized it should also be applied to pokegold, pokered, and pokeyellow.

data/battle_tower/classes.asm
```diff
@@ -2,6 +2,7 @@ BattleTowerTrainers:
 ; The trainer class is not used in Crystal 1.0 due to a bug.
 ; Instead, the sixth character in the trainer's name is used.
 ; See BattleTowerText in engine/events/battle_tower/trainer_text.asm.
+	table_width NAME_LENGTH - 1 + 1, BattleTowerTrainers
```
aaaaaa123456789, line 5: `NAME_LENGTH - 1 + 1` is confusing; if this is unavoidable, I'd add a comment.

data/moves/effects_pointers.asm
```diff
@@ -157,3 +158,4 @@ MoveEffectsPointers:
 	dw BeatUp
 	dw Fly
 	dw DefenseCurl
+	check_table_length NUM_MOVE_EFECTS
```
aaaaaa123456789, line 161: No newline at end of file

data/pokemon/palettes.asm
```diff
@@ -5,6 +5,8 @@ PokemonPalettes:
 ; only the middle two colors are included, not black or white.
 ; Shiny palettes are defined directly, not generated.
 
+	table_width PAL_COLOR_SIZE * 2 * 2, PokemonPalettes
```
aaaaaa123456789, line 8: Again, `* 2 * 2` might be confusing and worth a comment (although this one's clearer).

engine/items/item_effects.asm
```diff
@@ -192,6 +193,7 @@ ItemEffects:
 	dw PokeBallEffect      ; PARK_BALL
 	dw NoEffect            ; RAINBOW_WING
 	dw NoEffect            ; ITEM_B3
+	check_table_length ITEM_B3
```
aaaaaa123456789, line 196: Why `ITEM_B3`? Is there any reason to stop here? Is there a more suitable constant for it?

aaaaaa123456789 (changes requested):
Overall a very positive change; I've marked a few odd things here and there.

engine/items/item_effects.asm
Rangi42: The items past `ITEM_B3`, from `BRICK_PIECE` to `ITEM_BE` (mostly Mails, plus the TMs and HMs), don't have effect pointers. ¯\\\_(ツ)_/¯ It doesn't cause problems because they're all `ITEMMENU_NOUSE`. I'll add a comment about that.

data/battle_anims/objects.asm
```diff
@@ -18,6 +18,7 @@ ENDM
 
 BattleAnimObjects:
 ; entries correspond to ANIM_OBJ_* constants
+	table_width BATTLEANIMOBJ_LENGTH, BattleAnimObjects
```
dannye, line 21: Should this table have an assert?

constants/pokemon_data_constants.asm
```diff
@@ -67,39 +69,40 @@ GENDER_UNKNOWN EQU -1
...
+MON_ATK_EXP            rw
+MON_DEF_EXP            rw
+MON_SPD_EXP            rw
+MON_SPC_EXP            rw
+MON_DVS                rw
+MON_PP                 rb NUM_MOVES
+MON_HAPPINESS          rb
+MON_PKRUS              rb
+MON_CAUGHTDATA         rw
+MON_CAUGHTLEVEL        EQU _RS - 2
```
dannye, line 89: Doing subtractions on `_RS` feels unclean. Would it be better to use `MON_CAUGHTDATA EQU _RS` to avoid this?

constants/pokemon_data_constants.asm
```diff
@@ -67,39 +69,40 @@ GENDER_UNKNOWN EQU -1
...
+MON_CAUGHTTIME         EQU _RS - 2
+MON_CAUGHTGENDER       EQU _RS - 1
+MON_CAUGHTLOCATION     EQU _RS - 1
+MON_LEVEL              rb
+BOXMON_STRUCT_LENGTH EQU _RS
```
dannye, line 94: Alignment

data/battle_anims/objects.asm
Rangi42: Yes, `NUM_ANIM_OBJS`.

constants/pokemon_data_constants.asm
Rangi42: There's no exact equivalent of `UNION`s, but there is `RSSET` to "rewind" it.

constants/pokemon_data_constants.asm
Rangi42: Typically the length/size `EQU` constants at the end of lists aren't aligned with the members.

dannye (approved):

dannye (approved):

PR #811, opened 2021-03-13
Fix assert off-by-one error

dannye:
This was preventing pokered's Lavender theme from being added to pokecrystal.


Rangi42:
I fixed this as part of some cleanup just now, didn't notice you already had a PR. :P

dannye:
Cool. The issue still exists in pokegold.

PR #820, opened 2021-04-03
Map POKé in the assembler

iimarckus:
Many years ago, I introduced `#` in strings rather than “POKé” directly for two reasons: before the game was fully disassembled I couldn’t be sure there weren’t literal uses of the characters ‘P,’ ‘O,’ ‘K,’ ‘é’ in sequence, and when I wrote the original text preprocessor I didn’t expect to ever use multi‐character strings in charmaps. These concerns are no longer relevant, so we should make the source more readable and easier to edit with this pull request.

There is one instance of ‘P,’ ‘O,’ ‘K,’ ‘é,’ of course: PlacePOKeText. I replaced it with `db "P","O","K","é","@"` so rgbasm wouldn’t detect it.

The `#` charmap should be left in, both for compatibility and because there are untranslated strings that still use it. `#` represents ポケモン in Japanese and shouldn’t be translated to POKé; I would also oppose translating ポケモン in a charmap until we’ve disassembled the Japanese versions and can confirm that it wouldn’t conflict with other uses of ‘ポ,’ ‘ケ,’ ‘モ,’ ‘ン’ there.


Rangi42:
See https://github.com/pret/pokered/pull/316#issuecomment-812957744

(If Gen 2 were the only one, or `charmap "POKé", $54` had always been around, I'd probably be fine with it; but since the Gen 1 games *do* need literal `P O K é`, and we've used "`#`" for years, IMO it's simpler to leave as-is. )

PR #829, opened 2021-04-26
"POKé" > "Poké"

Zemaille:
Lowercasing the # charmap to fit along the rest of the lowercasing.


kanzure:
But isn't the first way how it is rendered visually?

PR #831, opened 2021-05-05
Update bugs_and_glitches.md

RenaKunisaki:
correct label for CheckWarp - existing one isn't searchable


Rangi42:
This is incorrect; the label `.CheckWarp` is local inside the scope of `DoPlayerMovement`.

PR #839, opened 2021-06-20
Fixing comments in BattleCommand_Teleport

Nescientist:
srl is a right shift, etc.

### DIFF
--- a/engine/battle/move_effects/teleport.asm
+++ b/engine/battle/move_effects/teleport.asm
@@ -15,33 +15,28 @@ BattleCommand_Teleport:
 	call GetBattleVar
 	bit SUBSTATUS_CANT_RUN, a
 	jr nz, .failed
-; Only need to check these next things if it's your turn
 	ldh a, [hBattleTurn]
 	and a
 	jr nz, .enemy_turn
-; Can't teleport from a trainer battle
 	ld a, [wBattleMode]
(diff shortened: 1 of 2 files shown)


engine/battle/move_effects/teleport.asm
```diff
@@ -15,33 +15,28 @@ BattleCommand_Teleport:
...
 	call BattleRandom
-	cp c
+	cp c ; Generate a number between 0 and c
 	jr nc, .loop_player
-; If that number is greater than 4 times your level, run away.
-	srl b
 	srl b
-	cp b
-	jr nc, .run_away
+	srl b ; b = enemyLevel / 4
```
Rangi42, line 37: I think if a comment applies to more than one line, it should be an indented "heading comment" above all the lines. Here `; b = enemyLevel / 4` would go above both of the `srl b` lines. Likewise for some of these other comments.

Nescientist: Fair enough (though that would've mirrored the comments to pokered's SwitchAndTeleportEffect function).

Rangi42: pokecrystal was disassembled after pokered and has more polished labels and comments; generally they should be ported *to* pokered, not *from* it, to keep them in sync.

Issue #807, opened 2021-03-05
Bug: AI doesn't favor Solar Beam, Flame Wheel and Moonlight during Harsh Sunlight


Idain:
When checking [sunny_day_moves.asm](https://github.com/pret/pokecrystal/blob/master/data/battle/ai/sunny_day_moves.asm) I noticed that the moves Solar Beam, Flame Wheel and Moonlight aren't in the list of moves preferred by the AI during harsh sunlight:

```
SunnyDayMoves:
	db FIRE_PUNCH
	db EMBER
	db FLAMETHROWER
	db FIRE_SPIN
	db FIRE_BLAST
	db SACRED_FIRE
	db MORNING_SUN
	db SYNTHESIS
	db -1 ; end
```
The fix is simple: Include these three moves into the list, respecting how they're ordered in [move_constants.asm](https://github.com/pret/pokecrystal/blob/master/constants/move_constants.asm).

I hope this is added to [bugs_and_glitches.md](https://github.com/pret/pokecrystal/blob/master/docs/bugs_and_glitches.md).

ShadowOne333:
This is the [sunny_day_moves.asm](https://github.com/pret/pokecrystal/blob/master/data/battle/ai/sunny_day_moves.asm) file properly fixed with the changes suggested by Idain.
If needed, I can make the PR to add this into the [bugs_and_glitches.md](https://github.com/pret/pokecrystal/blob/master/docs/bugs_and_glitches.md)

```
; AI_SMART prefers these moves during harsh sunlight.

SunnyDayMoves:
	db FIRE_PUNCH
	db EMBER
	db FLAMETHROWER
	db SOLARBEAM
	db FIRE_SPIN
	db FIRE_BLAST
	db FLAME_WHEEL
	db SACRED_FIRE
	db MORNING_SUN
	db SYNTHESIS
	db MOONLIGHT
	db -1 ; end
```

coltongit:
Fix implemented in #849

PR #849, opened 2021-09-02
Add fix for Sunny Day moves used by the "Smart" AI

coltongit:
Fixes https://github.com/pret/pokecrystal/issues/807

### DIFF
--- a/docs/bugs_and_glitches.md
+++ b/docs/bugs_and_glitches.md
@@ -58,6 +58,7 @@ Fixes in the [multi-player battle engine](#multi-player-battle-engine) category 
   - [Wild Pokémon can always Teleport regardless of level difference](#wild-pokémon-can-always-teleport-regardless-of-level-difference)
   - [`HELD_CATCH_CHANCE` has no effect](#held_catch_chance-has-no-effect)
   - [Credits sequence changes move selection menu behavior](#credits-sequence-changes-move-selection-menu-behavior)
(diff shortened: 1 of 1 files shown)


ElectroDeoxys:
The section title is inaccurate. The AI doesn't discourage these moves at all. The AI uses this list to figure out whether it's advantageous to use Sunny Day. So if the Enemy's Pokémon has any of the moves listed, it will encourage the use of Sunny Day.

coltongit:
right, my wording must've been a bit off

coltongit:
@ElectroDeoxys fixed the misnomer, hopefully this is better

ElectroDeoxys:
It's still misleading. The AI _doesn't_ encourage these moves. Like I said, it's a list of moves to figure whether it should encourage Sunny Day or not.

coltongit:
There isn't much of a better way to describe this than what's currently there now. Hopefully now it isn't misleading.

mid-kid:
"Sunny Day-encouraging moves" → "moves that encourage Sunny Day"

I was confused as well, please change this.

coltongit:
> 
> 
> "Sunny Day-encouraging moves" → "moves that encourage Sunny Day"
> 
> I was confused as well, please change this, and squash commits.

Change applied, I forget if you can git squash multiple commits already committed, been a while

mid-kid:
That's the exact purpose of squashing, but sure, I'll do it.

PR #850, opened 2021-09-03
Fix Fast Ship bug with Town Map

### LINKED ISSUE #812: Bug: Town Map does not place the cursor on the Fast Ship icon
From SatoMew:

Possible oversight in `TownMap_InitCursorAndPlayerIconPositions`:

```
.FastShip:
    ld [wPokegearMapPlayerIconLandmark], a
    ld a, LANDMARK_NEW_BARK_TOWN
    ld [wPokegearMapCursorLandmark], a
    ret
```

Should be:

```
.FastShip:
    ld [wPokegearMapPlayerIconLandmark], a
    ld a, LANDMARK_FAST_SHIP
    ld [wPokegearMapCursorLandmark], a
    ret
```

Original:

image

Intended:

image

coltongit:
Fixes #812

### DIFF
--- a/docs/bugs_and_glitches.md
+++ b/docs/bugs_and_glitches.md
@@ -87,6 +87,7 @@ Fixes in the [multi-player battle engine](#multi-player-battle-engine) category 
   - [`CheckOwnMon` only checks the first five letters of OT names](#checkownmon-only-checks-the-first-five-letters-of-ot-names)
   - [`CheckOwnMonAnywhere` does not check the Day-Care](#checkownmonanywhere-does-not-check-the-day-care)
   - [The unused `phonecall` script command may crash](#the-unused-phonecall-script-command-may-crash)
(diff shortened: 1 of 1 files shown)


coltongit:
This PR fixes the bug with the Fast Ship not setting the cursor to its own icon when using the Town Map with it.

Rangi42:
This looks intentional, not an oversight. If they left out the `cp LANDMARK_FAST_SHIP / jr z, .FastShip` check entirely, `wPokegearMapPlayerIconLandmark` and `wPokegearMapCursorLandmark` would both be set to `LANDMARK_FAST_SHIP`. I expect this special case was to avoid weird behavior when cycling the cursor forward/back from the Fast Ship position. Have you tested this fix in-game?

vulcandth:
I tested this in game; It does indeed place the cursor on the `Fast Ship`; however when you navigate the cursor away from it; you are unable to navigate the cursor back to it, until your close and reopen your pokegear.

Rangi42:
I think if it requires a more complex "fix", then this isn't really a bug. The presence of the Fast Ship icon makes it clear that's where you are, they didn't intend the cursor to possibly reach it.

PR #852, opened 2021-09-06
Document `PAR`/`BRN` stat debuff fix

### LINKED ISSUE #746: Bug: PRZ and BRN stat debuffs are not always applied to enemy mon when switching
Found a peculiar behavior of the battle engine while testing a ROM hack that features more aggressive AI switching that, as far as I can tell, applies to the base game:

### Problem Statement
Under certain conditions, an enemy mon can be paralyzed without the 3/4 reduction to speed, or burned without the 1/2 reduction to attack

### Scenario
1. An enemy trainer's mon is paralyzed or burned
1. The enemy trainer AI switches that mon out for another
1. The new mon that has been switched in faints
1. The AI selects the mon with PRZ/BRN as its next choice
1. The mon with PRZ/BRN now has its full speed or attack
*Note that this bug oddly does not apply if the enemy trainer AI switches the PRZ/BRN mon in when its current mon has not fainted*

### Investigation
The logic to apply the debuffs based on PRZ and BRN status effects is the function `ApplyStatusEffectOnEnemyStats` in `engine/battle/core.asm`. I attached a debugger to the game, setting a breakpoint on the address of this function. I found that it *did not* break when switching in a paralyzed or burned mon after a faint, but *did* during an AI switch.

### Remedy
`ApplyStatusEffectOnEnemyStats` is only ever called from the `InitEnemyMon` function. I hypothesized that this pathway is only conditionally taken, and this bug could be fixed by moving the call to somewhere that is unconditionally called whenever an enemy mon is thrown out. I made the somewhat inelegant change of moving the call to `ApplyStatusEffectOnEnemyStats` to the `Function_SetEnemyMonAndSendOutAnimation` function, after which I attached my debugger and reproduced the scenario, observing that the enemy stat debuffs applied under all switch-in/switch-out conditions.
**Related Commit**: https://github.com/peterjtalen/Pokemon_Super_Crystal/commit/e2e199a61e334bd82dbf5903bc8f86e03ad8cfbd

coltongit:
This PR adds documentation for the bug where the stat debuffs for paralysis and burns, under certain circumstances, aren't applied, and aims to have a solution to fix this particular bug.

Fixes #746 

This is a draft until the actual issue that correlates to this is debugged and tested more thoroughly.

Fix based on https://github.com/PeteyMcFly/Pokemon_Super_Crystal/commit/e2e199a61e334bd82dbf5903bc8f86e03ad8cfbd


Idain:
Looking at "InitEnemyMon", looks like it happens because it's only called during Link Battles and the Battle Tower, so it'll never apply to AI Trainers. It only applies to them when `CalcEnemyStats` is called, which only happens when using a held status healing item, raising/lowering/resetting stats, healing a Pokémon or using the moves Heal Bell and Psyche Up.

Your draft is reasonable, but you shouldn't delete `ApplyStatusEffectOnEnemyStats` from `InitEnemyMon`.

coltongit:
so readd that one line and all is good?

coltongit:
@Idain ready for testing

coltongit:
No problemo

Idain:
Btw, I thought you'd change your Github username to your new name.

coltongit:
Haven't yet, will soon enough, just it's a little complicated right now for reasons I'd rather keep private

coltongit:
Any word on how testing went with this yet @Idain?

docs/bugs_and_glitches.md
```diff
@@ -1560,6 +1561,25 @@ The `[hInMenu]` value determines this button behavior. However, the battle moves
 ```
 
 
+### AI Pokémon sometimes do not carry over stat debuffs when PAR/BRN
+
+Sometimes during battle, the AI doesn't get inflicted with the stat debuffs associated with PAR/BRN, especially when switched out and back in for any reason.
```
Rangi42, line 1566: Sounds like we don't understand the circumstances of this bug well enough to accurately describe what it is, let alone how to fix it. How was it observed? Is there some old video of it happening? Has it appeared in GSC bug lists for years and is only now being noticed by pret?

coltongit: I'd recommend asking the person who found this bug in the first place, @Iandesherr

coltongit:
@landesherr, has there been any progress on a documentation video or anything of the sort for the bugdocs on this yet?

Rangi42:
Closing this since there are open questions left to resolve in the issue (what are the exact circumstances, and what is a complete and accurate fix; a video or other source of the bug would also help).

Issue #856, opened 2021-10-16
Edit the GitHub Actions workflow to only run under pret


Rangi42:
Personal forks of pokecrystal (and gold, red, and yellow) shouldn't keep running a `make compare` CI build that will almost certainly fail.

Idain:
That explains why it always fails for me. Huh.

mid-kid:
We alread have `github.repository_owner == 'pret'` checks in the actions.yml file, this should be trivial.

PR #858, opened 2021-10-22
feat():Launch make compare only for pret repo and not fork #856

KqesaR:
Related to the suggestion of @Rangi42 on issue #856 to edit github actions to launch `make compare` only for the original repository and not on forks repository


Rangi42:
This ran `make compare` despite being a PR, so it looks good. Thank you!

PR #866, opened 2022-01-07
Replace hardcoded number

Idain:
I noticed this hardcoded value some time ago, and there are probably more cases in this file. If that's the case, please let me know to add more commits to this PR.

### DIFF
--- a/engine/battle/core.asm
+++ b/engine/battle/core.asm
@@ -3980,10 +3980,11 @@ InitEnemyMon:
 	inc de
 	ld a, [hl]
 	ld [de], a
-; The enemy mon's base Sp. Def isn't needed since its base Sp. Atk is also used to calculate Sp. Def stat experience.
+	; The enemy mon's base Sp. Def isn't needed since its base
+	; Sp. Atk is also used to calculate Sp. Def stat experience.
 	ld hl, wBaseStats
 	ld de, wEnemyMonBaseStats
-	ld b, NUM_STATS -1
+	ld b, NUM_STATS - 1
 .loop
 	ld a, [hli]
 	ld [de], a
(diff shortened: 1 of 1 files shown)


Rangi42:
`wBaseStats` and `EnemyMonBaseStats` are structs of 6, bytes, not 5; `NUM_STATS` is 6. Why does this loop stop before copying SpDef? Is it a bug? Are there any moves that work with the mons' base SpDef, and do they use the wrong value due to this?

Idain:
`b` decreases to 0 when it copies the value from `wBaseSpecialDefense` to the last byte of `wEnemyMonStats`, making the loop stop and working as expected.

Idain:
I remember that `wEnemyMonBaseStats` is used for Stat Exp. calculation, so that explains why it doesn't copy the Pokémon's base Sp. Def. (the enemy's base Sp. Atk is added to both the user's Sp. Atk and Sp. Def).

Rangi42:
> I remember that `wEnemyMonBaseStats` is used for Stat Exp. calculation, so that explains why it doesn't copy the Pokémon's base Sp. Def. (the enemy's base Sp. Atk is added to both the user's Sp. Atk and Sp. Def).

Good to know! In that case, please use `NUM_STATS - 1` for the `5` constant, with a comment explaining that Sp. Def doesn't need copying since the enemy mon's base Sp. Atk is used to calculate Sp. Def stat experience.

Idain:
Something similar happens in lines [3983](https://github.com/pret/pokecrystal/blob/fabdd8e0196d1c50ee0a4a40823d0f8a0372d93c/engine/battle/core.asm#L3983) and [6368](https://github.com/pret/pokecrystal/blob/fabdd8e0196d1c50ee0a4a40823d0f8a0372d93c/engine/battle/core.asm#L6368). Do you think we should also replace them with `NUM_STATS - 1`?

Rangi42:
> Something similar happens in lines [3983](https://github.com/pret/pokecrystal/blob/fabdd8e0196d1c50ee0a4a40823d0f8a0372d93c/engine/battle/core.asm#L3983) and [6368](https://github.com/pret/pokecrystal/blob/fabdd8e0196d1c50ee0a4a40823d0f8a0372d93c/engine/battle/core.asm#L6368). Do you think we should also replace them with `NUM_STATS - 1`?

Definitely yes, especially since there's that puzzled "`; Only the first five base stats are copied..`" comment which can now be more informative.

Idain:
Done, @Rangi42. Feel free to further review my PR and merge it when you can.

PR #870, opened 2022-01-16
Enumerate MobileAPI function constants

mid-kid:
What do you mean, of course I won't actually go and name them all anytime soon.


Rangi42:
If you want to squash+merge this as-is, go ahead.

PR #872, opened 2022-01-25
Document Return/Frustration bug and its fix

Idain:
I'm surprised this hasn't been documented until now.


Idain:
UPdate: This also happens when the happiness value is 1-2 (Return) and 253-254 (Frustration). I forgot to clarify this.

Idain:
@Rangi42, I accidentally only put two hash signs in front of the bugfix title instead of three. Oops.

PR #877, opened 2022-02-13
IlexForestKurtStepsDownMovement Capitilzation Typo

vulcandth:
Noticed this while attempting to port the GS-Event to Gold/Silver

The O in IlexFOrestKurtStepsDownMovement is captilized, when it should be lowercase to match the Variable Naming Convention.

Corrected the name, and tested the whole event in game to verify functionality.


Idain:
Nice catch!

PR #880, opened 2022-02-15
Clarify how wCurBattleMon works

Idain:


Rangi42:
It's for symmetry with `wCurPartyMon`, but frankly I'd rather remove that comment too. The label already tells you it's the cur(rent) (Poke)mon, and indexes naturally start from 0.

Idain:
1. It saves time so that one doesn't have to look at the code to figure out how it works.
2. I was replicating the style `wCurPartyMon`'s comment used. I can reduce it to two lines or even one if necessary.

Idain:
> 

You'd be surprised how easy people can get confused. I'd rather leave that comment (although it could be reduced to two lines if necessary).

Rangi42:
Alright, please do reformat both to two lines. Also can you open PRs for pokegold too when they're applicable to it?

aaaaaa123456789:
It could be much shorter and not even two lines: `; index of the player's mon currently in battle (0-5)`

Idain:
Ok, both comments have been reformated.

Idain:
Mmm. The maybe something like `; index of mon's current party location` could work?

mid-kid:
Nitpick, but I prefer
```
; comment
wVar:: db
```

The file currently contains a bit of both styles which bothers me.

Also "current" is about as descriptive as "buffer", it doesn't really denote what its purpose is. One "current" means "storage for value that reflects battle state" while the other "current" means "function parameter with no relevance to the game state". This definitely deserves a comment.

As for a comment, `; index of the mon in the party that's being processed (0-5)` maybe?

Idain:
That comment works for me? Anyone else has something to say?

Issue #881, opened 2022-02-21
"palfix.py" doesn't work with Python 3.9+


Idain:
This happens because palfix.py imports png.py from pokemontools, which uses a method called "tostring()". This method has been deprecated and replaced with "tobytes()" in Python 3. This is problematic for new people trying to use the palfix.py script, especially for those with very little programming knowledge.

The solution is simple and I would've made a PR right away, but I wanted to know how many people still use Python 2.7 and their opinions about this.

mid-kid:
Specifically this is about python 3.9+, as this script has always been compatible with python 3, it only broke recently. From what I gathered, the changed equivalent for this function was backwards compatible, but it should be tested.

Alternatively, this might be a good excuse to update png.py with the newest version of PyPNG. Since upstream keeps this module compatible with many python versions, it saves us having to test.

Rangi42:
The latest png.py requires Python 3, so tools/pokemontools/lz.py, tools/pokemontools/gfx.py, and tools/gfx.py would all need updating to not assume Python 2 (e.g. `print` function vs statement). We may as well make sure that all the other tools/\*.py scripts work with Python 3 as well.

Idain:
I don't mind the change as long as every Python script works with Python 3. I can't speak about the compression algorithms you mentioned since I never worked with them before, so I'll leave the decision up to you guys.

Idain:
Are both the latest png.py and dpcomp algorith easily replaceable or do they need some extra adjustments to work with pokecrystal?

Rangi42:
- current gfx.py algorithm: already written in Python, but compresses poorly
- lzss/mpcomp algorithm: relatively simple, especially without the matching-behavior flags, but needs translating from C
- dpcomp: more complicated, and needs translating from C, but gives best compression

Idain:
Is compression a priority for the moment? Because in that case we could use mpcomp in the meantime. Also we can adapt the other Python scripts in a short time for Python 3 users.

Rangi42:
Priority order:

1. Get palfix.py to work
2. Get gfx.py able to decompress 1/2bpp to PNG
3. Get gfx.py able to convert PNG to 1/2bpp (rgbgfx does this anyway)
4. Get gfx.py able to compress 1/2bpp to LZ (lzcomp does this anyway)
5. Make gfx.py's compression efficient

Idain:
[...]

PR #889, opened 2022-03-13
Add a `bpp2png` tool

Rangi42:
With this we can remove gfx.py, since its only remaining use case is to convert 1/2bpp into PNG. Then png.py will only be needed for palfix.py; we can either upgrade png.py as per #881, or rewrite palfix.py in C too and remove png.py. (I recommend the former; it's not worth a rewrite since the rgbgfx C++ rewrite might make it obsolete.)


Rangi42:
This inherits gfx.py's behavior of assuming 1bpp iff the filename ends in ".1bpp". Not sure if this should be an explicit depth flag like rgbgfx and gfx.c instead.

aaaaaa123456789:
> This inherits gfx.py's behavior of assuming 1bpp iff the filename ends in ".1bpp". Not sure if this should be an explicit depth flag like rgbgfx and gfx.c instead.

It should be an explicit flag for both 1 and 2 bits per pixel, defaulting to checking the extension if neither are supplied.

mid-kid:
Suggestion: make a lib/ directory for the libraries.

I still have trouble seeing the point to this when so many other tools exist for the task - it doesn't have to be within the repo. But if you *really* want to do this, at least make sure all of the files in the repository can be reconstructed losslessly.

> This inherits gfx.py's behavior of assuming 1bpp iff the filename ends in ".1bpp".

There were more behaviors like this, like having `.w32.2bpp` for a 32 tiles wide image (I think, please make sure). But ultimately, I think this is better handled by command-line flags, so there's not need to support any of this.

Rangi42:
Thanks. I've added a `-d` flag like rgbgfx and gfx.c, but left the file extension detection as a convenient fallback.

I'm not aware of any other tools for converting 1/2bpp to PNG (none of rgbgfx, superfamiconv, or gfx.c do so), except for gfx.py. And I want to get rid of gfx.py because most of what it does is now redundant (lzcomp does compression and decompression, rgbgfx converts PNG to 1/2bpp), and because it would be a hassle to convert from Python 2 to 3 (which we ought to do since PyPNG updated).

Losslessly reconstructing all the PNG files in the repo is not possible. If a PNG converted to a 2bpp used `--remove_xflip/--remove_yflip`, or `--remove_whitespace`, or so on, you can't tell where to introduce duplicate tiles. And of course `pokemon_animation_graphics` does even more processing; reversing it would [require](https://github.com/Rangi42/polishedcrystal/blob/master/utils/mon_anims.py) reading frames.asm and bitmasks.asm, which I think would be way overcomplicated for this tool.

One feature I might add is a `-i/--interleave` flag to reverse interleaving for 8x16 OAM graphics.

mid-kid:
>I'm not aware of any other tools for converting 1/2bpp to PNG (none of rgbgfx, superfamiconv, or gfx.c do so)

superfamiconv does...

mid-kid:
> Losslessly reconstructing all the PNG files in the repo is not possible. If a PNG converted to a 2bpp used [...]

I meant at least the files which don't lose any information in the png->2bpp conversion.

Rangi42:
The output is 2-bit grayscale, which matches what gfx.py and palfix.py output; it's what the repo uses already, since that maps naturally to the information available in 2bpp.

Rangi42:
> superfamiconv does... it just outputs the colors in reverse iirc so it might be useful to adapt it upstream or smth.

Is it an [undocumented](https://github.com/Optiroc/SuperFamiconv/blob/master/README.md) feature then?

mid-kid:
I mentioned a lib/ directory mostly so common.h and others could also go in there, but this is fine too.

Rangi42:
Alright, this now supports reading a .gbcpal file and outputting a 4-color indexed PNG.

```bash
tools/bpp2png -t -w 112 -p gfx/pokemon/charmander/front.gbcpal gfx/tilesets/johto.2bpp johto_red_transposed.png
```

tools/bpp2png.c
```diff
@@ -0,0 +1,238 @@
...
+		*width = pixels == 5 * 5 * 8 * 8 ? 5 * 8 // 5x5 mon pic
+			: pixels == 6 * 6 * 8 * 8 ? 6 * 8 // 6x6 mon front/back pic
+			: pixels == 7 * 7 * 8 * 8 ? 7 * 8 // 7x7 mon/trainer pic
+			: pixels == 2 * 4 * 8 * 8 ? 2 * 8 // 2x4 mon icon
+			: pixels == 2 * 12 * 8 * 8 ? 2 * 8 // 2x12 walking sprite
+			: pixels == 2 * 6 * 8 * 8 ? 2 * 8 // 2x6 standing sprite
+			: pixels == 2 * 2 * 8 * 8 ? 2 * 8 // 2x2 still sprite
+			: pixels == 4 * 4 * 8 * 8 ? 4 * 8 // 4x4 big sprite
+			: pixels < 16 * 8 * 8 ? (unsigned int)(pixels / 8)
+			: 128;
```
Rangi42, line 96:
```suggestion
			: pixels > 16 * 8 * 8 ? 16 * 8 // maximum width
			: (unsigned int)(pixels / 8);
```

tools/bpp2png.c
```diff
@@ -0,0 +1,238 @@
...
+	uint8_t *pal_data = read_u8(filename, &filesize);
+	if (filesize != 4 * 2) {
+		error_exit("Invalid .gbcpal file: \"%s\"\n", filename);
+	}
+	for (int i = 0; i < 4; i++) {
+		uint8_t b1 = pal_data[i * 2], b2 = pal_data[i * 2 + 1];
+#define RGB5_TO_RGB8(x) (uint8_t)((x) * 33 / 4)
+		palette[i][0] = RGB5_TO_RGB8(b1 & 0x1f); // red
+		palette[i][1] = RGB5_TO_RGB8(((b1 & 0xe0) >> 5) | ((b2 & 0x03) << 3)); // green
+		palette[i][2] = RGB5_TO_RGB8((b2 & 0x7c) >> 2); // blue
```
Rangi42, line 157:
```suggestion
		palette[i][0] = ((b1 & 0x1f) << 3) | ((b1 & 0x1c) >> 2); // red
		palette[i][1] = ((b2 & 0x03) << 6) | ((b1 & 0xe0) >> 2) | ((b2 & 0x03) << 1) | ((b1 & 0x80) >> 7); // green
		palette[i][2] = ((b2 & 0x7c) << 1) | ((b2 & 0x70) >> 4); // blue
```

Rangi42: Comparison: <https://godbolt.org/z/bPGovEc8r>

tools/bpp2png.c
```diff
@@ -0,0 +1,238 @@
...
+
+void read_gbcpal(Palette palette, const char *filename) {
+	long filesize;
+	uint8_t *pal_data = read_u8(filename, &filesize);
+	if (filesize != 4 * 2) {
+		error_exit("Invalid .gbcpal file: \"%s\"\n", filename);
+	}
+	for (int i = 0; i < 4; i++) {
+		uint8_t b1 = pal_data[i * 2], b2 = pal_data[i * 2 + 1];
+#define RGB5_TO_RGB8(x) (uint8_t)((x) * 33 / 4)
```
Rangi42, line 154:
```suggestion
#define RGB5_TO_RGB8(x) (uint8_t)(((x) << 3) | ((x) >> 2))
```

tools/bpp2png.c
```diff
@@ -0,0 +1,238 @@
...
+	if (!*width) {
+		// If no width is specified, try to guess an appropriate one
+		*width = pixels == 5 * 5 * 8 * 8 ? 5 * 8 // 5x5 mon pic
+			: pixels == 6 * 6 * 8 * 8 ? 6 * 8 // 6x6 mon front/back pic
+			: pixels == 7 * 7 * 8 * 8 ? 7 * 8 // 7x7 mon/trainer pic
+			: pixels == 2 * 4 * 8 * 8 ? 2 * 8 // 2x4 mon icon
+			: pixels == 2 * 12 * 8 * 8 ? 2 * 8 // 2x12 walking sprite
+			: pixels == 2 * 6 * 8 * 8 ? 2 * 8 // 2x6 standing sprite
+			: pixels == 2 * 2 * 8 * 8 ? 2 * 8 // 2x2 still sprite
+			: pixels == 4 * 4 * 8 * 8 ? 4 * 8 // 4x4 big sprite
```
Rangi42, line 94:
```suggestion
#define GUESS_SIZE(w, h) pixels == (w) * (h) * 8 * 8 ? (w) * 8
		*width = GUESS_SIZE(5, 5) // mon pic
			: GUESS_SIZE(6, 6) // mon front/back pic
			: GUESS_SIZE(7, 7) // mon/trainer pic
			: GUESS_SIZE(2, 4) // mon icon
			: GUESS_SIZE(2, 12) // walking sprite
			: GUESS_SIZE(2, 6) // standing sprite
			: GUESS_SIZE(2, 2) // still sprite
			: GUESS_SIZE(4, 4) // big sprite
```

PR #890, opened 2022-03-14
Replace flawed Metal Powder bugfix

Idain:
The Metal Powder fix doesn't work properly because if the user's attack (stored in `b`) is greater than `HIGH(MAX_STAT_VALUE)` (which is $3), it will set the value to 3 and the opponent's defense (stored in `c`) to `LOW(MAX_STAT_VALUE)` ($E7).

The person who implemented the fix treated it as if it was similar to `ThickClubBoost` and `LightBallBoost`, which is not the case.

With simply [fixing TruncateHL_BC](https://github.com/pret/pokecrystal/blob/master/docs/bugs_and_glitches.md#reflect-and-light-screen-can-make-special-defense-wrap-around-above-1024) for Multiplayer battles is enough, since it's called before `DittoMetalPowder`.

### DIFF
--- a/docs/bugs_and_glitches.md
+++ b/docs/bugs_and_glitches.md
@@ -18,7 +18,6 @@ Fixes in the [multi-player battle engine](#multi-player-battle-engine) category 
 - [Multi-player battle engine](#multi-player-battle-engine)
   - [Perish Song and Spikes can leave a Pokémon with 0 HP and not faint](#perish-song-and-spikes-can-leave-a-pokémon-with-0-hp-and-not-faint)
   - [Thick Club and Light Ball can make (Special) Attack wrap around above 1024](#thick-club-and-light-ball-can-make-special-attack-wrap-around-above-1024)
-  - [Metal Powder can increase damage taken with boosted (Special) Defense](#metal-powder-can-increase-damage-taken-with-boosted-special-defense)
   - [Reflect and Light Screen can make (Special) Defense wrap around above 1024](#reflect-and-light-screen-can-make-special-defense-wrap-around-above-1024)
   - [Moves with a 100% secondary effect chance will not trigger it in 1/256 uses](#moves-with-a-100-secondary-effect-chance-will-not-trigger-it-in-1256-uses)
   - [Belly Drum sharply boosts Attack even with under 50% HP](#belly-drum-sharply-boosts-attack-even-with-under-50-hp)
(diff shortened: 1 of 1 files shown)


Rangi42:
This is a real bug, as seen in the video; it shouldn't be removed. Please add a proper fix, or at least "*To do: document a fix.*".

Idain:
> 

By applying the Truncate fix you avoid any situation where the attack/defense values provided to `DittoMetalPowder` would be wrong, but if you want a specific fix without referencing `TruncateHL_BC`... Yeah, I'll leave your note in the meantime.

Idain:
I can't think of a solution where `TruncateBC_HL` isn't involved for the moment, but I left the note in case someone wants to help.

docs/bugs_and_glitches.md
```diff
@@ -180,38 +180,7 @@ Fixes in the [multi-player battle engine](#multi-player-battle-engine) category
...
-+	ret nz
-+	ld a, LOW(MAX_STAT_VALUE)
-+	cp c
-+	ret nc
-+
-+.cap
-+	ld bc, MAX_STAT_VALUE
- 	ret
-```
+_TODO: Document a proper fix._
```
Rangi42, line 183:
```suggestion
([Video](https://www.youtube.com/watch?v=rGqu3d3pdok&t=450))

_TODO: Document a proper fix._
```

(I'd also just leave this open and see if one of us has a fix, first.)

Idain: If we don't change it more new people will keep following the flawed guide and won't know why it's not working properly.

Rangi42:
@Idain ...Are you sure this is even broken? Did you actually observe it misbehaving in the debugger? When `DittoMetalPowder` returns, `bc` *is* used to store the defender's Defense or Special Defense (which Metal Powder may have increased), so I'm not sure why you're saying the user's Attack is in `b`. The attacker's Attack or Special Attack is in `hl`.

Idain:
Because `DittoMetalPowder` is called _after_ the stats have been truncated to one byte each. (`hl` being truncated to `b` and `bc` to `c`).

Rangi42:
Okay, so what `DittoMetalPowder` is doing is doubling the defense in `c`, or if that overflows, halving the attack in `b` instead. A fix with "more accurate damage calculations" like for Thick Club and Light Ball would move the `call DittoMetalPowder` to before the `call TruncateHL_BC`, and change it to simply double the 16-bit defense in `bc`.

Idain:
> Okay, so what `DittoMetalPowder` is doing is doubling the defense in `c`, or if that overflows, halving the attack in `b` instead. 

It multiplies `c` by 1.5x and if it overflows, if halves _both the attack in `b` and the defense in `c`_, which maintains the ratio between them unaltered.

For example, let's assume `b` = 150 and `c` = 200. Multiplying `c` by 1.5 = 300 and their ratio is 0.5 (150/300). If we halve both registers the ratio doesn't change and the calculation is still acurate.

Rangi42:
Oh yeah, Metal Powder is 150%, not 200%, unlike Light Ball and Thick Club.

Idain:
I'm testing the new solution now. After I confirm that it works I'll post a new commit with the doc updated.

Idain:
Ok, I committed the new version and tested with both high and low values in `bc` including the multiplication. Please let me know if you see anything wrong, @Rangi42.

Rangi42:
Thanks for catching this bug!

PR #896, opened 2022-03-25
Update trainer constant labels

FrenchOrange:
*The constants for Camper Harvey, Schoolboy Kipp and Super Nerd Dave had their typos corrected.
*The constants for two unused trainers were missing. They have now been added.


aaaaaa123456789:
This is a non-matching PR: it introduces constants that don't exist, shifting the ones that do.

PR #897, opened 2022-03-26
Update misspelled trainer const

FrenchOrange:
Fix typos in the names of three trainer event constants.


constants/event_flags.asm
```diff
@@ -577,7 +577,7 @@
 	const EVENT_BEAT_CAMPER_LLOYD
 	const EVENT_BEAT_CAMPER_DEAN
 	const EVENT_BEAT_CAMPER_SID
-	const EVENT_BEAT_CAMPER_HERVEY
+	const EVENT_BEAT_CAMPER_HARVEY
```
Rangi42, line 580: Since this event is not used anywhere, it should be `const_skip`. Same for `EVENT_BEAT_SCHOOLBOY_KIPP`.

Rangi42:
Thanks for updating this! Just a couple more points:

- Please change the `STAN` trainer constant to `STAN_UNUSED`, like `ERIC_UNUSED` in pokegold.
- Make a corresponding PR in pokegold that adds these `const_skip`s and spelling corrections.

PR #900, opened 2022-03-31
Use NUM_BATTLE_STATS instead of raw numbers

Idain:
The good thing about messing with the code is finding instances where a constant should be used.


Idain:
@Rangi42 @mid-kid, feel free to merge this when you have time.

PR #902, opened 2022-04-05
Intro Movie Magic Numbers

vulcandth:
Noticed some magic numbers in the Intro Movie. This PR is to help give them better context.

`$90` appears to be the Screen Height in Pixels (144).

`$7` should be represented as a decimal in this context.

 `$d8`  (~line number 921) Is most likely `27 * TILE_WIDTH`.

### DIFF
--- a/engine/movie/intro.asm
+++ b/engine/movie/intro.asm
@@ -32,9 +32,9 @@ CrystalIntro:
 	xor a
 	ldh [hSCX], a
 	ldh [hSCY], a
-	ld a, $7
+	ld a, 7
 	ldh [hWX], a
-	ld a, $90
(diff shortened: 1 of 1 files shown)


mid-kid:
Numbers written like `$xx` are often magic numbers, yes, but I want to mention you should be careful about just plainly converting them to decimal, instead of making sure they get the correct representation.
In this case since it's a coordinate a decimal representation is fine but since it's a number used exclusively when the window is hidden (the Y coordinate is set to the screen's height, thus, disabled), we should consider making `WINDOW_HIDE_X` and `WINDOW_HIDE_Y` constants. However, since that's a repo-wide change, I'll leave it be for now.

PR #904, opened 2022-04-14
Fix misspelled labels

Idain:
`MovePKMNWithoutMail_InsertMon` was written as "witout". Funny, indeed.


vulcandth:
Looks good to me :+1:

Rangi42:
Thanks, just needs to be done for pokegold as well!

PR #908, opened 2022-04-19
Correct ClearChannels:: Comment

vulcandth:
I noticed a potentially inaccurate comment while digging through the code. The comment states that `ClearChannels` doesn't seem to to be used. This is not the case, as it is used by `ExchangeMysteryGiftData` and `ExchangeNameCardData`.

This will need to be fixed in `pokegold` as well.

### DIFF
--- a/audio/engine.asm
+++ b/audio/engine.asm
@@ -2800,7 +2800,6 @@ ChannelPointers:
 
 ClearChannels::
 ; runs ClearChannel for all 4 channels
-; doesn't seem to be used, but functionally identical to InitSound
 	ld hl, rNR50
 	xor a
 	ld [hli], a


aaaaaa123456789:
No point in mentioning which functions use it, I'd say. Just mention it's a duplicate of this and that so interested ROM hackers can replace it.

vulcandth:
> No point in mentioning which functions use it, I'd say. Just mention it's a duplicate of this and that so interested ROM hackers can replace it.

That is fair, however there is value in knowing that it is _only_ used by the Mystery Gift code... which many Rom Hacks look at removing that section of code to free space up. Although I agree it is probably not necessary to mention the functions. I'll push a new commit.

Rangi42:
Is it entirely functionally identical? `_InitSound`, in addition to everything `ClearChannels` does, also disables the music while it works, clears everything in `wAudio`, and sets the volume to maximum.

vulcandth:
> Is it entirely functionally identical? `_InitSound`, in addition to everything `ClearChannels` does, also disables the music while it works, clears everything in `wAudio`, and sets the volume to maximum.

Sorry; I honestly didn't investigate that part.. the statement about it being functionally identical was from the original comment.. So I left it in place. Most of the code from `ClearChannels`/`ClearChannel` is reused in `_InitSound` with some additional stuff. It may still be worth noting although not identically the same, most of it is the same code... and one could optimize it to reduce the redundant code.

How about if I worded it:
```asm
; Runs ClearChannel for all 4 channels and shares much of
; the same code in _InitSound. It is only used by the
; mystery gift code.
```

Rangi42:
Or just:
```diff
 ; runs ClearChannel for all 4 channels
-; doesn't seem to be used, but functionally identical to InitSound
```
The comment is from https://github.com/pret/pokecrystal/commit/9acf56102ae27aa418657b5b481cb5058ca6883c, way before Mystery Gift was figured out; if it weren't already here, we wouldn't be bothering to comment `ClearChannels` at all.

vulcandth:
Done; as suggested by Rangi. Please _**Squash**_ when merging this.

Rangi42:
@vulcandth This diff looks good, please open the same for pokegold and both can merge.

Rangi42:
@vulcandth This diff looks good, please open the same for pokegold and both can merge.

PR #925, opened 2022-05-25
Remove Heavy Ball never happens comment

xfix:
It's not correct as Snorlax weights more than 409.6 kg (also due to a bug, Kadabra, Tauros and Sunflora as well, but I digress).

### DIFF
--- a/docs/design_flaws.md
+++ b/docs/design_flaws.md
@@ -550,7 +550,7 @@ Edit [engine/items/item_effects.asm](https://github.com/pret/pokecrystal/blob/ma
  ; else add 0 to catch rate if weight < 204.8 kg
  ; else add 20 to catch rate if weight < 307.2 kg
  ; else add 30 to catch rate if weight < 409.6 kg
- ; else add 40 to catch rate (never happens)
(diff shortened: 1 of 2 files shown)


Rangi42:
Thank you! Can you make the same change in pokegold?

mid-kid:
Are you sure this is the case? The values in the pokedex are in lbs, and I've never really checked if the calculation applied here converts properly enough to trigger the case for snorlax.

vulcandth:
Even though the dex entries store the values in lbs vs kg's... I think snorlax *should* still trigger the 40 catch rate. As +40 would be 903.0 lbs (409.6 kg) or more. However, If this can wait until I get home tonight. I'll load up bgb and step through the execution while catching snorlax with a heavy ball to verify.

xfix:
The conversion formula seems to work well enough. For Snorlax I got high value of 0x11 (4600 // 256 = 0x11) which is a correct result. Stepping further into a code I also do properly get +40 bonus to catch rate.

image

mid-kid:
Thanks for checking!

PR #926, opened 2022-05-26
Fix channel 3 wave 9 id

dannye:
Tiny oversight from https://github.com/pret/pokecrystal/pull/650


Rangi42:
What if we took this opportunity to add constants for the wave pattern IDs? Like how we use `A_`, `B_`, `G_`, etc instead of numbers for note IDs.

PR #935, opened 2022-06-13
Warning fixes

coltongit:
No real reason to have the missing field initializers warning in place in the tools makefile because no such warnings exist.
Also fixes a -Wobsolete warning that happens with the latest RGBDS and pokecrystal where `nop` after `halt` won't be the default anymore soon, so to continue using this behavior a makefile tweak is required.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -103,7 +103,7 @@ tools:
 	$(MAKE) -C tools/
 
 
-RGBASMFLAGS = -L -Weverything -Wnumeric-string=2 -Wtruncation=1
+RGBASMFLAGS = -L -H -Weverything -Wnumeric-string=2 -Wtruncation=1
 # Create a sym/map for debug purposes if `make` run with `DEBUG=1`
 ifeq ($(DEBUG),1)
 RGBASMFLAGS += -E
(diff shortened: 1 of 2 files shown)


coltongit:
Drafted until the -Wobsolete fix has made it's way to the GitHub actions build

dannye:
> No real reason to have the missing field initializers argument in place in the tools makefile because no such warnings exist.

The purpose of the extra warning flags is to immediately notify us if any such warning is ever generated in the future. Just because no such warning exists in the code right now, that's no reason to remove the warning flag.

dannye:
No. ([source](https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html#Warning-Options))

dannye:
Oh wait, sorry, I wasn't paying enough attention and thought this was about `missing-field-initializers` rather than `no-missing-field-initializers`.

coltongit:
> Oh wait, sorry, I wasn't paying enough attention and thought this was about `missing-field-initializers` rather than `no-missing-field-initializers`.

The latter iirc would supress those warnings rather than request them

coltongit:
Split off into #962 and #963 and both supersede this

Issue #785, opened 2020-11-17
BUG : Ai_Smart_Encore can lead to incorrect type matchup checks


Arcaseriam:
File : `engine/battle/ai/scoring.asm`

The raw call to `CheckTypeMatchup` might have some oversights.
- Contrarily to the general comment of this routine, since `wEnemyMoveStruct` is
    populated with the last player move, this will be used properly in
    `CheckTypeMatchup` (putting it in `a` is indeed not how to use the matchup
    routine, but it is irrelevant in that case).
- However, if the AI flow can lead to this clause being called with a 
    `hBattleTurn` set to 0, the type matchup will consider the player types
    instead of the enemy types.

- AI layers which can cause this oversight if called before `AI_Smart_Encore`
  (without `hBattleTurn` being set back to 1 unvoluntarily) :
  - `AI_Smart_Conversion2`
  - TODO : there could be some other very tricky situations like the last turn
    using Metronome->Conversion2 if the AI switching conserves `hBattleTurn`, etc.

- AI layers which unvoluntarily avoid this oversight (by setting `hBattleTurn` to 1) :
  - `AI_Types` (SuperNerd, Juggler, ExecutiveM/F, PokefanM/F).
  - `AI_Status` reaching the `.typeimmunity` jump label.
  - `AI_Smart_LeechHit`.
  - `AI_Smart_LockOn` when the player isn't locked on.
  - `AI_Smart_Mimic` when player used a move & enemy HP > 50%.
  - `AI_Smart_PriorityHit` when the enemy is slower or speedtied 
       and the player is not using Fly or Dig.
  - `AI_Smart_HiddenPower`.

Example (not encoutered in vanilla) :

- Player : Ho-Oh with `Mud-Slap`.
- Enemy : `SMART_AI` Raichu with `Conversion2, Encore` in that order (any other moves before or after).
- Turn 1 : Raichu uses another move. Ho-Oh uses Mud-Slap.
- Turn 2 : Raichu will wrongly consider types and will use Ho-Oh's after the AI_Smart_Conversion2 layer. It will see Ho-Oh as immune to Mud-Slap and consider this matchup its. Thus `Encore` is 72% greatly encouraged (-2 score) despite Mud-Slap being super-effective against Raichu.

FredrIQ:
I'm fairly sure that the AI in general assumes that SetEnemyTurn has been prepared, meaning that the problem here is Conversion2, not Encore?

coltongit:
[...]

PR #784, opened 2020-11-11
Documenting AI scoring

Arcaseriam:
Goals : 
1) comment yet uncommented AI (sub-)layers.
2) double-check already commented AI (sub-)layers.
3) add ASCII-compatible diagrams.
4) identify more code misconceptions/bugs.

PR #936, opened 2022-06-13
Document the Encore type matchup bug

coltongit:
Document the type matchup bug that was discussed in #785 and touched upon in #784.
Actual comment used for documentation was in https://github.com/pret/pokecrystal/pull/784/files#diff-8cf500d6ca0f463d796e749bdae9b8ff1d6e3586224e764cdde0109db91dce99R1772=.

This may be the result of Conversion2 itself, but would be appreciated if this were to be looked into further.


coltongit:
Drafted until the root cause, whatever it is, is identified and the fix implemented into this PR

vulcandth:
Unless you plan on coming up with the solution.. their is little point in opening a PR for this as the Issue already exists. Other people can't work on this PR unless they are a maintainer... The PR should be opened by whoever comes up with the solution for the bug.

coltongit:
> Unless you plan on coming up with the solution.. their is little point in opening a PR for this as the Issue already exists. Other people can't work on this PR unless they are a maintainer... The PR should be opened by whoever comes up with the solution for the bug.
> 
> Not to discourage you from opening PR's.. but for example I can't contribute to this PR.. as i'm not a maintainer. So If I found a solution.. i'd have to contact you and wait for you to update your PR.

The reason for this PR is so that it has a better chance of it's true root cause being identified and a fix figured out for it. If you want to help with it, you could always suggest changes to the document itself with the new fix in question.

Rangi42:
> The reason for this PR is so that it has a better chance of it's true root cause being identified and a fix figured out for it.

That's what issues are for.

PR #937, opened 2022-06-13
Document that `farcall` returns `c` in `a` for some code

### LINKED ISSUE #915: Comment that some code depends on `farcall` returning the value of `c` in `a`
See https://discord.com/channels/442462691542695948/487424856913346580/971165772632186980:

> `TryTileCollisionEvent` loads `wFacingTileID` into `c`, then `farcall`s `CheckFacingTileForStdScript`, which preserves `c`, but since it's a `farcall`, implicitly copies `c` back into `a` on return.  Then it calls `CheckCutTreeTile` which checks this value.  Notice how every other <code>Check<i>&lt;x&gt;</i>Tile</code> call in the same function explicitly loads `wFacingTileID` again.

coltongit:
Fixes #915
Other code may also do the same thing, would be worth a fine-toothed comb through to check again

### DIFF
--- a/engine/overworld/events.asm
+++ b/engine/overworld/events.asm
@@ -1085,10 +1085,13 @@ TryTileCollisionEvent::
 	call GetFacingTileCoord
 	ld [wFacingTileID], a
 	ld c, a
+	; CheckFacingTileForStdScript preserves c, and
+	; farcall copies c back into a.
 	farcall CheckFacingTileForStdScript
 	jr c, .done
 
-	; CheckCutTreeTile expects c to be copied into a by farcall
+	; CheckCutTreeTile expects a == [wFacingTileID], which
+	; it still is after the previous farcall.
 	call CheckCutTreeTile
 	jr nz, .whirlpool
 	farcall TryCutOW
--- a/engine/overworld/events.asm
+++ b/engine/overworld/events.asm
@@ -1087,6 +1087,7 @@ TryTileCollisionEvent::
 	ld c, a
 	farcall CheckFacingTileForStdScript
 	jr c, .done
+
 	; CheckCutTreeTile expects c to be copied into a by farcall
 	call CheckCutTreeTile
 	jr nz, .whirlpool
--- a/engine/overworld/events.asm
+++ b/engine/overworld/events.asm
@@ -1081,6 +1081,13 @@ LoadScriptBDE::
 	scf
 	ret
 
+; TryTileCollisionEvent loads wFacingTileID into c,
+; then farcalls CheckFacingTileForStdScript,
+; which preserves c, but since it's a farcall,
+; implicitly copies c back into a on return.
+; Then it calls CheckCutTreeTile which checks this value.
+; Every other Check<x>Tile call in the same function
+; explicitly loads wFacingTileID again.
 TryTileCollisionEvent::
 	call GetFacingTileCoord
 	ld [wFacingTileID], a
--- a/engine/overworld/events.asm
+++ b/engine/overworld/events.asm
@@ -1087,7 +1087,6 @@ TryTileCollisionEvent::
 	ld c, a
 	farcall CheckFacingTileForStdScript
 	jr c, .done
-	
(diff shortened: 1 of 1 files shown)


mid-kid:
That's a huge ass comment for something that could be commented as "expects c to be copied into a".

engine/overworld/events.asm
```diff
@@ -1081,6 +1081,7 @@ LoadScriptBDE::
 	scf
 	ret
 
+; expects c to be copied into a
```
vulcandth, line 1084: Please move this comment above `call CheckCutTreeTile` like this:

```diff
-; expects c to be copied into a
 TryTileCollisionEvent::
 	call GetFacingTileCoord
 	ld [wFacingTileID], a
 	ld c, a
 	farcall CheckFacingTileForStdScript
 	jr c, .done

+	; CheckCutTreeTile expects c to be copied into a by farcall
 
 	call CheckCutTreeTile
 	jr nz, .whirlpool
 	farcall TryCutOW
 	jr .done
```

engine/overworld/events.asm
```diff
@@ -1087,7 +1087,8 @@ TryTileCollisionEvent::
 	ld c, a
 	farcall CheckFacingTileForStdScript
 	jr c, .done
-
+	
+	; CheckCutTreeTile expects c to be copied into a by farcall
 	call CheckCutTreeTile
```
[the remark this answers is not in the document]
Rangi42: That might be a correction to make to STYLE.md instead. In practice we do generally keep comments "attached" to the line or "paragraph" they're explaining:

```
	; This call is unusual somehow.
	call SomeFunction
```

```
	; Call the function at de
	ld a, [de]
	ld l, a
	inc de
	ld a, [de]
	ld h, a
	ld de, .return
	push de
	jp hl

.return
```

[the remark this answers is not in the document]
vulcandth: I support making the change in STYLE.md then. As I actually prefer keeping them attached personally.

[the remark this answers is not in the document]
coltongit: Yeah, this is something for STYLE.md

vulcandth (approved):
Alright. Looks good to me! Thanks :)

engine/overworld/events.asm
```diff
@@ -1087,7 +1087,8 @@ TryTileCollisionEvent::
 	ld c, a
 	farcall CheckFacingTileForStdScript
 	jr c, .done
-
+
```
Rangi42, line 1090:
```suggestion

```

engine/overworld/events.asm
```diff
@@ -1087,7 +1087,7 @@ TryTileCollisionEvent::
 	ld c, a
 	farcall CheckFacingTileForStdScript
 	jr c, .done
-
+	; CheckCutTreeTile expects c to be copied into a by farcall
```
Rangi42, line 1090: I just meant, remove the trailing space.
```suggestion

	; CheckCutTreeTile expects c to be copied into a by farcall
```

engine/overworld/events.asm
```diff
@@ -1088,6 +1088,7 @@ TryTileCollisionEvent::
 	farcall CheckFacingTileForStdScript
```
Rangi42, line 1088:
```suggestion
	; CheckFacingTileForStdScript preserves c, and
	; farcall copies c back into a.
	farcall CheckFacingTileForStdScript
```

engine/overworld/events.asm
```diff
@@ -1088,6 +1088,7 @@ TryTileCollisionEvent::
...
 	jr c, .done
 
+	; CheckCutTreeTile expects c to be copied into a by farcall
```
Rangi42, line 1091:
```suggestion
	; CheckCutTreeTile expects a == [wFacingTileID], which
	; it still is after the previous farcall.
```

Rangi42 (approved):
This will also need doing for pokegold.

vulcandth:
Created the pokegold PR.

PR #942, opened 2022-06-14
Adding the legacy audio script to stand beside the one in modern pokecrystal

### LINKED ISSUE #759: Provide macros/script/audio.asm for integration with old projects
PR https://github.com/pret/pokecrystal/pull/650 overhauled the way music macros work for new projects, while providing backwards-compatibility with older macros.

However, music writers have expressed woes about music written for new repos not being backwards-compatible with older repos, which are in use by a big number of active projects.

To alleviate this issue, it'd be nice to make a drop-in replacement for `macros/script/audio.asm` that can be used with the old repositories, to add the new macros while keeping compatibility with the old ones. This would be provided on the wiki, provide installation instructions, and be possible to be linked along with songs that use the new macro format.

This would involve taking the new file, adding the legacy compatibility macros, making either a new, local const macro or using `enum` again, and making sure a repo version right before this PR still matches with this, and a version from december 2017 if possible, taking into account the RGBDS version used at the time.

Bonus points for this file still being compatible with current repos.

...this, or adding instructions explaining how to build an audio file as its own object while `INCLUDE`ing a different set or macros.

### LINKED ISSUE #873: Add legacy music command names to the docs
These are used in many custom songs, so it would be helpful to see the correspondence between legacy and modern music commands.

coltongit:
`legacy_audio.asm` allows older files from before PR #650 to work with the latest pokecrystal (for now until compat macros are implemented).
This PR also restores pre #650 music commands in `legacy_music_commands.md`.
This draft PR fixes #759 and fixes #873


coltongit:
@Rangi42 Any thoughts on this? This approach could work as a stopgap until legacy.asm gets it's compatibility macros mentioned in #759.

PR #944, opened 2022-06-15
Add ReadNoiseSample design flaw fix

### LINKED ISSUE #924: Incrementation Distorts Percussion in ReadNoiseSample
This issue ranges from SW97 all the way to Crystal.  In `ReadNoiseSample`, the first byte is split into a purely cosmetic hi nybble to make importing RBY noise data a matter of copy-paste, and a lo nybble for the length.

All the RBY percussion is already there, and drums that are two or more notes long indicate the same note length structure as in RBY.  However, because of the way `ReadNoiseSample` stores this first byte, this length is one frame longer than intended, meaning minimum note length is two instead of one.  This makes the old drums with two or more notes sound slower than they should be.

Would this be counted as a bug?

```	ld a, [de]
	inc de

	cp sound_ret_cmd
	jr z, .quit

	and $f
	inc a
	ld [wNoiseSampleDelay], a```

coltongit:
Wiki page for this at https://github.com/pret/pokecrystal/wiki/Fix-ReadNoiseSample-instrumentation-distorting-percussion
Fixes #924

### DIFF
--- a/docs/design_flaws.md
+++ b/docs/design_flaws.md
@@ -12,6 +12,7 @@ These are parts of the code that do not work *incorrectly*, like [bugs and glitc
 - [`ITEM_C3` and `ITEM_DC` break up the continuous sequence of TM items](#item_c3-and-item_dc-break-up-the-continuous-sequence-of-tm-items)
 - [Pokédex entry banks are derived from their species IDs](#pokédex-entry-banks-are-derived-from-their-species-ids)
 - [Identical sine wave code and data is repeated five times](#identical-sine-wave-code-and-data-is-repeated-five-times)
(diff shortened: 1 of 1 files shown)


Rangi42:
The point of a wiki article is that [@TempoQuill](https://github.com/TempoQuill) or whoever else can freely describe the Gen 1+2 audio engine differences and resolutions with as much detail and personal input as they want. As I recall, the discussion in #924 and on Discord was confused/controversial enough that this one particular "flaw"/fix shouldn't go in these "official" docs. (I was hoping TempoQuill would more thoroughly write up their observations on the wiki...)

PR #943, opened 2022-06-14
pokecrystal using `hardware.inc`

coltongit:
~~This draft PR shows what pokecrystal would look like if it was using hardware.inc instead of it's present constants file.~~
This PR now aims to convert pokecrystal to using hardware.inc above what it uses now for constants.
Resolves #914

PR #947, opened 2022-06-17
Remove rNR20/rNR40 and add rPCM12/rPCM34

### LINKED ISSUE #946: Remove non-existent rNR20 and rNR40 hardware register constants
As described in the [Pan Docs](https://gbdev.io/pandocs/Sound_Controller.html), channels 2 and 4 do not have a Tone Envelope/Sweep Register. However, pokecrystal and pokegold define `rNR20` and `rNR40` constants, not because those registers are directly used, but just as offsets to pass to `ClearChannel`, which expects `hl` to point to some rNRx0-1-2-3-4 series of registers. We should use `rNR21 - 1` and `rNR41 - 1` with appropriate comments (even just `; there is no rNR20`).

coltongit:
Fixes #946
pokegold companion PR at https://github.com/pret/pokegold/pull/92


coltongit:
Two previously unknown constants have been identified thanks to `hardware.inc` and #943

Idain:
It'd be also good to restore the comment specifying those two new constants are read-only so people don't make mistakes.

Idain:
> > It'd be also good to restore the comment specifying those two formerly unknown constants are read-only so people don't make mistakes.
> 
> That's what the `(R)` is for, if you read the rest of the file and the other audio-related constants you'll see that

Now that we're at it, it's not very obvious `(R)` means "Read-only".

coltongit:
> > > It'd be also good to restore the comment specifying those two formerly unknown constants are read-only so people don't make mistakes.
> > 
> > 
> > That's what the `(R)` is for, if you read the rest of the file and the other audio-related constants you'll see that
> 
> Now that we're at it, it's not very obvious `(R)` means "Read-only".

The other audio related constants use `(R)` and `(R/W)` as well. Should they also be tweaked if we're gonna tweak this one?

constants/hardware_constants.asm
```diff
@@ -163,6 +161,6 @@ DEF rUNKNOWN2            EQU $ff72 ; (00h) - Bit 0-7 (Read/Write)
 DEF rUNKNOWN3            EQU $ff73 ; (00h) - Bit 0-7 (Read/Write)
 DEF rUNKNOWN4            EQU $ff74 ; (00h) - Bit 0-7 (Read/Write) - CGB Mode Only
 DEF rUNKNOWN5            EQU $ff75 ; (8Fh) - Bit 4-6 (Read/Write)
-DEF rUNKNOWN6            EQU $ff76 ; (00h) - Always 00h (Read Only)
-DEF rUNKNOWN7            EQU $ff77 ; (00h) - Always 00h (Read Only)
+DEF rPCM12               EQU $ff76 ; Channel 1 & 2 Amplitude (R)
```
ISSOtm, line 164: Why have those if they are not used anywhere?

ISSOtm (changes requested):

PR #949, opened 2022-06-18
Remove unused serial constants ported from #943

coltongit:
Backported the removal of two unused constants from PR #943 to here
As @ISSOtm pointed out, these two are unused in the repo


Rangi42 (approved):
Thanks! Please do the same for pokegold.

coltongit:
pokegold companion PR at https://github.com/pret/pokegold/pull/95

PR #950, opened 2022-06-18
Start work on modern audio compatibility with legacy pokecrystal

### LINKED ISSUE #759: Provide macros/script/audio.asm for integration with old projects

### LINKED ISSUE #873: Add legacy music command names to the docs

coltongit:
Fixes #759
Supersedes and resolves #942
Closes #873


macros/modern.asm
```diff
@@ -0,0 +1,12 @@
+; Modern support for legacy pokecrystal and fanhacks that use older codebase
+; Allows porting scripts with as few edits as possible.
+
+; macros/rst.asm
+DEF farcall EQUS "callba"
+DEF callfar EQUS "callab"
```
Rangi42, line 6: #759 specifically requested modern *music* script support for older projects. Throwing in extra undiscussed stuff isn't an improvement, it's a breakage: in this case, `farcall` has always been defined, and this would be a compile error.

Rangi42: Also if older projects use older rgbds versions, they may not support `DEF` syntax. (I assume that's why you did `transpose: MACRO` and not `MACRO transpose` below, too.)

coltongit: older rgbds variants of this may require different syntax, yes. It might be best to give that a shot after this PR's ready to merge

coltongit:
Would this be OK to merge as-is or would something that allows older rgbds to recognize `DEF` be recommended beforehand?

coltongit:
@Rangi42 Are there any other ideas for some of the macros that could be added before merge? I'm not sure where to begin with the rest

coltongit:
Once the wiki page is finalized from what was mentioned in #759, this PR can be closed and so can #759 and #873

coltongit:
Superseded by https://github.com/pret/pokecrystal/wiki/Add-audio-scripts-to-use-modern-audio-files-with-legacy-pokecrystal, edit as you wish

PR #959, opened 2022-06-20
add pycache to gitignore

coltongit:
Port pycache gitignore addition from mid-kid's i18n work (based on https://github.com/mid-kid/pokecrystal/commit/b6914f3c06d1459b324f39c2e8e5bf6c8f52aaa5)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,6 +27,7 @@ gfx/pokemon/*/frames.asm
 *.exe
 
 # precompiled python
+__pycache__
 *.pyc
 *$py.class
 


vulcandth:
Why are we adding this?

coltongit:
> Why are we adding this?

In case pycache files are somehow dropped into the source directory, wouldn't want to clutter the repo with them needlessly

Idain:
But we haven't had any problems with them. If someone wants to avoid it for their personal project, that's their responsibility.

coltongit:
I'll close this for now, if it does become a problem in the future I'll rebase and we'll have another look

mid-kid:
`*.pyc` should be enough to ignore the `__pycache__` dir anyway.

mid-kid:
I think it's good to have ignores for common files that are generated by the tools we have in this repo, or are frequently used in combination with this repo. Yes that includes pycache and mac files, though vim is arguable.

mid-kid:
> Didn't realise there were python files in the repo (for compiling the images I guess?)

They're auxiliary tools not involved in the build process but useful for a few things.

> If we're ignoring mac files we should be ignoring windows files too, plus whatever android and every 9000 file managers on linux decide to make. Using a mac is a choice, and not one everyone makes.

Mac is the only platform that generates files that git can pick up like this. Thumbs.db on windows is hidden and doesn't get picked up, and the 9000 file managers on linux don't create files like this either.

I very much prefer avoiding people shooting themselves in the foot with this. Our audience is people who might be unaware of the existence of this metadata file, let alone know about advanced git configuration.

mid-kid:
> Didn't realise there were python files in the repo (for compiling the images I guess?)

They're auxiliary tools not involved in the build process but useful for a few things.

> If we're ignoring mac files we should be ignoring windows files too, plus whatever android and every 9000 file managers on linux decide to make. Using a mac is a choice, and not one everyone makes.

Mac is the only platform that generates files that git can pick up like this. Thumbs.db on windows is hidden and doesn't get picked up, and the 9000 file managers on linux don't create files like this either.

I very much prefer avoiding people shooting themselves in the foot with this. Our audience is people who might be unaware of the existence of this metadata file, let alone know about advanced git configuration.

PR #962, opened 2022-06-22
Fix warnings in lzcomp

coltongit:
First part of #935 addressed


mid-kid:
We've decided way long ago to ignore the warnings through CFLAGS rather than modifying the program, to make it easier to sync with aaaaaa123456's upstream, since he refuses to write with these sort of warnings in mind.

PR #963, opened 2022-06-22
Fix -Wobsolete warning when building in latest rgbds master

coltongit:
Second part of #935 addressed

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -103,7 +103,7 @@ tools:
 	$(MAKE) -C tools/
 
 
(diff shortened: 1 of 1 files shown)


coltongit:
Draft until this is able to build in GitHub actions

mid-kid:
The `-H` option is not recognized in older versions of rgbds, so we need to bump rgbdscheck.asm and the version numbers in INSTALL.md, once the version this issue occurs in is released. That said, I'm not entirely sure this is worth bumping the rgbds version requirement for...

Rangi42:
The instructions say to use 0.5.2, not master.

PR #969, opened 2022-06-24
Remove all redundant comments in WRAM, needs review

### LINKED ISSUE #741: Remove redundant documentation comments
Some values in WRAM have comments that reiterate what's already described with their associated constants or data.

In particular, the `wPlayerSubStatus*` labels have long comments to remove. (The `SUBSTATUS_*` constants should also be renamed `SUBSTATUS1_*`, `SUBSTATUS2_*`, etc, to make it clear which of the five substatus bytes they correspond to.)

coltongit:
Removes all the redundant comments relating to bitfields, requires extra review

There may be some that still require the comments so we know what they are, PLEASE let me know here so I know which ones to keep and which ones to not

Fixes #741

### DIFF
--- a/ram/wram.asm
+++ b/ram/wram.asm
@@ -29,23 +29,12 @@ wCurMusicByte:: db
 wCurChannel:: db
 wVolume::
 ; corresponds to rNR50
-; Channel control / ON-OFF / Volume (R/W)
-;   bit 7 - Vin->SO2 ON/OFF
-;   bit 6-4 - SO2 output level (volume) (# 0-7)
-;   bit 3 - Vin->SO1 ON/OFF
-;   bit 2-0 - SO1 output level (volume) (# 0-7)
 	db
 wSoundOutput::
 ; corresponds to rNR51
-; bit 4-7: ch1-4 so2 on/off
-; bit 0-3: ch1-4 so1 on/off
 	db
 wPitchSweep::
 ; corresponds to rNR10
-; bit 7:   unused
-; bit 4-6: sweep time
-; bit 3:   sweep direction
-; but 0-2: sweep shift
 	db
 
 wMusicID:: dw
@@ -57,15 +46,10 @@ wMusicNoiseSampleSet:: db
 wSFXNoiseSampleSet:: db
 
 wLowHealthAlarm::
-; bit 7: on/off
-; bit 4: pitch
-; bit 0-3: counter
 	db
 
 wMusicFade::
 ; fades volume over x frames
-; bit 7: fade in/out
-; bit 0-5: number of frames for each volume level
 ; $00 = none (default)
 	db
 wMusicFadeCount:: db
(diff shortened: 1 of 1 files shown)


vulcandth (commented):
I believe you are still missing many comments that need to be removed such as `wBattleWeather` and `wPlayerScreens` just to name two.

Also; I'm not sure what Rangi/Mid-Kid might want... but i'd think we'd want to verify that the comment has a related constant/data before deleting it. If it does not, then we should consider either creating those constants (if they can be used) or keeping the comments.

For example, `wSoundOutput` does not have any defined constants... but could potentially use them. (_**I probably defined too many**_)  
in `constants/audio_constants.asm`

```asm
; wSoundOutput
DEF SOUND_OUTPUT_S01_CH1_F EQU 0
DEF SOUND_OUTPUT_S01_CH2_F EQU 1
DEF SOUND_OUTPUT_S01_CH3_F EQU 2
DEF SOUND_OUTPUT_S01_CH4_F EQU 3
DEF SOUND_OUTPUT_S02_CH1_F EQU 0
DEF SOUND_OUTPUT_S02_CH2_F EQU 1
DEF SOUND_OUTPUT_S02_CH3_F EQU 2
DEF SOUND_OUTPUT_S02_CH4_F EQU 3
DEF SOUND_OUTPUT_S01_S02_CH1_ON EQU (1 << SOUND_OUTPUT_S01_CH1_F) | (1 << SOUND_OUTPUT_S02_CH1_F)
```

And in `PlayDanger:` of `audio/engine.asm`

```diff
 .noreset
     ; Make sure the danger sound is kept on
     or 1 << DANGER_ON_F
     ld [wLowHealthAlarm], a

     ; Enable channel 1 if it's off
     ld a, [wSoundOutput]
-    and $11
+    and SOUND_OUTPUT_S01_S02_CH1_ON 
     ret nz
     ld a, [wSoundOutput]
-    or $11
+    or SOUND_OUTPUT_S01_S02_CH1_ON 
     ld [wSoundOutput], a
     ret
```

Please don't act on my comments for now; Let Rangi/Mid-kid provide their feedback first.

mid-kid:
> Also; I'm not sure what Rangi/Mid-Kid might want... but i'd think we'd want to verify that the comment has a related constant/data before deleting it. If it does not, then we should consider either creating those constants (if they can be used) or keeping the comments.

Yeah, this was the intention, not willy-nilly deleting them all. Of course it implies actually using the constants in the code, instead of arbitrary bit values, like you've exemplified.

coltongit:
I started with this as a base, and am gonna be adding back whichever ones have constants associated with them, just wasn't sure which ones because the relevant issue wasn't exactly the clearest about it

coltongit:
Drafting so I get a chance to figure out what to do for this one

Rangi42:
> I started with this as a base, and am gonna be adding back whichever ones have constants associated with them, just wasn't sure which ones because the relevant issue wasn't exactly the clearest about it

If an issue seems broad or unclear, please just don't open a PR. It's totally fine if we have open issues for unresolved things sitting around until someone knows what to do with them. You also don't have to handle every aspect of broad issues, but open PRs for whichever pieces you do have solutions for already.

PR #970, opened 2022-06-26
wVirtualOAM->wShadowOAM as per convention

coltongit:
As noticed by @ISSOtm in #943


vulcandth (changes requested):
You changed the file permissions for several of the executable files. Please revert the permission changes.

For Example:
`.github/webhook.sh
 100755 → 100644`
`tools/compare.sh 100755 → 100644`
... ect.

coltongit:
> You changed the file permissions for several of the executable files. Please revert the permission changes.
> 
> For Example: `.github/webhook.sh 100755 → 100644` `tools/compare.sh 100755 → 100644` ... ect.

I'm not sure why that happened, with git it shouldn't have

PR #972, opened 2022-06-27
Add object priority mode constant and remove unnecessary constants

coltongit:
One of the unknown register constants, labelled as `rUNKNOWN1` in `hardware_constants.asm`, was looked up in the Game Boy Pan Docs and the actual register it matches to is the Object Priority Mode register `rOPRI`.
The rest of these unknown constants linked to registers that aren't even proper 'registers' in the Game Boy itself.
This PR aims to add this constant definition and remove the other unknown constant definitions in `hardware_constants.asm`.


Rangi42:
Sure, why not; please do the same for gold+red+yellow.

coltongit:
> Sure, why not; please do the same for gold+red+yellow.

Done.

PR #976, opened 2022-07-21
urlencode special characters like 'é' in documentation link anchors

coltongit:
The [website version of the repo documentation](https://pret.github.io/pokecrystal) has had several broken links for a long while now, namely links that use the `é` character as-is. Using the link (I presume HTML, please correct me if I'm wrong) equivalent `%C3%A9` in place of them fixes all the links in the website version of the repo documentation.


Rangi42:
They also work in the GitHub `/blob/` URL view, so that's good. :+1: 

Can you update tools/toc.py too?

```diff
 import sys
 import re
 from collections import namedtuple
+from urllib.parse import quote
```

```diff
 def name_to_anchor(name):
 	# GitHub's algorithm for generating anchors from headings
 	# https://github.com/jch/html-pipeline/blob/master/lib/html/pipeline/toc_filter.rb
 	anchor = name.strip().lower()                   # lowercase
 	anchor = re.sub(punctuation_regexp, '', anchor) # remove punctuation
 	anchor = anchor.replace(' ', '-')               # replace spaces with dash
 	anchor = re.sub(specialchar_regexp, '', anchor) # remove misc special chars
+	anchor = quote(anchor)                          # url encode
 	return anchor
```

Rangi42 (approved):

vulcandth:
Something else I noticed as a consequence of this PR. All of the entries you identified fail to work properly on the Github Mobile App. This occurs in both `pret:master` and your `docs-website-fix` branch. This is most likely a problem with the Mobile App itself... however I should ask if perhaps we should just avoid using the `é` in Titles/Links. Thoughts @Rangi42?

coltongit:
> Something else I noticed as a consequence of this PR. All of the entries you identified fail to work properly on the Github Mobile App. This occurs in both `pret:master` and your `docs-website-fix` branch. This is most likely a problem with the Mobile App itself... (Clicking on a link in the table of contents sends you to the top.. instead of the bug/design section) however I should ask if perhaps we should just avoid using the `é` in Titles/Links. Thoughts @Rangi42?
> 
> Other than that.. this lgtm.

Rather than that, it'd be best to urlencode all instances of special characters like é through `toc.py` before they are accepted imo.
I'll look into the GitHub mobile app problem and see if there's anything that can be done on our end to fix it. If not, it will have to be taken up with GitHub upstream.

vulcandth (approved):

vulcandth:
I was looking at some alternatives using html to get it to work.. but the mobile app is just dumb. I don't believe it is worth the effort to jump through hoops to avoid the minor inconvenience in the Mobile app's "bug". If you think of something later; then just let me know :+1: Thanks!

Rangi42:
> Something else I noticed as a consequence of this PR. All of the entries you identified fail to work properly on the Github Mobile App. This occurs in both `pret:master` and your `docs-website-fix` branch. This is most likely a problem with the Mobile App itself... (Clicking on a link in the table of contents sends you to the top.. instead of the bug/design section) however I should ask if perhaps we should just avoid using the `é` in Titles/Links. Thoughts @Rangi42?

Do the plain-text links work on mobile, though? Because I've noticed many anchors not working in the mobile app, like the line-number ones (`#L42` or `#L100-L120`).

PR #979, opened 2022-07-28
Define scene constants alongside scene_scripts

### LINKED ISSUE #978: Define scene constants alongside their `scene_script` pointers
Right now we have constants/scene_constants.asm, which defines constants for each map that have to correlate to the map's scene IDs.

https://github.com/pret/pokered/pull/367/files is introducing a `dw_const` macro to pokered that will simultaneously declare a pointer and define a corresponding constant. It's particularly useful in Gen 1 because there were tables of text and script pointers, and bg and object events referred directly to them.

pokecrystal could do the same:

```diff
 AzaleaTown_MapScripts:
 	def_scene_scripts
-	scene_script .DummyScene0 ; SCENE_AZALEATOWN_NOTHING
-	scene_script .DummyScene1 ; SCENE_AZALEATOWN_RIVAL_BATTLE
-	scene_script .DummyScene2 ; SCENE_AZALEATOWN_KURT_RETURNS_GS_BALL
+	scene_script .DummyScene0, SCENE_AZALEATOWN_NOTHING
+	scene_script .DummyScene1, SCENE_AZALEATOWN_RIVAL_BATTLE
+	scene_script .DummyScene2, SCENE_AZALEATOWN_KURT_RETURNS_GS_BALL
```

The advantages are not having to edit two files, related data being kept together, and symmetry with pokered.

The disadvantage is that many `scene_script`s are just dummies, because the scene ID is meant for a `coord_event` instead. Furthermore, although Game Freak defined a `scene_script` for every valid scene ID, you don't actually have to. None of those Azalea Town `scene_script`s need to exist because they're all for `coord_event`s. So editing the macro this way would either force people to define dummy `scene_script`s, or we'd need to complicate the macro to allow defining only a constant.

Alternatively, we could do the same thing as `object_const_def`, and define a sequence of scene IDs in each map file. This would hopefully not be confusing when a different map uses the constant, as in `setmapscene AZALEA_TOWN, SCENE_AZALEATOWN_RIVAL_BATTLE`.

Rangi42:
Fixes #978.


maps/BattleTower1F.asm
```diff
@@ -7,8 +7,8 @@
 
 BattleTower1F_MapScripts:
 	def_scene_scripts
-	scene_script .Scene0 ; SCENE_DEFAULT
-	scene_script .Scene1 ; SCENE_FINISHED
+	scene_script .Scene0, SCENE_BATTLETOWER1F_CHECKSTATE
+	scene_script .Scene1, SCENE_BATTLETOWER1F_NOTHING
```
vulcandth, line 11: I personally prefer `SCENE_MAPNAME_DEFAULT` for the initially set default scenes.. as this would match up better with what i'm doing in pokered's PR. Also my PR uses `SCRIPT_MAPNAME_NOOP` for the dummy scripts; perhaps something similar for the nothing scenes?

Just thinking about consistency.

maps/IndigoPlateauPokecenter1F.asm
```diff
@@ -17,12 +17,12 @@ IndigoPlateauPokecenter1F_MapScripts:
...
-	setmapscene BRUNOS_ROOM, SCENE_DEFAULT
-	setmapscene KARENS_ROOM, SCENE_DEFAULT
-	setmapscene LANCES_ROOM, SCENE_DEFAULT
-	setmapscene HALL_OF_FAME, SCENE_DEFAULT
+	setmapscene WILLS_ROOM, SCENE_WILLSROOM_LOCK_DOOR
+	setmapscene KOGAS_ROOM, SCENE_KOGASROOM_LOCK_DOOR
+	setmapscene BRUNOS_ROOM, SCENE_BRUNOSROOM_LOCK_DOOR
+	setmapscene KARENS_ROOM, SCENE_KARENSROOM_LOCK_DOOR
+	setmapscene LANCES_ROOM, SCENE_LANCESROOM_LOCK_DOOR
+	setmapscene HALL_OF_FAME, SCENE_HALLOFFAME_ENTER
```
vulcandth, line 25: Possibly align the constants here.

maps/BattleTower1F.asm
Rangi42: I chose not to use `SCENE_*_DEFAULT` because it doesn't say what the default scene is for. Sometimes it's for `NOTHING`, sometimes for a one-time event like meeting an NPC that gets permanently switched to a `NOTHING` scene, sometimes it might switch back to the default scene later.

pokecrystal has used `SCENE_NOTHING` for do-nothing scenes for a while; maybe pokered should use `NOTHING` instead of `NOOP`? It's longer, but also an actual word. Thoughts @dannye @mid-kid ?

maps/IndigoPlateauPokecenter1F.asm
Rangi42: Eh, we don't align things like this either:
```
	ifequal SUNDAY, PlateauRivalScriptDone
	ifequal TUESDAY, PlateauRivalScriptDone
	ifequal THURSDAY, PlateauRivalScriptDone
	ifequal FRIDAY, PlateauRivalScriptDone
	ifequal SATURDAY, PlateauRivalScriptDone
```

maps/BattleTower1F.asm
vulcandth: I think that not using `SCENE_*_DEFAULT` is fine then.. I'd still prefer consistency for whatever we decide to go with for the Do nothing scenes/scripts. 

Also another note... Should we rename the scripts as well? Like:
```asm
	scene_script .CheckStateScene, SCENE_BATTLETOWER1F_CHECKSTATE
	scene_script .NothingScene,    SCENE_BATTLETOWER1F_NOTHING
```

dannye: I don't know enough about gen 2 to really know whether `SCENE_*_DEFAULT` is or is not appropriate.

At least in gen 1, a lot of maps' "default" script handles multiple things and executes every frame indefinitely so "default" is a perfectly adequate name imo (doesn't mean it *always* has to be named like this, if a better name exists on a case-by-case basis). It's also nice to know - at a glace - that the "default" script is the one that will execute on each new save until the map's script is explicitly updated.
If some part of this (with regard to how the scripting engines work) doesn't apply to gen 2, then I think it's okay if pokered and pokecrystal are a bit inconsistent.

I like `NOOP` more than `NOTHING`.

> It's longer, but also an actual word.

It may not be a dictionary word, but it's not like it's just made up either. Not really a relevant point in my opinion.

mid-kid: I don't mind having a consistent `<MAPNAME>_NOOP`, DEFAULT, or NOTHING or similar, but the value of the "nothing" scene isn't consistent across every map. Like rangi mentioned, scene 0 is often a one-time event.

Rangi42: I'll change `NOTHING` to `NOOP` for consistency then.

mid-kid (approved):

vulcandth (approved):
If you don't want to change the Scene pointer names as well in this PR then this looks good! I'll open an issue for it later, and I can take care of it as some point.

PR #980, opened 2022-07-28
`is_any(v, A, B, ...)` instead of `v == A || v == B || ...`

Rangi42:
These should optimize just as well as before; see <https://godbolt.org/z/sf7fqMMf3>.


Rangi42:
From review: too much abstraction for too little benefit.

Issue #985, opened 2022-08-12
Enable makefile to automatically pick up and use a default local rgbds version


etdv-thevoid:
Currently, pokecrystal can be built using a local rgbds version instead of a global installed version. See [INSTALL.md](https://github.com/pret/pokecrystal/blob/master/INSTALL.md#build-with-a-local-rgbds-version) for the current instructions.

It would be convenient to also allow for the makefile to automatically default to whatever rgbds version is stored locally in `pokecrystal/rgbds/`.

PR #986, opened 2022-08-12
Build using default local rgbds folder

etdv-thevoid:
Addresses #985

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,7 +25,6 @@ gfx/pokemon/*/frames.asm
 
 # tool binaries
 *.exe
-*.dll
 
 # precompiled python
 *.pyc
--- a/.gitignore
+++ b/.gitignore
@@ -23,8 +23,12 @@ gfx/pokemon/*/frames.asm
 *.map
 *.sym
 
+# local tools
+rgbds/
+
 # tool binaries
 *.exe
+*.dll
 
 # precompiled python
 *.pyc
--- a/.gitignore
+++ b/.gitignore
@@ -23,6 +23,9 @@ gfx/pokemon/*/frames.asm
 *.map
 *.sym
 
+# local tools
+rgbds*/
+
 # tool binaries
 *.exe
 
--- a/.gitignore
+++ b/.gitignore
@@ -23,9 +23,6 @@ gfx/pokemon/*/frames.asm
 *.map
 *.sym
 
-# local tools
-rgbds/
-
 # tool binaries
 *.exe
 *.dll
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -4,8 +4,25 @@ These instructions explain how to set up the tools required to build **pokecryst
 
 If you run into trouble, ask for help on IRC or Discord (see [README.md](README.md)).
 
+- [Microsoft Windows](#microsoft-windows)
+  - [Windows 10 and 11](#windows-10-and-11)
+  - [Legacy Windows](#legacy-windows)
+- [macOS](#mac-os)
+- [Linux](#linux)
+  - [Debian or Ubuntu](#debian-or-ubuntu)
+  - [OpenSUSE](#opensuse)
+  - [Arch Linux](#arch-linux)
+  - [Termux](#termux)
+  - [Other distros](#other-distros)
+- [Build pokecrystal](#build-pokecrystal)
+  - [Build with a local rgbds version](#build-with-a-local-rgbds-version)
 
-## Windows 10
+
+## Microsoft Windows
+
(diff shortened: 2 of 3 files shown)


.gitignore
```diff
@@ -23,8 +23,12 @@ gfx/pokemon/*/frames.asm
 *.map
 *.sym
 
+# local tools
+rgbds/
+
 # tool binaries
 *.exe
+*.dll
```
vulcandth, line 31:
```suggestion
```
This is not necessary if you are ignoring the `rgbds/` directory.

vulcandth (changes requested):
I personally don't really see this as a big need. People can easily install rgbds wherever they want to. We shouldn't encourage people to install it in a particular directory. Also, most people just choose to install it globally anyways.

However, i'll let Rangi and Mid-kid weigh in on this.

.gitignore
etdv-thevoid: Quite true. But if you install locally on Windows, there are .dll files included with the rgbds pre-compiled downloads that are needed. And if you don't have a global version installed, those are needed.

If you store that version rgbds in a folder such as `pokecrystal/rgbds-v0.5.2/` like INSTALL.md already explains, the .dll files will be picked up.

vulcandth: No, the pokecrystal repo does not produce any `*.dll` files, so they should not be in our `.gitignore`. If the user wants to install rgbds *inside* the pokecrystal repo they can smiply add `rgbds/` to their `.git/info/exclude`.

vulcandth: Furthermore, rbgds is a seperate repo.. They could add any new files they want to anytime, if it helps them with their goals.

etdv-thevoid: I think you are correct with everything you've said so far. But...

> No, the pokecrystal repo does not produce any `*.dll` files, so they should not be in our `.gitignore`. If the user wants to install rgbds _inside_ the pokecrystal repo they can smiply add `rgbds/` to their `.git/info/exclude`. (Or similarly rgbds-v0.5.2/)

At bare minimum, I believe rgbds/ should be added to the .gitignore, or the instructions to use a local version of rgbds should be removed from the INSTALL.md instructions.

These two things are in contradiction. If you follow the approved instructions, you will have things in your repo structure not caught by the default ignore rules.

vulcandth:
This is very similar to #582, that was closed, which I agree with mid-kid's comment in it.
This is also loosely related to #583
I'm going to still let @Rangi42 or @mid-kid give a quick input. However, my stance is still that we shouldn't be encouraging people to necessarily install rgbds in pokecrystal/rgbds/. Users are free to install it wherever, instead we should ensure that they *can* install it anywhere and provide information on how to do so.

etdv-thevoid:
Considering all the above, would it be better this all be scrapped, and instead detailed instructions be added to INSTALL.md explaining how to add `rgbds*/` to `.git/info/exclude`?

vulcandth:
> Considering all the above, would it be better this all be scrapped, and instead detailed instructions be added to INSTALL.md explaining how to add `rgbds*/` to `.git/info/exclude`?

I would prefer this, yes. Currently Mord5 has a PR open to update `INSTALL.md` on pokered https://github.com/pret/pokered/pull/378. If you want to coordinate with them there, we can eventually port that to the rest of the Gen I/II repos.

etdv-thevoid:
> Note the `RGBDS ?=` in the Makefile. This means you can do `RGBDS=rgbds/ make crystal` to use the executables in rgbds/; or even `RGBDS=~/Desktop/rgbds-1.2.3/ make crystal` do it doesn't depend on a particular directory name.

Oh, I was well aware. I just find that tedious is all, so I always end up editing the makefile to use a local file anyways. I have too many versions of rgbds installed!

I'll see about updating the language in INSTALL.md pertaining to local copies of rgbds.

Rangi42:
> I'll see about updating the language in INSTALL.md pertaining to local copies of rgbds.

Please discuss it in #pokecrystal first. INSTALL.md already explains how to use `make RGBDS=...`.

etdv-thevoid:
@Rangi42, I was referring to @vulcandth's comment above about adding to exclude if using a local version, since that info is missing.

> > Considering all the above, would it be better this all be scrapped, and instead detailed instructions be added to INSTALL.md explaining how to add `rgbds*/` to `.git/info/exclude`?
> 
> I would prefer this, yes. Currently Mord5 has a PR open to update `INSTALL.md` on pokered [pret/pokered#378](https://github.com/pret/pokered/pull/378). If you want to coordinate with them there, we can eventually port that to the rest of the Gen I/II repos.

PR #995, opened 2022-08-31
Document fish bug properly

Idain:
The Surf glitch only happens in Pokémon Gold and Silver, it was fixed in Crystal.

Also, `CheckFacingObject` is enough to check if the player is facing an object. Using `GetFacingObject` is an overkill since it also tries to load the object's data in register `d`, which is unnecessary in this specific check.

Tested on vanilla Crystal.


vulcandth:
Thanks Idain! I'm going to review this afternoon. I have a few of my own tests I'd like to do when I get off work on both pokegold/pokecrystal.

Idain:
Nice, let me know if you need anything. Even though I'm at work I always have Discord open (but don't tell my boss, shhhh).

Rangi42:
Nice catch! In that case we can add a "You can Surf on top of NPCs" bug to pokegold's docs/bugs_and_glitches.md.

vulcandth:
> Nice catch! In that case we can add a "You can Surf on top of NPCs" bug to pokegold's docs/bugs\_and\_glitches.md.

Yeah, that was the plan. Just going finish up my testing (make sure we didn't miss anything) and execute that.

PR #993, opened 2022-08-30
Some Mobile code labeling and removed magic numbers

ariscop:
Leaves almost no wram0 references which arent labeled
mobile_46 being an exception, c608 has several layouts and needs a union defined

PR #998, opened 2022-08-31
mobile/mobile_45.asm -> lib/mobile/mail.asm

mid-kid:
This file is part of the official mobile adapter SDK, although unlike
lib/mobile/main.asm, this file was included as source code the
developers could modify.

### DIFF
--- a/lib/mobile/mail.asm
+++ b/lib/mobile/mail.asm
@@ -1,5 +1,14 @@
-pushc
-setcharmap ascii
+; A library included as part of the Mobile Adapter GB SDK.
+
+INCLUDE "macros/code.asm"
+INCLUDE "macros/const.asm"
+INCLUDE "macros/gfx.asm"
+INCLUDE "constants/gfx_constants.asm"
+INCLUDE "constants/hardware_constants.asm"
+INCLUDE "constants/mobile_constants.asm"
+INCLUDE "constants/text_constants.asm"
+
+SECTION "Mobile Adapter SDK Mail", ROMX
 
 String_114000:
 	db "---", 0
@@ -5154,5 +5163,3 @@ Function115d80:
 	ld d, $a0
 	ld e, $0
 	ret
-
-popc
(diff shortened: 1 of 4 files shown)


lib/mobile/mail.asm
```diff
@@ -1,3 +1,14 @@
+INCLUDE "charmap.asm"
```
Rangi42, line 1:
```suggestion
; A library included as part of the Mobile Adapter GB SDK.

INCLUDE "charmap.asm"
```

Rangi42:
Can this refer to "Email" not just "Mail"?

mid-kid:
The term "mail" appears multiple times across the strings in the file as well as other mobile adapter-related material (though that's probably because in japanese they call it メール). I don't feel like creating a situation where some parts say "email" and some parts say "mail".

Rangi42:
I'd like to distinguish between email and in-game Mail.

mid-kid:
I'm not sure I understand the concern, then. If you're concerned about searchability, I'm sure queries for "mail" will turn up "email" too, unfortunately. If you're concerned about prefixes, then either `MobileMail_` or `GBMail_` seem more appropriate to me. But sure, if we go with email, we should try do it consistently.

vulcandth:
> I'm not sure I understand the concern, then. If you're concerned about searchability, I'm sure queries for "mail" will turn up "email" too, unfortunately. If you're concerned about prefixes, then either `MobileMail_` or `GBMail_` seem more appropriate to me. But sure, if we go with email, we should try do it consistently.

I'd be fine with `email` or `MobileMail_`. 

However, I feel like this shouldn't hold us up. When we get closer to finishing the identification of the mobile code we can re-address how we want to name things. I'd just pick something that is searchable and won't turn up held Pokémon mail for now, that way it will make it easier to search and rename later if we want. (I'm not as concerned about searching the other way around at the moment.) So `email` or `MobileMail_` for seems fine for that purpose.

vulcandth (approved):
After taking another look at this; I think this is fine as is, imo.

mid-kid:
Merging for now since it doesn't seem to conflict with #993, feel free to change anything if there's any issues.

PR #999, opened 2022-08-31
Fix copypasta error in bugdocs

### LINKED ISSUE #997: Copy+paste error in bug docs
https://github.com/pret/pokecrystal/blob/master/docs/bugs_and_glitches.md#checkownmonanywhere-does-not-check-the-day-care

coltongit:
Fixes #997

### DIFF
--- a/docs/bugs_and_glitches.md
+++ b/docs/bugs_and_glitches.md
@@ -2439,7 +2439,6 @@ This bug can prevent you from talking to Eusine in Celadon City or encountering 
 ```diff
  	; If there are no monsters in the party,
  	; the player must not own any yet.
--; BUG: CheckOwnMon does not check the Day-Care (see docs/bugs_and_glitches.md)
  	ld a, [wPartyCount]
  	and a
  	ret z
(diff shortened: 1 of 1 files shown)


docs/bugs_and_glitches.md
```diff
@@ -2439,7 +2439,7 @@ This bug can prevent you from talking to Eusine in Celadon City or encountering
 ```diff
  	; If there are no monsters in the party,
  	; the player must not own any yet.
--; BUG: CheckOwnMon does not check the Day-Care (see docs/bugs_and_glitches.md)
```
Rangi42, line 2442: Just remove everything in the diff prior to the actual `BUG:` comment.

coltongit: Is replacing that with ... OK too?

docs/bugs_and_glitches.md
```diff
@@ -2437,13 +2437,7 @@ This bug can prevent you from talking to Eusine in Celadon City or encountering
...
 
 ```diff
- 	; If there are no monsters in the party,
- 	; the player must not own any yet.
--; BUG: CheckOwnMon does not check the Day-Care (see docs/bugs_and_glitches.md)
- 	ld a, [wPartyCount]
- 	and a
- 	ret z
-
+...
```
Rangi42, line 2440: Why?
```suggestion
...
```

coltongit: I don't have that character/I'm not sure how to type it. Thanks

Issue #1001, opened 2022-09-01
Rename `macros/wram.asm` to `macros/ram.asm`


vulcandth:
The macros in `macros/wram.asm` are, and will be used, in other ram locations.

Idain:
In which locations are they being used?

vulcandth:
> In which locations are they being used?

@Idain 
The following MACRO's are used elsewhere (there may be more)
`box` is used in sram.asm
`mailmsg` is used in wram.asm & sram.asm
`hall_of_fame` is used in wram.asm & sram.asm

And there will be plans to add more with the identification of the mobile code.

Idain:
Interesting. I support the name change!

PR #1003, opened 2022-09-04
Renamed `macros/wram.asm` to `macros/ram.asm`

vulcandth:
Fixes: #1001 

If one of you can give this a quick glance before merge.


Rangi42:
Looks good, thanks! Please do the same for the other repos.

PR #1004, opened 2022-09-04
Use RS Group offset constants for `slot_reel`

vulcandth:
Noticed this was still done the "old way"; I prefer using the rs group for offset constants.

### DIFF
--- a/engine/games/slot_machine.asm
+++ b/engine/games/slot_machine.asm
@@ -15,17 +15,19 @@ DEF SLOTS_NO_BIAS EQU -1
 DEF REEL_SIZE EQU 15
 
 ; Constants for slot_reel offsets (see macros/wram.asm)
-DEF REEL_ACTION        EQUS "(wReel1ReelAction - wReel1)"
-DEF REEL_TILEMAP_ADDR  EQUS "(wReel1TilemapAddr - wReel1)"
-DEF REEL_POSITION      EQUS "(wReel1Position - wReel1)"
-DEF REEL_SPIN_DISTANCE EQUS "(wReel1SpinDistance - wReel1)"
-DEF REEL_SPIN_RATE     EQUS "(wReel1SpinRate - wReel1)"
-DEF REEL_OAM_ADDR      EQUS "(wReel1OAMAddr - wReel1)"
-DEF REEL_X_COORD       EQUS "(wReel1XCoord - wReel1)"
-DEF REEL_MANIP_COUNTER EQUS "(wReel1ManipCounter - wReel1)"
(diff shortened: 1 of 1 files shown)


mid-kid:
I'd try to comment the label offsets...

engine/games/slot_machine.asm
```diff
@@ -15,17 +15,21 @@ DEF SLOTS_NO_BIAS EQU -1
...
+DEF REEL_SPIN_DISTANCE rb
+DEF REEL_SPIN_RATE     rb
+DEF REEL_OAM_ADDR      rw
+DEF REEL_X_COORD       rb
+DEF REEL_MANIP_COUNTER rb
+DEF REEL_MANIP_DELAY   rb
+DEF REEL_FIELD_0B      rb
+DEF REEL_FIELD_0C      rb
+DEF REEL_FIELD_0D      rb
+DEF REEL_FIELD_0E      rb
```
Rangi42, line 31:
```suggestion
                       rb_skip 4
```

vulcandth:
> I'd try to comment the label offsets...

@mid-kid What exactly did you have in mind for this? Did you want to add what the offsets are as constants? Cause we don't really do that for any of the other rb group's that I can tell. There is a comment already above this that associates these offset constants with the wram labels they are associated with.

Are you thinking like the following? Or am I misunderstanding?
```asm
rsreset
DEF REEL_ACTION             rb ; $00
DEF REEL_TILEMAP_ADDR rw ; $01
DEF REEL_POSITION          rb ; $03
...
```

Rangi42:
Just `; 0`, `; 1`, etc would be fine.

engine/games/slot_machine.asm
vulcandth: REEL_FIELD_0B is used; So I left it in.

Rangi42: Looks like it's only used for the countdown after starting the `REEL_ACTION_WAIT_DROP_REEL` action.

```diff
ReelAction_WaitDropReel:
-	ld hl, REEL_FIELD_0B
+	ld hl, REEL_DROP_COUNTER
	add hl, bc
	ld a, [hl]
	and a
	jr z, .DropReel
	dec [hl]
	ret

.DropReel:
	ld hl, REEL_ACTION
	add hl, bc
-	dec [hl]
+	dec [hl] ; REEL_ACTION_CHECK_DROP_REEL
	ld hl, REEL_SPIN_RATE
	add hl, bc
	ld [hl], 8
	ret
```

mid-kid:
I meant something like `; wReel*TilemapAddr`, I'd much rather get rid of address/offset comments...

PR #1005, opened 2022-09-10
Rename the `object_struct` field labels and constants

### LINKED ISSUE #990: `object_struct` wram labels should match their offset constants
Currently the wram labels for the `object_struct` have things such as:

```
\1StandingMapX::      db
\1StandingMapY::      db
\1LastMapX::          db
\1LastMapY::          db
```

However the offset constants use:

```
DEF OBJECT_NEXT_MAP_X          rb ; 10
DEF OBJECT_NEXT_MAP_Y          rb ; 11
DEF OBJECT_MAP_X               rb ; 12
DEF OBJECT_MAP_Y               rb ; 13
```

In the above examples, the wram labels should be renamed to match the offset constants. There are a few others here than need addressed.

Also we should investigate the true purpose of the following, and come up with better names for these:

```
DEF OBJECT_1C                  rb ; 1c
DEF OBJECT_1D                  rb ; 1d
DEF OBJECT_1E                  rb ; 1e
DEF OBJECT_1F                  rb ; 1f
```

Rangi42:
Fixes #990

### DIFF
--- a/constants/map_object_constants.asm
+++ b/constants/map_object_constants.asm
@@ -1,43 +1,43 @@
 ; object_struct members (see macros/ram.asm)
 rsreset
-DEF OBJECT_SPRITE              rb ; 00
-DEF OBJECT_MAP_OBJECT_INDEX    rb ; 01
-DEF OBJECT_SPRITE_TILE         rb ; 02
-DEF OBJECT_MOVEMENTTYPE        rb ; 03
-DEF OBJECT_FLAGS1              rb ; 04
-DEF OBJECT_FLAGS2              rb ; 05
-DEF OBJECT_PALETTE             rb ; 06
-DEF OBJECT_DIRECTION_WALKING   rb ; 07
-DEF OBJECT_FACING              rb ; 08
-DEF OBJECT_STEP_TYPE           rb ; 09
-DEF OBJECT_STEP_DURATION       rb ; 0a
-DEF OBJECT_ACTION              rb ; 0b
-DEF OBJECT_STEP_FRAME          rb ; 0c
-DEF OBJECT_FACING_STEP         rb ; 0d
-DEF OBJECT_NEXT_TILE           rb ; 0e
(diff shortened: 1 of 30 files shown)


constants/map_object_constants.asm
```diff
@@ -31,13 +31,13 @@ DEF OBJECT_MOVEMENT_BYTE_INDEX rb ; 1b
 DEF OBJECT_1C                  rb ; 1c
```
vulcandth, line 31:
```suggestion
DEF OBJECT_ANON_JUMPTABLE_INDEX                  rb ; 1c
```
You'll have to fix the formatting and the associated wram label, and everywhere it's used.

constants/map_object_constants.asm
```diff
@@ -31,13 +31,13 @@ DEF OBJECT_MOVEMENT_BYTE_INDEX rb ; 1b
...
 DEF OBJECT_1D                  rb ; 1d
 DEF OBJECT_1E                  rb ; 1e
```
vulcandth, line 33: These seem to be used for a couple purposes... One of which seems to store the address of the start of another object used in `StepFunction_TrackingObject`. So it's fine they remain as they are now.. just wanted to mention it.

vulcandth (changes requested):
After that one requested change; I think this is good to merge. Thanks 👍

constants/map_object_constants.asm
vulcandth: (If you agree with this, I can push the required changes to the PR for you. Since I know you are busy.)

vulcandth (approved):

vulcandth (approved):

mid-kid:
`CMDQUEUE_ANON_JUMPTABLE_INDEX` -> `CMDQUEUE_JUMPTABLE_INDEX` honestly. The only thing that makes it "anonymous" is an implementation detail.

vulcandth (approved):

mid-kid (approved):

PR #1010, opened 2022-09-27
Generate the correct Stadium base data for every ROM

### LINKED ISSUE #752: Reverse-engineer the Crystal "base" data inserted by tools/stadium.c
This is 24 bytes of data starting with ASCII "`base`" that come right before the Stadium 2 checksums in Crystal. It's not present in G/S.

One version of this data is in the US 1.0, US 1.1, and AU ROMs; another is in the EU ROMs; another is in the US 1,0 and 1.1 debug ROMs. Some of the early Crystal builds have still other versions, or no such data.

https://github.com/pret/pokecrystal/blob/master/tools/stadium.c#L27-L37

The original xtal, xtal_revise, and xtal_AUSTRALIA pmcenv each have a crystal_base0.bin file; and xtal_euro pmcenv has a crystal_base1.bin file. These are particular ROM builds without the base or Stadium data at the end.

It may not be possible to reverse-engineer, in which case tools/stadium.c will continue to insert it from hard-coded values.

Rangi42:
Fixes #752.

The `-e` flag was tested and works for all the European ROMs mentioned in #626.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -159,17 +159,10 @@ pokecrystal_debug_opt   = -Cjv -t PM_CRYSTAL -i BYTE -n 0 -k 01 -l 0x33 -m 0x10 
 pokecrystal11_debug_opt = -Cjv -t PM_CRYSTAL -i BYTE -n 1 -k 01 -l 0x33 -m 0x10 -r 3 -p 0
 pokecrystal11_vc_opt    = -Cjv -t PM_CRYSTAL -i BYTE -n 1 -k 01 -l 0x33 -m 0x10 -r 3 -p 0
 
-pokecrystal_base         = us
-pokecrystal11_base       = us
-pokecrystal_au_base      = us
-pokecrystal11_vc_base    = us
-pokecrystal_debug_base   = dbg
-pokecrystal11_debug_base = dbg
-
 %.gbc: $$(%_obj) layout.link
 	$(RGBLINK) -n $*.sym -m $*.map -l layout.link -o $@ $(filter %.o,$^)
 	$(RGBFIX) $($*_opt) $@
-	tools/stadium --base $($*_base) $@
+	tools/stadium $@
 
 
 ### LZ compression rules
--- a/tools/stadium.c
+++ b/tools/stadium.c
@@ -67,7 +67,7 @@ uint16_t crc_table[256];
 	file[(off) + 1] = (uint8_t)(((v) & 0x00FF) >> 0); \
 } while (0)
 
-// Checksums of every bank in the base ROM, crystal_base0.bin
+// CRCs of every bank in the base ROM, crystal_base0.bin
 uint16_t base0_crcs[NUMBANKS] = {
 	0x9650, 0x8039, 0x2D8F, 0xD75A, 0xAC50, 0x5D55, 0xE94B, 0x9886,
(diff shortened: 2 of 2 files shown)


Rangi42:
The base data stores a bit for each bank of the built ROM, indicating whether it matches the bank in the base file (crystal_base0.bin for US, crystal_base1.bin for EU). We don't want to store those whole files, so stadium.c just stores CRCs for each bank of the base .bin files, and compares them with CRCs of the built ROM banks. This runs the risk of CRC collisions, but works in practice for all the official ROMs.

If we want to avoid storing any information derived from those two .bin files, we could do as @mid-kid suggested: store CRCs for the banks that have 1 bits (since those can be derived from the matching ROMs), and store flags for the rest indicating that they don't match. (So the `base0_crcs` and `base1_crcs` arrays could store `int32_t`s instead of `uint16_t`s, with `-1` indicating a non-match.) Then if you build a custom ROM, it might have even fewer banks than the original ROM that match the base .bin data, but it won't have more. (Technically it *could* but we wouldn't be able to know that from the data available.)

mid-kid:
@vulcandth The idea is to store a "dummy" CRC value for the banks we can't recover. If the value is a dummy, it's automatically a 1.

Rangi42:
We do already source data from the leaks, since we build `CRYSTAL_ps3_010328d.bin` and `CRYSTAL_ps3_us_revise_010710d.bin`. And the `crystal_base0/1.bin` files would need mentioning in a comment to explain what the 0 and 1 bit flags even mean. So I'd rather just leave all the base bank checksums as-is.

Other changes to make (do them to pokegold as well when applicable):

- [ ] Group the constants by what chunk of data they refer to, instead of whether they're an offset or a size
- [ ] `DATAOFF -> STADIUMOFF` (or anything more specific than "data")
- [ ] Signature of `calculate_checksum` and `calculate_crc` should match `memcpy` and `memset` (offset pointer, not base pointer and start index)

Rangi42:
@mid-kid I've made the changes you requested. Will do the same to pokegold after this merges.

mid-kid:
LGTM now, thanks for making the changes.

Makefile
```diff
@@ -159,17 +159,10 @@ pokecrystal_debug_opt   = -Cjv -t PM_CRYSTAL -i BYTE -n 0 -k 01 -l 0x33 -m 0x10
...
-pokecrystal_au_base      = us
-pokecrystal11_vc_base    = us
-pokecrystal_debug_base   = dbg
-pokecrystal11_debug_base = dbg
-
 %.gbc: $$(%_obj) layout.link
 	$(RGBLINK) -n $*.sym -m $*.map -l layout.link -o $@ $(filter %.o,$^)
 	$(RGBFIX) $($*_opt) $@
-	tools/stadium --base $($*_base) $@
+	tools/stadium $@
```
vulcandth, line 165: This is a lot simpler for now; However we will have to modify this area of the Makefile again for the i18n branches in the future. Shouldn't be too difficult.

vulcandth (approved):
Yeah this looks fine to me too. Good work.

Makefile
Rangi42: Yeah, I'm not trying to anticipate how we'll end up supporting i18n builds. (I'd like the eventual solution to not be disruptive for users who just want to use one language as a hack base, usually English.)

PR #1014, opened 2022-10-15
Misspelt word

Idain:
These are the only instances I found. If someone has the patience to keep searching, be my guest.

### DIFF
--- a/engine/battle_anims/functions.asm
+++ b/engine/battle_anims/functions.asm
@@ -1831,7 +1831,7 @@ BattleAnimFunction_Absorb:
 	ret
 
 BattleAnimFunction_Wrap:
-; Plays out object Frameset. Use anim_incobj to move to next frameset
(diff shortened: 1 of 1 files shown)


engine/battle_anims/functions.asm
```diff
@@ -3614,7 +3614,7 @@ BattleAnimFunction_MetronomeHand:
 
 BattleAnimFunction_MetronomeSparkleSketch:
 ; Sideways wave motion while also moving downward until it disappears at y coord $20
-; Obj Param: Is used internally only
+; Obj Param: It's used internally only
```
aaaaaa123456789, line 3617: I'd remove the verb entirely to match the way other comments are written. (`Obj Param: used internally only`.)

Idain: On my way.

Rangi42 (approved):
Looks good, thanks! Please make the same changes for pokegold.

PR #1018, opened 2022-11-08
`.CheckNPC` label and comment correction

vulcandth:
The `.is_npc` label is incorrect. In fact, returning 1 means that there is NOT an NPC in front.

Also I expanded on the `.no_bump` comment (returning 2) as this actually does the SAME as if you returned 0.

See:  
https://github.com/pret/pokecrystal/blob/41d5ea0482f4ef577df600d2d8b5cad70f74a396/engine/overworld/player_movement.asm#L267-L271  
and  
https://github.com/pret/pokecrystal/blob/41d5ea0482f4ef577df600d2d8b5cad70f74a396/engine/overworld/player_movement.asm#L325-L330

The former of which has an unused/unreferenced local label next to the `.bump` label. Presumably this would have handled the `.no_bump` functionality at one time.

### DIFF
--- a/engine/overworld/player_movement.asm
+++ b/engine/overworld/player_movement.asm
@@ -613,7 +613,8 @@ ENDM
 .CheckNPC:
 ; Returns 0 if there is an NPC in front that you can't move
 ; Returns 1 if there is no NPC in front
-; Returns 2 if there is a movable NPC in front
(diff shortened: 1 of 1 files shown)


mid-kid:
I wonder if `wWalkingIntoNPC` deserves renaming as a result of this, but that'd be the subject of a different PR, since there's more variables following the same scheme.

PR #1021, opened 2022-11-30
Document PRZ/BRN enemy switched stat debuff bug

### LINKED ISSUE #746: Bug: PRZ and BRN stat debuffs are not always applied to enemy mon when switching

vulcandth:
Alright I've done some debugging and testing and I believe I have this bug figured out.

The Problem: 
1. `ApplyStatusEffectOnEnemyStats` is only ever called from `InitEnemyMon`.
2. Enemy Mon's (wild, trainer, ect) use `LoadEnemyMon:` for first load and during switching (both forced and AI), with the exception of link battles and the BattleTower, which both use `InitEnemyMon` instead.

Some of the first few lines of `LoadEnemyMon` checks for if we are in a link battle or a BattleTower-Battle:
https://github.com/pret/pokecrystal/blob/c4e3f2bbc23fe731deb177ed5743fdc6ac6572e0/engine/battle/core.asm#L5967-L5975

Since all other Enemy Mons use the full `LoadEnemyMon` routine, they never call `ApplyStatusEffectOnEnemyStats`. Since the stats are recalculated back to base stats during load, the Speed/Attack stat is never lowered based on Status. By simply adding the `ApplyStatusEffectOnEnemyStats` to the end of the `LoadEnemyMon` resolves this issue. This is similar to how `InitEnemyMon` and `InitBattleMon` functions.

@Rangi42 Questions:
>Does it affect link battles too? InitEnemyMon and Function_SetEnemyMonAndSendOutAnimation are called in link battles, but I'd expect a switch-related bug to have been noticed sooner in them.

It does not affect link battles or the BattleTower.

>Does the fix break anything? In particular, if an enemy mon already has PRZ/BRN when the battle starts, would this fix prevent the stat drop? If so, we can't just leave both ApplyStatusEffectOnEnemyStats calls in, since that would change this bug from no drop to a double drop.

I did a fair bit of testing. It appears to work as intended. I also tested an enemy mon already having PRZ/BRN when the battle starts, using some debugging.. There is no double stat drop. Also looking over the code and where it is placed, I don't see any other conflicts that the `ApplyStatusEffectOnEnemyStats` routine would cause here.

Steps to Reproduce problem: (Using forced switch method)
1. Find a trainer with two pokemon.
2. Thunder wave the first pokemon.
3. Observe `wEnemyMonSpeed` stats decrease.
4. Use Whirlwind
5. Kill other mon.
6. PRZ pokemon will return with full Speed stat.

Resolves #746

### DIFF
--- a/engine/battle/core.asm
+++ b/engine/battle/core.asm
@@ -6407,7 +6407,7 @@ LoadEnemyMon:
 	ld bc, NUM_EXP_STATS * 2
 	call CopyBytes
 
-; BUG: PRZ and BRN stat debuffs sometimes don't apply to switched mons (see docs/bugs_and_glitches.md)
+; BUG: PRZ and BRN stat reductions don't apply to switched Pokémon (see docs/bugs_and_glitches.md)
 	ret
 
 CheckSleepingTreeMon:
--- a/engine/battle/core.asm
+++ b/engine/battle/core.asm
@@ -6407,6 +6407,7 @@ LoadEnemyMon:
 	ld bc, NUM_EXP_STATS * 2
 	call CopyBytes
 
+; BUG: PRZ and BRN stat debuffs sometimes don't apply to switched mons (see docs/bugs_and_glitches.md)
 	ret
 
 CheckSleepingTreeMon:
--- a/docs/bugs_and_glitches.md
+++ b/docs/bugs_and_glitches.md
@@ -50,6 +50,7 @@ Fixes in the [multi-player battle engine](#multi-player-battle-engine) category 
   - [Love Ball boosts catch rate for the wrong gender](#love-ball-boosts-catch-rate-for-the-wrong-gender)
   - [Fast Ball only boosts catch rate for three Pokémon](#fast-ball-only-boosts-catch-rate-for-three-pok%C3%A9mon)
   - [Heavy Ball uses wrong weight value for three Pokémon](#heavy-ball-uses-wrong-weight-value-for-three-pok%C3%A9mon)
+  - [PRZ and BRN stat reductions don't apply to switched Pokémon](#prz-and-brn-stat-reductions-dont-apply-to-switched-pok%C3%A9mon)
   - [Glacier Badge may not boost Special Defense depending on the value of Special Attack](#glacier-badge-may-not-boost-special-defense-depending-on-the-value-of-special-attack)
   - ["Smart" AI encourages Mean Look if its own Pokémon is badly poisoned](#smart-ai-encourages-mean-look-if-its-own-pok%C3%A9mon-is-badly-poisoned)
   - ["Smart" AI discourages Conversion2 after the first turn](#smart-ai-discourages-conversion2-after-the-first-turn)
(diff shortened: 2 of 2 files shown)


Rangi42:
Thanks for taking the time to test this bugfix! Please undo the tail call optimization though; it obscures that the only fix necessary is to add a `call`, and there' no limit to how much "nearby optimization" other fixes could do but don't.

Also I'd like to revise some language before merge, e.g. "mons" outside of casual code comments.

vulcandth:
Alright I rolled back the tailcall optimization commit. Feel free to adjust wording as you see fit, or I can go back and reword things if you want. 👍

mid-kid:
Issues like these always make me wonder if `InitEnemyMon` and `LoadEnemyMon` are appropriate names - if they're used only in certain circumstances (i.e. only for AI mon), maybe it can use a name that reflects that.

vulcandth:
> EDIT: You know this by now, don't feel obliged to act on this, but I haven't read into the code as much as you have, so maybe you have an idea so something like this can be done in a different PR.

Yeah... The names are annoyingly similar. I was already thinking the same as you.. but i'll save that for for another PR.. for another time. I'd still need to figure out why GF decided to break them out the way they did. Anywho. 👍

engine/battle/core.asm
```diff
@@ -6407,6 +6407,7 @@ LoadEnemyMon:
 	ld bc, NUM_EXP_STATS * 2
 	call CopyBytes
 
+; BUG: PRZ and BRN stat debuffs sometimes don't apply to switched mons (see docs/bugs_and_glitches.md)
```
Rangi42, line 6410:
```suggestion
; BUG: PRZ and BRN stat reductions don't apply to switched Pokémon (see docs/bugs_and_glitches.md)
```

PR #1027, opened 2022-12-29
Ensure linked songs are in the same bank

dannye:
If someone decides to shuffle song data around, this prevents certain songs from silently breaking.

### DIFF
--- a/audio.asm
+++ b/audio.asm
@@ -77,23 +77,32 @@ SECTION "Songs 4", ROMX
 
 INCLUDE "audio/music/viridiancity.asm"
 INCLUDE "audio/music/celadoncity.asm"
+
 INCLUDE "audio/music/wildpokemonvictory.asm"
 INCLUDE "audio/music/successfulcapture.asm"
+assert BANK(Music_WildPokemonVictory) == BANK(Music_SuccessfulCapture)
+
 INCLUDE "audio/music/gymleadervictory.asm"
 INCLUDE "audio/music/mtmoonsquare.asm"
 INCLUDE "audio/music/gym.asm"
 INCLUDE "audio/music/pallettown.asm"
 INCLUDE "audio/music/profoakspokemontalk.asm"
 INCLUDE "audio/music/profoak.asm"
+
 INCLUDE "audio/music/lookrival.asm"
 INCLUDE "audio/music/aftertherivalfight.asm"
+assert BANK(Music_LookRival) == BANK(Music_AfterTheRivalFight)
+
 INCLUDE "audio/music/surf.asm"
 INCLUDE "audio/music/nationalpark.asm"
 INCLUDE "audio/music/azaleatown.asm"
 INCLUDE "audio/music/cherrygrovecity.asm"
 INCLUDE "audio/music/unioncave.asm"
+
 INCLUDE "audio/music/johtowildbattle.asm"
 INCLUDE "audio/music/johtowildbattlenight.asm"
+assert BANK(Music_JohtoWildBattle) == BANK(Music_JohtoWildBattleNight)
+
 INCLUDE "audio/music/johtotrainerbattle.asm"
 INCLUDE "audio/music/lookyoungster.asm"
 INCLUDE "audio/music/tintower.asm"
@@ -106,8 +115,11 @@ INCLUDE "audio/music/pokemonmarch.asm"
 INCLUDE "audio/music/goldsilveropening.asm"
 INCLUDE "audio/music/goldsilveropening2.asm"
 INCLUDE "audio/music/lookhiker.asm"
+
 INCLUDE "audio/music/lookrocket.asm"
 INCLUDE "audio/music/rockettheme.asm"
+assert BANK(Music_LookRocket) == BANK(Music_RocketTheme)
+
 INCLUDE "audio/music/mainmenu.asm"
 INCLUDE "audio/music/lookkimonogirl.asm"
 INCLUDE "audio/music/pokeflutechannel.asm"


audio.asm
```diff
@@ -77,23 +77,32 @@ SECTION "Songs 4", ROMX
 
 INCLUDE "audio/music/viridiancity.asm"
 INCLUDE "audio/music/celadoncity.asm"
+
 INCLUDE "audio/music/wildpokemonvictory.asm"
 INCLUDE "audio/music/successfulcapture.asm"
+assert BANK(Music_WildPokemonVictory) == BANK(Music_SuccessfulCapture)
```
vulcandth, line 83:
```suggestion
assert BANK(Music_WildPokemonVictory) == BANK(Music_SuccessfulCapture), \
	"audio/music/successfulcapture.asm must be in the same bank as audio/music/wildpokemonvictory.asm!"
```

audio.asm
```diff
@@ -77,23 +77,32 @@ SECTION "Songs 4", ROMX
...
 INCLUDE "audio/music/gymleadervictory.asm"
 INCLUDE "audio/music/mtmoonsquare.asm"
 INCLUDE "audio/music/gym.asm"
 INCLUDE "audio/music/pallettown.asm"
 INCLUDE "audio/music/profoakspokemontalk.asm"
 INCLUDE "audio/music/profoak.asm"
+
 INCLUDE "audio/music/lookrival.asm"
 INCLUDE "audio/music/aftertherivalfight.asm"
+assert BANK(Music_LookRival) == BANK(Music_AfterTheRivalFight)
```
vulcandth, line 94:
```suggestion
assert BANK(Music_LookRival) == BANK(Music_AfterTheRivalFight), \
	"audio/music/aftertherivalfight.asm must be in the same bank as audio/music/lookrival.asm!"
```

audio.asm
```diff
@@ -77,23 +77,32 @@ SECTION "Songs 4", ROMX
...
+
 INCLUDE "audio/music/surf.asm"
 INCLUDE "audio/music/nationalpark.asm"
 INCLUDE "audio/music/azaleatown.asm"
 INCLUDE "audio/music/cherrygrovecity.asm"
 INCLUDE "audio/music/unioncave.asm"
+
 INCLUDE "audio/music/johtowildbattle.asm"
 INCLUDE "audio/music/johtowildbattlenight.asm"
+assert BANK(Music_JohtoWildBattle) == BANK(Music_JohtoWildBattleNight)
```
vulcandth, line 104:
```suggestion
assert BANK(Music_JohtoWildBattle) == BANK(Music_JohtoWildBattleNight), \
	"audio/music/jotowildbattlenight.asm must be in the same bank as audio/music/johtowildbattle.asm!"
```

audio.asm
```diff
@@ -106,8 +115,11 @@ INCLUDE "audio/music/pokemonmarch.asm"
 INCLUDE "audio/music/goldsilveropening.asm"
 INCLUDE "audio/music/goldsilveropening2.asm"
 INCLUDE "audio/music/lookhiker.asm"
+
 INCLUDE "audio/music/lookrocket.asm"
 INCLUDE "audio/music/rockettheme.asm"
+assert BANK(Music_LookRocket) == BANK(Music_RocketTheme)
```
vulcandth, line 121:
```suggestion
assert BANK(Music_LookRocket) == BANK(Music_RocketTheme), \
	"audio/music/rockettheme.asm must be in the same bank as audio/music/lookrocket.asm!"
```

vulcandth (changes requested):
Thanks! I personally prefer to be a little more descriptive in my asserts. Anything that will help the user easier understand what the problem is. Let me know what your thoughts are on adding comments like I suggested.

dannye:
It's a fair point. I'm hesitant to start hard-coding paths/filenames in the assert message because updating them could be a nuisance. The very similar [asserts in sfx.asm](https://github.com/pret/pokecrystal/blob/master/audio/sfx.asm#L25) do not have error messages. And I don't expect these asserts to be triggered by newcomers. If someone is making big changes to the layout of the songs, they probably mostly know what they're doing and wouldn't be stumped for long by the assertion failure.

Rangi42:
I'm not really sure about adding this, since many other separate files INCLUDEd in main.asm are supposed to be in the same bank too. We could wait for a "sectionful refactor" to make them all explicit.

dannye:
> many other separate files INCLUDEd in main.asm are supposed to be in the same bank too.

Sure, but it's rather counterintuitive that song files sometimes depend on other song files, so it's much easier to make the incorrect assumption that song files can be moved freely, compared to two side-by-side code `INCLUDE`s so I think it's more warranted here.

> We could wait for a "sectionful refactor" to make them all explicit.

It could be re-evaluted at that time but I don't think it has to wait necessarily.

vulcandth:
I agree with @dannye here. I would have assumed most song files are moveable myself. It might have tripped me up too. I think in this case it is warranted.

vulcandth:
> It's a fair point. I'm hesitant to start hard-coding paths/filenames in the assert message because updating them could be a nuisance. The very similar [asserts in sfx.asm](https://github.com/pret/pokecrystal/blob/master/audio/sfx.asm#L25) do not have error messages. And I don't expect these asserts to be triggered by newcomers. If someone is making big changes to the layout of the songs, they probably mostly know what they're doing and wouldn't be stumped for long by the assertion failure.

Sounds good. Now that I think about it.. if people are moving them to other banks.. they will probably see the assert in the code while they are doing it. So the actual message is probably not necessary here.

vulcandth (approved):

Rangi42 (approved):

PR #1039, opened 2023-02-24
Document scripting engine fix for MBC30

mid-kid:


mid-kid:
Turns out the bit it's masking was used for something, drafting...

mid-kid:
Fixed

vulcandth (approved):
Makes sense and looks good to me. Assuming you tested it already (I'm sure you have), you can merge it unless you want Rangi to take at look as well.

mid-kid:
I haven't extensively tested this, but I've checked every use of wScriptBank and wScriptStack and nothing indicated it's actually being used anywhere, so I think it's fine.

Issue #1048, opened 2023-05-04
BUG: free_space.awk doesn't work with rgbds 0.6.1.


SonicRay100:
free_space.awk doesn't work with rgbds 0.6.1 while with rgbds 0.6.0 it works just fine.

What I have tested:

- Tested both crystal and crystal11.
- Make clean and recompiled (also deleting all compiled files that make clean doesn't delete) before testing between 0.6.1 and 0.6.0.
- Tested with a newly downloaded repository.
- Tested free_space.py and that doesn't work either.

What happens:

free_space.awk:

- shows 0% free space is available.
- Does not detect any individual banks.

free_space.py:

- Shows 100% free space is available.
- Does detect individual banks.
- Individual banks all show 100% free space available.

![1](https://user-images.githubusercontent.com/69745689/236128928-03daff12-e82b-4a72-9744-d0c405924643.png)

SonicRay100:
Can confirm that pull request  #1049 fixes this issue for free_space.awk.

PR #1049, opened 2023-05-04
Fix tools/free_space.awk for new rgbds

mid-kid:
See https://github.com/pret/pokecrystal/issues/1048


SonicRay100:
Works perfectly.

mid-kid:
@vulcandth this ought to be merged into pokegold and PRd into pokered and other gen1 repos. I didn't merge this myself because of that.

PR #1052, opened 2023-05-31
Use BattleAnim_Dummy in data/moves/animations.asm

### LINKED ISSUE #1047: Dummy labels in data/moves/animations.asm
There's a number of dummy animation labels in this file: `BattleAnim_0`, `BattleAnim_252`, `BattleAnim_253`, `BattleAnim_254` and `BattleAnim_MirrorMove` all point to the same dummy/empty animation. It seems silly to have five labels pointing to the same address — they should be a single `BattleAnim_Dummy` IMO.

vulcandth:
Resolves #1047

For the one that used to be `BattleAnim_MirrorMove` that is now `BattleAnim_Dummy` should we add a comment next to it mentioning mirror move? Might help making searching for it easier.

### DIFF
--- a/data/moves/animations.asm
+++ b/data/moves/animations.asm
@@ -120,7 +120,7 @@ BattleAnimations::
 	dw BattleAnim_FocusEnergy
 	dw BattleAnim_Bide
 	dw BattleAnim_Metronome
-	dw BattleAnim_Dummy
+	dw BattleAnim_MirrorMove
 	dw BattleAnim_Selfdestruct
 	dw BattleAnim_EggBomb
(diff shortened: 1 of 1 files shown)


Rangi42 (changes requested):
I disagree that these should all share one dummy label. `MusicCommands` is another instance where we have multiple labels for one `ret`, because they're all different things. At the very least I'd keep `BattleAnim_MirrorMove` as a second label.

vulcandth:
I added the second label; but I still feel that its unnecessary to have all the redundant labels.

Rangi42 (approved):

PR #1054, opened 2023-05-31
Define relationship between map_object and object_events

### LINKED ISSUE #1031: Further define relationship between the `map_object` struct and `object_events`
The relationship between the `map_object` struct and `object_event`'s should be further defined.

`object_event`'s are loaded into the `map_object` struct by means of `CopyMapObjectEvents`. There is a clear and corresponding relationship between these two. The only items in the `map_object` struct that isn't copied from the Map's `object_event`'s is `MAPOBJECT_OBJECT_STRUCT_ID` which is initially set to `-1` until the object appears on screen (_in which case it is then loaded with the index of the corresponding `object_struct`_), and the two extra bytes at the end of the struct represented by the `rb_skip 2`.

https://github.com/pret/pokecrystal/blob/c01409be5a9930d6f0687ce53c1c898e1855c884/constants/map_object_constants.asm#L99-L118

https://github.com/pret/pokecrystal/blob/c01409be5a9930d6f0687ce53c1c898e1855c884/macros/scripts/maps.asm#L105-L132

https://github.com/pret/pokecrystal/blob/c01409be5a9930d6f0687ce53c1c898e1855c884/home/map.asm#L610-L634

https://github.com/pret/pokecrystal/blob/41d5ea0482f4ef577df600d2d8b5cad70f74a396/constants/script_constants.asm#L115

Recommendation: `OBJECT_EVENT_SIZE` should be defined by the `map_object` struct. Example:

```diff
 ; map_object struct members (see macros/ram.asm)
 rsreset
 DEF MAPOBJECT_OBJECT_STRUCT_ID rb ; 0
 DEF MAPOBJECT_SPRITE           rb ; 1
 DEF MAPOBJECT_Y_COORD          rb ; 2
 DEF MAPOBJECT_X_COORD          rb ; 3
 DEF MAPOBJECT_MOVEMENT         rb ; 4
 DEF MAPOBJECT_RADIUS           rb ; 5
 DEF MAPOBJECT_HOUR_1           rb ; 6
 DEF MAPOBJECT_HOUR_2           rb ; 7
 rsset MAPOBJECT_HOUR_2
 DEF MAPOBJECT_TIMEOFDAY        rb ; 7
 DEF MAPOBJECT_PALETTE          rb ; 8
 rsset MAPOBJECT_PALETTE
 DEF MAPOBJECT_TYPE             rb ; 8
 DEF MAPOBJECT_SIGHT_RANGE      rb ; 9
 DEF MAPOBJECT_SCRIPT_POINTER   rw ; a
 DEF MAPOBJECT_EVENT_FLAG       rw ; c
+DEF OBJECT_EVENT_SIZE EQU _RS - 1 ; object_event's begin with MAPOBJECT_SPRITE
                                rb_skip 2
 DEF MAPOBJECT_LENGTH EQU _RS
```

EDIT: To re-iterate the importance of this, you can't reorder/add/remove items from the `object_event` macro without also modifying the `map_object` struct.

vulcandth:
Opening this PR to further demonstrate what I'm suggesting should be done in issue #1031. Please review and look it over. If you don't think this is a good idea, or feel we want to have more conversation before this; feel free to close this PR or draft it.

Resolves #1031

### DIFF
--- a/constants/map_object_constants.asm
+++ b/constants/map_object_constants.asm
@@ -97,6 +97,7 @@ DEF RELATIVE_ATTRIBUTES EQU 1 << RELATIVE_ATTRIBUTES_F
 DEF ABSOLUTE_TILE_ID    EQU 1 << ABSOLUTE_TILE_ID_F
 
 ; map_object struct members (see macros/ram.asm)
+; struct initialized by object_event macro (see macros/scripts/maps.asm)
 rsreset
 DEF MAPOBJECT_OBJECT_STRUCT_ID rb ; 0
 DEF MAPOBJECT_SPRITE           rb ; 1
--- a/constants/script_constants.asm
+++ b/constants/script_constants.asm
@@ -112,7 +112,8 @@ DEF CALLBACK_SIZE     EQU  3 ; callback
 DEF WARP_EVENT_SIZE   EQU  5 ; warp_event
 DEF COORD_EVENT_SIZE  EQU  8 ; coord_event
 DEF BG_EVENT_SIZE     EQU  5 ; bg_event
-DEF OBJECT_EVENT_SIZE EQU 13 ; object_event
+; An object_event is a map_object without its initial MAPOBJECT_OBJECT_STRUCT_ID or final padding
+DEF OBJECT_EVENT_SIZE EQU MAPOBJECT_LENGTH - 3 ; 13
 
 ; A coord_event for scene -1 will always activate,
 ; regardless of the map's scene variable value.


mid-kid:
If you're moving `OBJECT_EVENT_SIZE` to be more closely related to `MAPOBJECT_*`, I'd name it something more apt like `MAPOBJECT_STRUCT_LENGTH` (we don't have a standard naming scheme for struct sizes...), and move all the other script command macros to their respective structure definitions, too, creating them where missing.

constants/map_object_constants.asm
```diff
@@ -114,6 +115,7 @@ DEF MAPOBJECT_TYPE             rb ; 8
 DEF MAPOBJECT_SIGHT_RANGE      rb ; 9
 DEF MAPOBJECT_SCRIPT_POINTER   rw ; a
 DEF MAPOBJECT_EVENT_FLAG       rw ; c
+DEF OBJECT_EVENT_SIZE EQU _RS - 1 ; object_event's begin with MAPOBJECT_SPRITE
```
Rangi42, line 118:
```suggestion
DEF OBJECT_EVENT_SIZE EQU _RS - 1 ; skip MAPOBJECT_OBJECT_STRUCT_ID
```

Rangi42 (changes requested):
Hmm. I liked keeping all the `*OBJECT_LENGTH` constants together. Although `object_event` fields do happen to correspond directly to a subset of `map_object` fields. How about changing script_constants.asm:

```
; An object_event is a map_object without its leading MAPOBJECT_OBJECT_STRUCT_ID
DEF OBJECT_EVENT_SIZE EQU MAPOBJECT_LENGTH - 1 ; 13
```

Rangi42:
This is also an opportunity to define field constants for all of the warp/coord/bg/object event macros, instead of hard-coding their total sizes.

vulcandth:
> Hmm. I liked keeping all the `*OBJECT_LENGTH` constants together. Although `object_event` fields do happen to correspond directly to a subset of `map_object` fields. How about changing script_constants.asm:
> 
> ```
> ; An object_event is a map_object without its leading MAPOBJECT_OBJECT_STRUCT_ID
> DEF OBJECT_EVENT_SIZE EQU MAPOBJECT_LENGTH - 1 ; 13
> ```

Unfortunately It is without the leading `MAPOBJECT_OBJECT_STRUCT_ID` and the last two unused bytes at the end of the struct. (See the `rb_skip 2`).

So it would have to be something like:
```
DEF OBJECT_EVENT_SIZE EQU MAPOBJECT_LENGTH - 3 ; 13
```

Rangi42:
I'm okay with this:
```
; An object_event is a map_object without its leading MAPOBJECT_OBJECT_STRUCT_ID or final padding
DEF OBJECT_EVENT_SIZE EQU MAPOBJECT_LENGTH - 3 ; 13
```

mid-kid (approved):

Rangi42 (approved):

PR #1058, opened 2023-07-19
Rename BATTLETYPE_SHINY to BATTLETYPE_FORCESHINY

SnorlaxMonster:
The name BATTLETYPE_SHINY gave the impression that it applied to all battles with Shiny Pokémon, rather than merely battles with forced Shiny Pokémon (i.e. Red Gyarados only).

Due to recent discussions, this misconception about the meaning of BATTLETYPE_SHINY came up a few times. See https://bulbapedia.bulbagarden.net/wiki/Talk:Escape#How_.28not.29_to_flee for one example.


Rangi42:
Hmm. This is the first time I've ever heard of such a confusion, but on the other hand, `BATTLETYPE_FORCESHINY` would be consistent with `BATTLETYPE_FORCEITEM`.

PR #1072, opened 2023-08-16
Fix RTC overflow comment

### LINKED ISSUE #1059: Bad comment in RTC-checking code
https://github.com/pret/pokecrystal/blob/1d64befdb7caf0db43f30810fb19fe5b1fba7786/engine/menus/main_menu.asm#L284

This code here checks for the RTC overflow bit, which gets set when the RTC day count exceeds _512_. However, since it is bit 6 of the control register, which also contains the upper portion of the day count, someone accidentally documented it as an actual check for $4000. (In actuality, bits 1-5 of that register are unmapped.)

This should just say it is checking whether the RTC overflow bit is set.

aaaaaa123456789:
Fixes #1059

### DIFF
--- a/engine/menus/main_menu.asm
+++ b/engine/menus/main_menu.asm
@@ -281,7 +281,7 @@ MainMenu_PrintCurrentTimeAndDay:
 
 .PlaceBox:
 	call CheckRTCStatus
-	and %10000000 ; Day count exceeded 16383
+	and %10000000 ; RTC overflow bit set (carry over from day count bit 8)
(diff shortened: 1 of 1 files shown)


aaaaaa123456789:
Sure, why not.

Idain:
And so it was never resolved.

Rangi42:
This will be resolved with a real constant when https://github.com/pret/pokegold/pull/120 is ported to pokecrystal.

Issue #1032, opened 2023-01-17
Further Object Renaming consideration


vulcandth:
I know at this point we are probably tired of renaming these things... but I've always found their prefixes rather confusing... It is easy to get the two mixed up.

Reminder of how these work: Each `object_event` macro in the map header is loaded into the `map_object` struct. When the object is visible on the screen, meets time of day requirements, and a few other checks, it is copied from the `map_object` struct into the `object_struct`. When objects are no longer visible (unless a don't delete flag is specified), they are deleted from `object_struct`.

My recommendation is to rename the `object_struct` to something like `active_object_struct`. There are a few other recommendations in regards to the macro name's, associated labels, and constants. See the original names on the left, and the recommended changes on the right.

| Original Labels | Recommended Labels |
|------| --------|
|`object_struct` | `active_object_struct` |
|`wObjectStructs` | `wActiveObjects`|
|`wPlayerStruct` / `wPlayerSprite` | `wPlayerActiveObject`/`wPlayerActiveObjectSprite` |
|`wObject#Struct` / `wObject#Sprite` | `wActiveObject#`/`wActiveObject#Sprite` |
|`OBJECT_LENGTH` | `ACTIVE_OBJECT_LENGTH` |

| Original Labels | Recommended Labels |
|------| --------|
|`map_object` | `map_object_struct` |
| `wMapObjects` | (Existing name is good.) |
| `wPlayerObject`/`wPlayerObjectSprite` | `wPlayerMapObject`/`wPlayerMapObjectSprite` |
| `wMap#Object`/`wMap#ObjectSprite` | `wMapObject#` / `wMapObject#Sprite` |
| `MAPOBJECT_LENGTH` | `MAP_OBJECT_LENGTH` |

I am just opening this issue simply start a conversation... I too would like to avoid the unnecessary renaming of things if we are not all in agreeance. Thanks!

Rangi42:
Followup to #655.

vulcandth:
@mid-kid 
https://github.com/pret/pokecrystal/issues/639#issuecomment-522173584
>`wMapObjects` is still left over from when we renamed "map object"s to "object event"s. I vaguely recall having talked about this issue before, and decided there wasn't a good solution, but I don't remember why.
Anyway, anything called MapObject should probably be renamed to ObjectEvent, unless someone reminds me of a reason not to.

The label `ObjectEvent` is currently already used here:

```asm
ObjectEvent::
	jumptextfaceplayer ObjectEventText

[...]

Issue #639, opened 2019-06-28
Misc things to rename.


mid-kid:
Map setup command constants don't match their function counterpart.

See `constants/map_setup_constants.asm` and `engine/overworld/map_setup.asm`. For example `map_face_down` vs `SpawnInFacingDown`.

While it is true that some of these functions are used for more than just map setup scripts, we should make sure the names match to the best of our ability, to make them easier to find.

mid-kid:
As an extension, `RestoreFacingAfterWarp` doesn't seem to be doing what the name implies.

Rangi42:
Taking this as a "misc things to rename" issue:

It's only clear that `TILESET_BATTLE_TOWER` is for the indoor maps because `BATTLE_TOWER_OUTSIDE` exists. Can it be renamed to `TILESET_BATTLE_TOWER_INSIDE`?

mid-kid:
There's discrepancies between `object_struct` as defined in `constants/map_object_constants.asm` and `macros/wram.asm`, for example, `OBJECT_FACING` maps to `\1Direction`, where `OBJECT_FACING_STEP` maps to `\1Facing`.

mid-kid:
The length of the `map_object` struct is `OBJECT_LENGTH` while the length of the `object_struct` struct is `OBJECT_STRUCT_LENGTH`. I propose to fix this to be `MAPOBJECT_LENGTH` and `OBJECT_LENGTH`, to better correspond with their other constant names.

Rangi42:
Not many people mess with the sprite system internals, so the change in meaning of `OBJECT_LENGTH` should be okay.

mid-kid:
The `buttonsound` and `waitbutton` commands are related to eachother: They're both used to wait for the player to press a button to end a textbox. However, the former places the blinking "▼" at the bottom-right corner of the textbox, and plays a sound. I don't think `buttonsound` is a good name, but I don't know of any good alternatives right now.

Rangi42:
Maybe `waitbuttonsfx`/`WaitButtonSfx`? Or `promptwaitbutton`/`PromptWaitButton`, as a parallel to text `prompt` which also shows a cursor.

Rangi42:
gfx/intro/\*.tilemap need renaming to go with their tileset graphics. Also 011.tilemap is actually the attrmap corresponding to 012.tilemap.

Rangi42:
[...]

PR #1076, opened 2023-08-22
Map/object/event renaming

vulcandth:
Resolves: #1032 #655 

This is what I ended up going with:  
  
| Original Labels | Recommended Labels |  
|------| --------|  
|\`object\_struct\` | (current name is fine) |  
|\`wObjectStructs\` | (current name is fine) |  
|\`wPlayerStruct\` / \`wPlayerSprite\` | `wPlayerObjectS`truct/\`wPlayerSprite\` |  
|\`wObject#Struct\` / \`wObject#Sprite\` | `wObjectS`truct#/\`wObject#Sprite\` |  
|\`NUM\_OBJECT\_STRUCTS\` | (current name is fine) |  
|\`OBJECT\_LENGTH\` | (current name is fine) |

| Original Labels | Recommended Labels |  
|------| --------|  
|\`map\_object\` | \`object\_event\_struct\` |  
| \`wMapObjects\` | `wO`bjectEventStructs |  
| \`wPlayerObject\`/\`wPlayerObjectSprite\` | `wPlayerObjectEventS`truct/\`wPlayerObjectEventSprite\` |  
| \`wMap#Object\`/\`wMap#ObjectSprite\` | \`wObjectEventStruct#\` / \`wObjectEvent#Sprite\` |

|\`NUM\_OBJECTS\` | `NUM_OBJECT_EVENT_STRUCTS` |  
| \`MAPOBJECT\_LENGTH\` | \`OBJECT\_EVENT\_LENGTH\` |  
| `MAPOBJECT_*` | `OBJECT_EVENT_*` |  
  
Anywho; this is just to keep the conversation going; I don't mind reworking this if we decide to do something different.

### DIFF
--- a/constants/map_object_constants.asm
+++ b/constants/map_object_constants.asm
@@ -1,39 +1,39 @@
 ; object_struct members (see macros/ram.asm)
 rsreset
-DEF OBJECT_SPRITE           rb ; 00
-DEF OBJECT_MAP_OBJECT_INDEX rb ; 01
-DEF OBJECT_SPRITE_TILE      rb ; 02
-DEF OBJECT_MOVEMENT_TYPE    rb ; 03
-DEF OBJECT_FLAGS1           rb ; 04
-DEF OBJECT_FLAGS2           rb ; 05
-DEF OBJECT_PALETTE          rb ; 06
-DEF OBJECT_WALKING          rb ; 07
-DEF OBJECT_DIRECTION        rb ; 08
-DEF OBJECT_STEP_TYPE        rb ; 09
-DEF OBJECT_STEP_DURATION    rb ; 0a
-DEF OBJECT_ACTION           rb ; 0b
-DEF OBJECT_STEP_FRAME       rb ; 0c
-DEF OBJECT_FACING           rb ; 0d
-DEF OBJECT_TILE             rb ; 0e
-DEF OBJECT_LAST_TILE        rb ; 0f
-DEF OBJECT_MAP_X            rb ; 10
-DEF OBJECT_MAP_Y            rb ; 11
-DEF OBJECT_LAST_MAP_X       rb ; 12
-DEF OBJECT_LAST_MAP_Y       rb ; 13
-DEF OBJECT_INIT_X           rb ; 14
-DEF OBJECT_INIT_Y           rb ; 15
-DEF OBJECT_RADIUS           rb ; 16
-DEF OBJECT_SPRITE_X         rb ; 17
-DEF OBJECT_SPRITE_Y         rb ; 18
-DEF OBJECT_SPRITE_X_OFFSET  rb ; 19
-DEF OBJECT_SPRITE_Y_OFFSET  rb ; 1a
-DEF OBJECT_MOVEMENT_INDEX   rb ; 1b
-DEF OBJECT_STEP_INDEX       rb ; 1c
-DEF OBJECT_1D               rb ; 1d
-DEF OBJECT_1E               rb ; 1e
-DEF OBJECT_JUMP_HEIGHT      rb ; 1f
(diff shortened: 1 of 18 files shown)


mid-kid:
Reflecting on past changed, I'm not sure overloading "object event" to mean anything other than "a map object's event" was a good idea. However, we went with it, and I prefer consistency over correctness, so these changes are overall fine.

One thing I'll request though is to reserve the term "struct" for the macro names, e.g. `wObjects` contains a bunch of `object_struct`s, `wObjectEvents` contains `object_event_struct`, etc.

mid-kid:
wObjects isn't related to maps specifically, it's related to the sprite animation subsystem, so it wouldn't make sense to call it wMapObjects, anyway

Rangi42:
I don't see any *errors* here. There are some quirks like now having a field named `OBJECT_OBJECT_EVENT_STRUCT`, or the easily confusible `object_event` and `object_event_struct` macros. I don't know if the greater consistency here is worth that. I'll defer to @mid-kid since you motivated this in #639 anyway.

mid-kid:
I'll be honest, I'm completely out of this effort, and don't fully remember what I intended, so take my ramblings with a grain of salt.
As for object_event and object_event_struct, those still make sense even if they don't look "right" at first glance. One is in rom, the other is in ram, and both directly reflect eachother, they're the same.
I do agree with the struct field constants looking awkward now, but that's an artifact of the map_object->object_event rename. IMO it wasn't the most well thought-out move, but it only makes sense to finish what we started instead of leaving it in this haphazard state. For now I guess we can go with `OBJECT_EVENT_INDEX` instead of `OBJECT_OBJECT_EVENT_INDEX`. I'd have to read the code of the sprite subsystem again, but if there's any trend towards calling these "Sprites" instead (like we have `UpdateSprites`), I'd consider renaming the object_struct to sprite_struct.

vulcandth:
I don't _**personally**_ like calling them "Sprites"... as the game often adds sprites to the screen that don't use these structs. (Pokemon Fly leaves are an example.)... Furthermore; the battle Anim system, party menu sprites, ect.. don't use these structs... only the overworld. However, i'm not going to make a big fuss about it.

We **_could_** do `ow_object_struct` and `OW_OBJECT_EVENT_INDEX`. where `ow` stands for Overworld.

I'm just pitching ideas. Think on it.

mid-kid:
> the game often adds sprites to the screen that don't use these structs

The object structs are the ones that keep the info for the OAMs, are they not?

vulcandth:
> > the game often adds sprites to the screen that don't use these structs
> 
> The object structs are the ones that keep the info for the OAMs, are they not? And they're used by menus and the title screen, i.e. anything that isn't battle anims, iirc.

I know for a fact that the fly leaf animations write directly to wShadowOAM bypassing the object structs.. Let me double check the other menus.

vulcandth:
Yeah; the party menu uses the `sprite_anim_struct`. IT doesn't use the `object_struct` or `map_object`/`object_event_struct`.

mid-kid:
oh, I see. In that case yeah, object may be renamed to something more specific.

mid-kid:
If you shorten a term, like "overworld_obj", be consistent about it, so "OverworldObjs" and "OVERWORLD_OBJ_LENGTH". In this case I'd just spell it out.

As for the "overworld event" term, it sounds awkward and might be confused for tile events which is annoying. I like the idea of going back on calling them "object events" and reserving that term explicitly for the static version of the struct in a map script, but we'd have to make sure the usage of this term is consistent. `*MapObject*` -> `*ObjectEvent*` was initially proposed in order to align to the new naming in map scripts, and some things were renamed accordingly, which would have to be renamed back.

PR #1077, opened 2023-08-22
CI copy `.map` files to symbols branch

vulcandth:
~~Omit symbols from `*.map` files; They are really so much easier to read after this is done.~~

I also thought it would be nice to have the `.map` files on the `symbols` branch as well. If you disagree, then I can easily remove that commit.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -66,11 +66,12 @@ jobs:
         if: ${{ github.event_name == 'push' && github.repository_owner == 'pret' }}
         run: |
           cp -v *.sym symbols/
+          cp -v *.map symbols/
 
       - name: Update symbols
         if: ${{ github.event_name == 'push' && github.repository_owner == 'pret' }}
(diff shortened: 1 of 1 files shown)


mid-kid:
I use this to figure out in what section a symbol is located. While that's not as useful in pokecrystal right now, there's some places where it's a godsent...

vulcandth:
Where-as I almost never need to know which section a symbol is, and if I do I can just figure that out from using the sym and map files... I primarily use the map files to find empty space in banks only.

However that being said... Let's just keep the symbols in the file like before. 



Side note: if you like the other commit of moving the .map files to the symbols branch in the CI.. we could still do that in this PR. Otherwise just close this PR and the related Issue please.

mid-kid:
I tend to use `./tools/free_space.awk BANK=all pokecrystal.map` for that.

mid-kid:
That said, I like the idea of adding the .map file to the symbols branch.

PR #1094, opened 2023-11-06
Clarify and correct usage of breakable whitespace characters

SnorlaxMonster:
In [charmap.asm](https://github.com/pret/pokecrystal/blob/master/charmap.asm), $1f (¯) is described as a "soft linebreak", while $25 (%) is described as "soft linebreak in landmark names". I don't think these are particularly good descriptions of what these codepoints represent.

Firstly, it is $1f that is used [in landmark names](https://github.com/pret/pokegold/blob/master/data/maps/landmarks.asm), not $25 ($25 appears to be entirely unused).

As for what these codepoints actually do: When [displayed on the Town Map](https://github.com/pret/pokecrystal/blob/master/engine/pokegear/townmap_convertlinebreakcharacters.asm), they render as line breaks. [Everywhere else](https://github.com/pret/pokecrystal/blob/master/home/text.asm#L238), the two render differently: $1f renders a space, whereas $25 renders nothing. In light of this, I think a more accurate description is that $1f is a breakable space (as opposed to the standard space character $7f, which is effectively an unbreakable space) and $25 is a [zero-width space](https://en.wikipedia.org/wiki/Zero-width_space).

Also, I question the choice of symbols used to represent these two characters. I don't think a macron or percent sign intuitively represents an optional line break, especially given how rarely these two codepoints are actually used. I think just angle-bracketed terms (e.g. `<BSP>` for $1f and `<ZWSP>` for $25) like almost every other control character would be clearer.

Other than landmark names, macrons are used within Japanese strings in comments and untranslated mobile data. Both of these usages appear to be in error. In untranslated mobile data, they are used in various strings, but these usages actually should be `<WO>` (the Japanese control character at $25) instead of the English control character at this code point.

Macrons are used in comments in [moves/grammar.asm](https://github.com/pret/pokecrystal/blob/master/data/moves/grammar.asm), in reference to move usage dialogue (particularly Japanese dialogue). However, in the Japanese version these appear to actually be "hard" linebreaks ($22) or "を　" instead. (It's possible the latter are `<WO>` internally, but in this instance I don't think the internal representation is as important as the display, since they are used purely in comments.) Notably, in the Japanese version these strings use 2-tile linebreaks (`<LINE>`), whereas the two breakable whitespace characters become 1-tile linebreaks (`<LF>`) on the Town Map in the English version.

In light of these issues, I have made the following changes:
- Clarified the comments for $1f and $25 in charmap.asm
- Renamed $1f and $25 to `<BSP>` and `<ZWSP>`, instead of ¯ and %, respectively
- Updated all usages of $1f and $25 to use the new names
- Corrected untranslated Japanese strings to use `<WO>` instead of ¯ for code point $1f
- Corrected Japanese strings in comments in grammar.asm to align with what is actually displayed in the Japanese version


charmap.asm
```diff
@@ -34,7 +34,7 @@
 	charmap "<TM>",      $5c ; "TM"
 	charmap "<TRAINER>", $5d ; "TRAINER"
 	charmap "<ROCKET>",  $5e ; "ROCKET"
-	charmap "<DEXEND>",  $5f
+	charmap "<DEXEND>",  $5f ; unused
```
Rangi42, line 37: Other chars are unused too, like `"<TRAINER>"` and `"<ROCKET>"`; I don't think they all need such comments.
```suggestion
	charmap "<DEXEND>",  $5f
```

charmap.asm
```diff
@@ -6,10 +6,10 @@
 	charmap "<PLAY_G>",  $14 ; "<PLAYER>くん" or "<PLAYER>ちゃん"; same as "<PLAYER>" in English
 	charmap "<MOBILE>",  $15
 	charmap "<CR>",      $16
-	charmap "¯",         $1f ; soft linebreak
+	charmap "<BSP>",     $1f ; breakable space (usually " ", but renders as line break on the Town Map)
```
Rangi42, line 9:
```suggestion
	charmap "<BSP>",     $1f ; breakable space (usually " ", or "<LF>" on the Town Map)
```

charmap.asm
```diff
@@ -6,10 +6,10 @@
...
 	charmap "<LF>",      $22
 	charmap "<POKE>",    $24 ; "<PO><KE>"
-	charmap "%",         $25 ; soft linebreak in landmark names
+	charmap "<ZWSP>",    $25 ; zero-width breakable space (usually not displayed, but renders as line break on the Town Map), unused
```
Rangi42, line 12: I think `WBR` is more accurate and matches the HTML `<wbr>` tag.
```suggestion
	charmap "<WBR>",     $25 ; word-break opportunity (usually skipped, or "<LF>" on the Town Map)
```

Rangi42 (changes requested):
Mostly good! I'd rename ZWSP to WBR. Also macros/legacy.asm needs to support the previous Unicode charmaps.

charmap.asm
SnorlaxMonster: No specific issue with removing that comment, although I added that because most of the characters under "Actual characters" specifically list when they are unused. Regardless, you're right that it makes more sense to either apply it to all of the control characters or none of them, so that would belong in a separate pull request.

charmap.asm
SnorlaxMonster: `<WBR>` is [equivalent to](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/wbr) U+200B ZERO-WIDTH SPACE (which is commonly shortened to ZWSP), so they mean the same thing. However, given it is being rendered in angle-brackets, using the HTML element does make more sense (as people are more likely to be familiar with that name), and I think "word-break opportunity" is a more self-explanatory name. Happy to apply that change.

SnorlaxMonster:
I've applied those suggested changes, updated the other locations that reference `<ZWSP>` to reference `<WBR>` instead, and added % and ¯ to macros/legacy.asm

home/text.asm
```diff
@@ -235,8 +235,8 @@ ENDM
 	dict "<PROMPT>",  PromptText
 	dict "<PKMN>",    PlacePKMN
 	dict "<POKE>",    PlacePOKE
-	dict "%",         NextChar
-	dict "¯",         " "
+	dict "<WBR>",    NextChar
```
Rangi42, line 238:
```suggestion
	dict "<WBR>",     NextChar
```

Rangi42:
Thanks again! Just merge master to this, because charmap.asm was moved to constants/charmap.asm.

SnorlaxMonster:
I've merged master into my feature branch, so there are no longer conflicts.

Rangi42 (approved):

PR #1100, opened 2023-12-24
Fix documentation about Magikarp length calculation and bugfix about "Magikarp lengths can be miscalculated"

Idain:
This pull request tackles two documentation topics about Magikarp's length:

1. The maximum length of a Magikarp when judged by the Guru Fisher in Lake of Rage.
2. A badly documented fix in "Magikarp lengths can be miscalculated".

## 1. Wrong comment about maximum Magikarp's length in `CalcMagikarpLength`

First of all, let's talk about `CalcMagikarpLength`. This function first calculates the length in millimeters by using the Magikarp's DVs and the Trainer ID, and then convert it to feet and inches. The maximum length a Magikarp can have (in millimeters) is 1625mm (bugged version)/1755mm (fixed version), but the comment inside the function says the maximum is 1786mm ([the bug](https://github.com/pret/pokecrystal/blob/master/docs/bugs_and_glitches.md#magikarp-lengths-can-be-miscalculated) makes it so that only `b` is used when calculating the length, instead of `bc`). To check this, I've run various calculations, both accounting for the bug in `CalcMagikarpLength.BCLessThanDE` and with the fixed version.

To prove this, let's assume `bc` has the maximum possible value (`$ffff`). According to `CalcMagikarpLength`, if `bc ≥ $ff00: [wMagikarpLength] = c + 1370`, whose code is located [here](https://github.com/pret/pokecrystal/blob/9a917e35760210a1f34057ecada2148f1fefc390/engine/events/magikarp.asm#L228C2-L228C2), so let's start calculating:

```
; If bc = $ffff, then c = $ff = 255
length = c + 1370
length = 255 + 1370
length = 1625
```

Q: "According to the comment inside `CalcMagikarpLength`, the formula is different if `bc < $ff00`, so maybe the result could be 1786mm?"

A: I also accounted for this, and no, the max possible result is either 1590mm (with the bug) or 1755mm (fixed version). If you fix the function, the last threshold in the `MagikarpLength` table is used, which is 65510. Let's make two calculations, one for each version (the formula is `z * 100 + (bc - x) / y`):

### Bugged version

```
; The highest threshold is x = 65410, where y = 2 and z = 14. We'll use bc = $ff00 - $1 = 65279

length = z * 100 + (bc - x) / y
length = 14 * 100 + (65279 - 65410) / 2
length = 1400 + (65405) / 2  ; The result is stored in bc and becomes -131 = 65405
length = 1400 + 190  ; The result is $7FBE = 32702, but the code only takes the low byte, which is $BE = 190
length = 1590
```

### Fixed version

```
; With the code fixed, the highest threshold is x = 65510, where y = 1 and z = 15. We'll use bc = 65509

length = z * 100 + (bc - x) / y
length = 15 * 100 + (65509 - 65510) / 1
length = 1500 + (65535) / 1  ; The result is stored in bc and becomes -1 = 65535
length = 1500 + 255  ; The code only takes the low byte of the result, which is $FF = 255
length = 1755
```

## 2. Wrong fix in "Magikarp lengths can be miscalculated"

The [bugfix](https://github.com/pret/pokecrystal/blob/master/docs/bugs_and_glitches.md#magikarp-lengths-can-be-miscalculated) doesn't account for cases where `b > d`, so instead of returning, it tries to compare the low bytes `bc` and `de` as if the high bytes were equal, which will give the wrong result.

### DIFF
--- a/docs/bugs_and_glitches.md
+++ b/docs/bugs_and_glitches.md
@@ -2416,6 +2416,7 @@ CopyPokemonName_Buffer1_Buffer3:
  	cp d
  	ret c
 -	ret nc
++	ret nz
  	ld a, c
  	cp e
  	ret
--- a/engine/events/magikarp.asm
+++ b/engine/events/magikarp.asm
@@ -110,7 +110,7 @@ CalcMagikarpLength:
 
 ; This function is poorly commented.
 
-; In short, it generates a value between 190 and 1786 using
+; In short, it generates a value between 190 and 1625 using
 ; a Magikarp's DVs and its trainer ID. This value is further
 ; filtered in LoadEnemyMon to make longer Magikarp even rarer.
 


engine/events/magikarp.asm
```diff
@@ -110,12 +110,13 @@ CalcMagikarpLength:
 
 ; This function is poorly commented.
 
-; In short, it generates a value between 190 and 1786 using
+; In short, it generates a value between 190 and 1755 using
 ; a Magikarp's DVs and its trainer ID. This value is further
 ; filtered in LoadEnemyMon to make longer Magikarp even rarer.
+; Due to the bug at .BCLessThanDE, the max possible value is 1625.
```
Rangi42, line 116: The bugfix should diff removing this.

Rangi42:
Nicely explained, thank you!

engine/events/magikarp.asm
Idain: Nice catch!

Idain:
@Rangi42, done. Let me know if there's something else to address.

mid-kid:
While we're at it, I'd add a `; BUG:` marker to this, replacing some redundant bits of explanation if possible.

Idain:
@mid-kid, how is it now?

docs/bugs_and_glitches.md
```diff
@@ -2407,15 +2407,20 @@ CopyPokemonName_Buffer1_Buffer3:
 
 ### Magikarp lengths can be miscalculated
 
+The function `CalcMagikarpLength` is supposed to determine a Magikarp's length by using the Trainer's ID and the Magikarp's DVs, `xor`ing the values, store the result in `bc`, use it as an index to select an entry from the `MagikarpLengths` table and calculate the Magikarp's length, but it doesn't work as intended. The local function `.BCLessThanDE` is supposed to compare the index (stored in `bc`) with a threshold (stored in `de`) from the table to select the correct entry, but it only compares the high bytes of both inputs, which means that the wrong entry might be selected.
+
+This bug also causes the first entry of the `MagikarpLengths` table to go unused and `bc` values between 65280 and 65509 to use the wrong formula. As a side effect, the highest possible length is capped at 1625mm (before being converted to feet and inches).
```
Rangi42, line 2412:
```suggestion
`CalcMagikarpLength.BCLessThanDE` only compares the high bytes of `bc` and `de`. This causes the first entry of the `MagikarpLengths` table to go unused, and `bc` values between 65,280 and 65,509 to use the wrong formula. As a side effect, the highest possible length is capped at 1,625mm (before being converted to feet and inches).
```

engine/events/magikarp.asm
```diff
@@ -108,35 +108,34 @@ CalcMagikarpLength:
...
+; if bc = 2710-7709:    x =  7710,  y =  50,  z =  6
+; if bc = 7710-17709:   x = 17710,  y = 100,  z =  7
+; if bc = 17710-32709:  x = 32710,  y = 150,  z =  8
+; if bc = 32710-47709:  x = 47710,  y = 150,  z =  9
+; if bc = 47710-57709:  x = 57710,  y = 100,  z = 10
+; if bc = 57710-62709:  x = 62710,  y =  50,  z = 11
+; if bc = 62710-64709:  x = 64710,  y =  20,  z = 12
+; if bc = 64710-65209:  x = 65210,  y =   5,  z = 13
+; if bc = 65210-65409:  x = 65410,  y =   2,  z = 14
+; if bc = 65410-65509:  x = 65510,  y =   1,  z = 15
```
Rangi42, line 138: I'm questioning whether this entire list needs to be here at all. Especially if fixing the bug would change any of it.

docs/bugs_and_glitches.md
Idain: I admit I can be very verbose at times. Ahahahaha...

engine/events/magikarp.asm
Idain: I'm indifferent to it, but I can delete it if you want to.

engine/events/magikarp.asm
```diff
@@ -108,36 +108,18 @@ CalcMagikarpLength:
 ;   de: wEnemyMonDVs
 ;   bc: wPlayerID
```
SnorlaxMonster, line 109:
```suggestion
;   de: Magikarp's DVs (MON_DVS / wEnemyMonDVs )
;   bc: Trainer ID (MON_ID / wPlayerID)
```
Those are only the inputs when called from `LoadEnemyMon` in [engine/battle/core.asm](https://github.com/pret/pokecrystal/blob/master/engine/battle/core.asm). When called from `CheckMagikarpLength`, the inputs are `MON_DVS` and `MON_ID`. (This distinction is important, as `MON_ID` is the Original Trainer's ID, not the player's.)

engine/events/magikarp.asm
```diff
@@ -108,36 +108,18 @@ CalcMagikarpLength:
...
 
-; This function is poorly commented.
-
-; In short, it generates a value between 190 and 1786 using
-; a Magikarp's DVs and its trainer ID. This value is further
-; filtered in LoadEnemyMon to make longer Magikarp even rarer.
+; It generates a value between 190 and 1755 using a Magikarp's DVs
+; and its trainer ID. This value is further filtered in LoadEnemyMon
+; to make longer Magikarp even rarer.
```
SnorlaxMonster, line 113:
```suggestion
; It generates a length (in mm) between 190 and 1755 using a Magikarp's DVs
; and its Trainer ID.
```
The comment about being further filtered in `LoadEnemyMon` assumes this function is only called from `LoadEnemyMon` (ignoring the case when it is called from `CheckMagikarpLength`). IMO, what a calling function does with the return value is out of the scope of what belongs in comments on this function.

engine/events/magikarp.asm
```diff
@@ -108,36 +108,18 @@ CalcMagikarpLength:
...
 
 ; The value is generated from a lookup table.
-; The index is determined by the dv xored with the player's trainer id.
-
-; bc = rrc(dv[0]) ++ rrc(dv[1]) ^ rrc(id)
+; The index is determined by the DV xored with the player's trainer ID
+; and then stored in bc.
```
SnorlaxMonster, line 117:
```suggestion
; The index is determined by the DV xored with the Magikarp's Trainer ID
; and then stored in bc.
```
The player's Trainer ID is only used when called from `LoadEnemyMon`. When this function is called from `CheckMagikarpLength`, it uses Magikarp's Original Trainer's Trainer ID, not the player's. Even though the Magikarp technically doesn't have a Trainer ID when it is generated in the wild, I think it's clear enough what is being referred to if this is just generalized as "the Magikarp's Trainer ID".

engine/events/magikarp.asm
```diff
@@ -108,36 +108,18 @@ CalcMagikarpLength:
...
-; This function is poorly commented.
-
-; In short, it generates a value between 190 and 1786 using
-; a Magikarp's DVs and its trainer ID. This value is further
-; filtered in LoadEnemyMon to make longer Magikarp even rarer.
+; It generates a value between 190 and 1755 using a Magikarp's DVs
+; and its trainer ID. This value is further filtered in LoadEnemyMon
+; to make longer Magikarp even rarer.
 
 ; The value is generated from a lookup table.
```
SnorlaxMonster, line 115:
```suggestion
; The length is generated from a lookup table.
```

SnorlaxMonster (changes requested):
`CalcMagikarpLength` is called both when the Fisherman is evaluating the length of your Magikarp (from `CheckMagikarpLength`), and when encountering a Magikarp in the wild (from `LoadEnemyMon`). However, the comments in various parts assume it is being called from a specific one of those two places. This is important because when the Fisherman evaluates a Magikarp, he uses the Magikarp's Original Trainer's ID number. But when `LoadEnemyMon` is calculating the length of a wild Magikarp, it uses the player's Trainer ID (since the wild Magikarp doesn't have a Trainer ID yet).

I was actually planning to submit a PR to address these issues, but noticed there was already a PR adjusting the comments on the same function, so figured it would make more sense to amend this one.

engine/events/magikarp.asm
Idain: I see. I'll address your suggestions when I get my laptop back.

engine/events/magikarp.asm
Idain: In this case it can feel misleading if we don't also clarify it uses the Player's ID when fighting against a wild Magikarp.

SnorlaxMonster: That's completely reasonable. Perhaps something like this would be better
```suggestion
; The index is determined by the DV xored with a Trainer ID 
; (the player's for wild Magikarp; the OT's when checking owned Magikarp)
; then stored in bc.
```

Idain: It sounds better.

engine/events/magikarp.asm
vulcandth: This would be `MON_OT_ID` now.

vulcandth (changes requested):

PR #1101, opened 2024-01-01
Update to RGBDS 0.7.0

Rangi42:
These new features are used from RGBDS 0.7.0:

- Use `SIZEOF` and `STARTOF` for section types, instead of `*RAM_Begin/End` constants
- Use `ENDSECTION` at the end of `INCLUDE`d files which start their own `SECTION`s
  - These can be found via:  
    ``for f in `git grep -Elw '^SECTION'`; do git grep -ho "INCLUDE \"$f\""; done``
- `OAMDMACodeEnd` can be `.End` since `ENDL` restores scope

Also rgbdscheck.asm is rewritten to be more robust with older versions of RGBDS. (We can't define constants since `DEF` is required in 0.7.0 but not yet defined in old enough versions; likewise for backslash line continuations, `ELIF`, etc.)

Note that the object file revision is still 9, so tools/unnamed.py did not need updating.

### DIFF
--- a/rgbdscheck.asm
+++ b/rgbdscheck.asm
@@ -1,16 +1,6 @@
-MAJOR EQU 0
-MINOR EQU 6
-PATCH EQU 0
-
-WRONG_RGBDS EQUS "fail \"pokecrystal requires rgbds v0.6.0 or newer.\""
-
 IF !DEF(__RGBDS_MAJOR__) || !DEF(__RGBDS_MINOR__) || !DEF(__RGBDS_PATCH__)
-	WRONG_RGBDS
-ELSE
-IF (__RGBDS_MAJOR__ < MAJOR) || \
-	(__RGBDS_MAJOR__ == MAJOR && __RGBDS_MINOR__ < MINOR) || \
-	(__RGBDS_MAJOR__ == MAJOR && __RGBDS_MINOR__ == MINOR && __RGBDS_PATCH__ < PATCH) || \
-	(__RGBDS_MAJOR__ == MAJOR && __RGBDS_MINOR__ == MINOR && __RGBDS_PATCH__ == PATCH && DEF(__RGBDS_RC__))
-	WRONG_RGBDS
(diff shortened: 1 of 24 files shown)


vulcandth:
Lets also update the STYLE.md guide to mention we won't be using `::` to separate multiple instructions on one line.

Rangi42:
STYLE.md was last updated two years ago and may need further changes than that; can we just do it separately?

vulcandth (approved):
Sure sounds good.

rgbdscheck.asm
```diff
@@ -1,16 +1,6 @@
...
-ELSE
-IF (__RGBDS_MAJOR__ < MAJOR) || \
-	(__RGBDS_MAJOR__ == MAJOR && __RGBDS_MINOR__ < MINOR) || \
-	(__RGBDS_MAJOR__ == MAJOR && __RGBDS_MINOR__ == MINOR && __RGBDS_PATCH__ < PATCH) || \
-	(__RGBDS_MAJOR__ == MAJOR && __RGBDS_MINOR__ == MINOR && __RGBDS_PATCH__ == PATCH && DEF(__RGBDS_RC__))
-	WRONG_RGBDS
+	fail "pokecrystal requires rgbds v0.7.0 or newer."
 ENDC
+IF __RGBDS_MAJOR__ == 0 && __RGBDS_MINOR__ <  7
+	fail "pokecrystal requires rgbds v0.7.0 or newer."
```
mid-kid, line 5: is fail available on older rgbds?

Rangi42: Yes, at least as old as 0.3.0, which is the oldest one to have rgbds.gbdev.io documentation.

aaaaaa123456789:
> Lets also update the STYLE.md guide to mention we won't be using `::` to separate multiple instructions on one line.

This sounds unnecessary. This is already something extremely unusual, to the point "don't do it" is the default in basically every programming community for every language — if someone wants to do this, there's probably a very good reason for it.

PR #1108, opened 2024-01-28
Add defintion for OAM y-Coordinate offscreen value

vulcandth:
Reference: https://gbdev.io/pandocs/OAM.html#byte-0--y-position

Essentially, a y-coord of 160 is the screen's height (144) plus the height of a 8x16 sprite (2 tiles).


mid-kid:
I'd honestly rather a new constant for this, the math isn't really obvious

vulcandth:
> I'd honestly rather a new constant for this, the math isn't really obvious (OAM\_YCOORD\_DISABLED or similar, please think of a better name than this)

Why come up with a better name when you have a perfectly terrible name right here. LOL

Rangi42:
`OAM_YCOORD_HIDDEN`? Since they're not disabled, just off-screen.

PR #1114, opened 2024-02-25
Rename `MON_ID` to `MON_OT_ID`

SnorlaxMonster:
The constant name MON_ID implies it relates to the ID number of the Pokémon itself; however, it actually refers to the Trainer ID number of the Pokémon's Original Trainer. Renaming it to MON_TRAINER_ID makes it substantially more clear what its meaning is.

The original definition of MON_ID is preserved in `macros/legacy.asm`

### DIFF
--- a/macros/legacy.asm
+++ b/macros/legacy.asm
@@ -406,9 +406,6 @@ DEF happinesschecknpc     EQUS "HappinessCheckScript"
 charmap "%", "<BSP>"
 charmap "¯", "<WBR>"
 
-; constants/pokemon_data_constants.asm
-DEF MON_ID EQUS "MON_TRAINER_ID"
-
 ; constants/sprite_constants.asm
 DEF SPRITE_BUENA EQUS "SPRITE_BEAUTY"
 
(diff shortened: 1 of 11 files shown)


macros/legacy.asm
```diff
@@ -406,6 +406,9 @@ DEF happinesschecknpc     EQUS "HappinessCheckScript"
 charmap "%", "<BSP>"
 charmap "¯", "<WBR>"
 
+; constants/pokemon_data_constants.asm
+DEF MON_ID EQUS "MON_TRAINER_ID"
+
```
Rangi42, line 411: Thanks, but we don't need legacy support for *every* change (we rename plenty of functions, structs, etc). It's focused on supporting old map-scripts or music, i.e. things people make without "programming assembly".
```suggestion
```

SnorlaxMonster: Ah, no worries. Last time I renamed a constant I was asked to add to to the legacy macros, so figured that was something that should regularly be done.

vulcandth:
Calling it `MON_OT_ID` would probably be more consistent with the rest of the codebase.

Rangi42 (approved):

Rangi42:
I agree `MON_OT_ID` would be slightly better, but it's fine as-is if you want to just merge.

vulcandth (approved):
I'll probably make the change to `MON_OT_ID` myself and i'll merge this at the same time as Idain's PR real soon. Thanks for the contribution!

mid-kid:
I agree on MON_OT_ID, since we use the OT term everywhere.

PR #1115, opened 2024-03-16
Add more assert cases in the codebase

Idain:


Rangi42:
Neat, want to do a couple dozen more? `git grep -E '(inc|dec) [abcdehl] ; [A-Z0-9_]+$'` :)

Also, similar `assert`s have gone *before* the relevant line, which here is the `inc a`. Examples:

```asm
	assert PokemonPicPointers == UnownPicPointers
	ld hl, PokemonPicPointers
```

```asm
	assert BUG_CONTEST_PLAYER == 1
	dec a
```

```asm
	assert LOW(SERIAL_LINK_BYTE_TIMEOUT) == 0
	xor a ; LOW(SERIAL_LINK_BYTE_TIMEOUT)
```

Idain:
Yeah, I can do it. Give me a few minutes.

Rangi42:
Thanks a lot!

Idain:
I found some cases triggered by a `; TRUE` comment. I don't think we need `assert TRUE == 1`, do we...?

Idain:
Also, there are a lot of cases for `wBattleMode` where it decreases, making the assumption `WILD_BATTLE == 1` and `TRAINER_BATTLE == 2` (around ~65 cases). Do I also have to add `assert`s for those as well?

Idain:
Have fun, @Rangi42. :)

mid-kid:
My biggest fear is excessive `assert`s cluttering the codebase.

One source of excessive asserts as exemplified by this PR is 0 and 1-index checks (and sets, i.e. `xor a`).

Do we *really* want this? Personally, I'd argue no. Most instances of this are trivial enums that won't be reordered (e.g. WILD_/TRAINER_BATTLE), or instances where the 0 value has a special meaning (e.g. LINK_NULL).

mid-kid:
Years ago we considered macros like `lda x` to pick between `ld a, x` and `xor a` depending on the value of `x`. IMO this is mildly more acceptable than the assert approach, but a hard sell in terms of readability.

Rangi42:
I'd be very against a `lda` macro, where it's not clear how many bytes or cycles it would take or what the effect on flags would be. In which case I'd rather just leave out some of the more trivial `assert`s.

mid-kid:
`lda` would ideally only be used in places where cycle count doesn't matter, but I digress. Not a relevant topic here.

> In particular, this is a lot of assert LINK_NULL == 0 and assert WILD_BATTLE == 1, which aren't even likely to change.

Exactly, I think the asserts are cluttery due to this.

Idain:
I can delete those asserts without problems.

Idain:
Feel free to check again. @Rangi42 @mid-kid

engine/battle/core.asm
```diff
@@ -3826,7 +3827,8 @@ TryToRunAwayFromBattle:
 	cp BATTLEACTION_FORFEIT
 	ld a, DRAW
 	jr z, .fled
-	dec a ; LOSE
+	assert LOSE + 1 == DRAW
+	dec a
```
vulcandth, line 3831: I doubt WIN/LOSE/DRAW will be reordered.. I would expect other issues if someone tried.

vulcandth (commented):
I am all for asserts, but not for every little thing. I think most of these that are left are fine. Basically we want to leave it to instances where people are likely to change, and likely to miss something when doing so.

engine/battle/core.asm
Rangi42: Also since this operation is `dec`, I think `assert DRAW - 1 == LOSE` would be more obviously related if we even keep this `assert`. (This is the only place it occurs, so I think it's fine to have one.)

engine/battle_anims/functions.asm
```diff
@@ -1844,8 +1844,10 @@ BattleAnimFunc_Wrap:
 	ld hl, BATTLEANIMSTRUCT_FRAMESET_ID
 	add hl, bc
 	ld a, [hl]
-	inc a ; BATTLE_ANIM_FRAMESET_BIND_2
-	      ; BATTLE_ANIM_FRAMESET_BIND_4
+	assert BATTLE_ANIM_FRAMESET_BIND_1 + 1 == BATTLE_ANIM_FRAMESET_BIND_2 \ 
+	&& BATTLE_ANIM_FRAMESET_BIND_2 + 1 == BATTLE_ANIM_FRAMESET_BIND_3 \
+	&& BATTLE_ANIM_FRAMESET_BIND_3 + 1 == BATTLE_ANIM_FRAMESET_BIND_4
```
Rangi42, line 1849: We have very few `\` line continuations, but they are indented afterwards.
```suggestion
	assert BATTLE_ANIM_FRAMESET_BIND_1 + 1 == BATTLE_ANIM_FRAMESET_BIND_2 \ 
		&& BATTLE_ANIM_FRAMESET_BIND_2 + 1 == BATTLE_ANIM_FRAMESET_BIND_3 \
		&& BATTLE_ANIM_FRAMESET_BIND_3 + 1 == BATTLE_ANIM_FRAMESET_BIND_4
```

engine/battle_anims/functions.asm
```diff
@@ -2528,15 +2531,16 @@ BattleAnimFunc_Dizzy:
...
+	inc a
 	jr .got_frameset
 
 .not_flipped
 	ld hl, BATTLEANIMSTRUCT_VAR1
 	add hl, bc
-	ld a, [hl] ; BATTLE_ANIM_FRAMESET_61
-	           ; BATTLE_ANIM_FRAMESET_9C
+	ld a, [hl] ; BATTLE_ANIM_FRAMESET_CHICK_1
+	           ; BATTLE_ANIM_FRAMESET_IMP
```
Rangi42, line 2543:
```suggestion
	ld a, [hl] ; BATTLE_ANIM_FRAMESET_CHICK_1 or BATTLE_ANIM_FRAMESET_IMP
```

engine/battle/core.asm
Idain: You have a point.

Idain:
So, what's your verdict? @Rangi42 @vulcandth

engine/battle_anims/functions.asm
```diff
@@ -1244,7 +1244,9 @@ BattleAnimFunc_Sing:
 	ld hl, BATTLEANIMSTRUCT_PARAM
 	add hl, bc
 	ld a, BATTLE_ANIM_FRAMESET_MUSIC_NOTE_1
-	add [hl] ; BATTLE_ANIM_FRAMESET_MUSIC_NOTE_2 BATTLE_ANIM_FRAMESET_MUSIC_NOTE_3
+	assert BATTLE_ANIM_FRAMESET_MUSIC_NOTE_1 + 1 == BATTLE_ANIM_FRAMESET_MUSIC_NOTE_2 \
+		&& BATTLE_ANIM_FRAMESET_MUSIC_NOTE_2 + 1 == BATTLE_ANIM_FRAMESET_MUSIC_NOTE_3
```
Rangi42, line 1248:
```suggestion
	assert BATTLE_ANIM_FRAMESET_MUSIC_NOTE_1 + 1 == BATTLE_ANIM_FRAMESET_MUSIC_NOTE_2
	assert BATTLE_ANIM_FRAMESET_MUSIC_NOTE_2 + 1 == BATTLE_ANIM_FRAMESET_MUSIC_NOTE_3
```

Rangi42 (approved):
Thank you!

PR #1121, opened 2024-05-10
Add Hatch_*_STEPS constants

etdv-thevoid:
Adds `HATCH_*_STEPS` constants for the various hatch step cycle count values. Updates all `data/pokemon/base_stats/*` files to use the appropriate constant as well as the odd eggs in `data/events/odd_eggs.asm`.


Rangi42 (changes requested):
Sorry, I don't think this change is a good idea. "Steps taken to hatch" is a meaningfully numeric quantity, just like "base stat values" or "PP" or so on. "This Pokemon takes a 'slower' number of steps to hatch" is not informative.

Contrast that with the growth rate constants, which are sequential identifiers for formulas in a lookup table. Or the gender constants, which are also numeric data, but are named after it (so `GENDER_F25` is 63, aka `25 percent` female). (And even for the gender constants, I could see just using raw percentage values, except `GENDER_UNKNOWN` would still be special.)

Anyway. I could *maybe* see having self-explanatory constants, like `DEF HATCH_1280 EQU 5`, `DEF HATCH_2560 EQU 10`, etc (since one step *cycle* is 256 steps). But I also think just having the raw numbers is okay.

mid-kid:
Thank you for your first contribution! We really appreciate people making a conscious effort to not only spot wrinkles in code style/names/etc but also fix them.

That said, with any sort of constants the question is: does the value make more sense as a number, or does that numer have some kind of hidden meaning? In this case, an "egg cycle" is a multiple of 256 steps, and I think the numbers make sense on their own. The `GENDER_F` constants were introduced to clarify that the percentage represents the female, but I don't inmediately see any similar reason to this change. Do these hatch cycle constants reflect any common official or fan names for them? Do they map to other values in other games?

I'll leave this up for others to opine on, but to me it looks a bit unnecessary.

etdv-thevoid:
Thanks for the feedback, both of you.

I went ahead and renamed the constants to how many steps they equal and added some comments to clary intent.

Idain:
I feel constants here are limiting, because I could make a Pokémon hatch in, let's say, 14 cycles, and that's a value a lot of romhackers would want to change for existing or even some fakemon.

PR #1123, opened 2024-05-22
Use constants for VBlank handlers

Rangi42:
Note that some constants would be different in G/S.

### DIFF
--- a/constants/wram_constants.asm
+++ b/constants/wram_constants.asm
@@ -324,3 +324,16 @@ DEF DAYCARELADY_ACTIVE_F         EQU 7
 	const UNLOCKED_UNOWNS_S_TO_W_F
 	const UNLOCKED_UNOWNS_X_TO_Z_F
 DEF NUM_UNLOCKED_UNOWN_SETS EQU const_value
+
+; hVBlank::
+; VBlankHandlers indexes (see home/vblank.asm)
+	const_def
+	const VBLANK_NORMAL       ; 0
+	const VBLANK_CUTSCENE     ; 1
+	const VBLANK_SOUND_ONLY   ; 2
+	const VBLANK_CUTSCENE_CGB ; 3
+	const VBLANK_SERIAL       ; 4
+	const VBLANK_CREDITS      ; 5
+	const VBLANK_DMA_TRANSFER ; 6
+	const VBLANK_UNUSED       ; 7
+DEF NUM_VBLANK_HANDLERS EQU const_value
--- a/engine/battle_anims/anim_commands.asm
+++ b/engine/battle_anims/anim_commands.asm
@@ -24,11 +24,11 @@ _PlayBattleAnim:
 	call BattleAnimRequestPals
 	call BattleAnimDelayFrame
 
-	ld c, 1
+	ld c, VBLANK_CUTSCENE
 	ldh a, [rKEY1]
 	bit 7, a ; check CGB double speed mode
 	jr nz, .got_speed
-	ld c, 3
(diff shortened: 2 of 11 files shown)


mid-kid:
Looks great!

constants/wram_constants.asm
```diff
@@ -324,3 +324,16 @@ DEF DAYCARELADY_ACTIVE_F         EQU 7
 	const UNLOCKED_UNOWNS_S_TO_W_F
 	const UNLOCKED_UNOWNS_X_TO_Z_F
 DEF NUM_UNLOCKED_UNOWN_SETS EQU const_value
+
+; hVBlank::
```
vulcandth, line 328: I wouldn't in this PR, but similarly how `macros/wram.asm` was renamed to `macros/ram.asm` we should probably rename this file to `constants/ram_constants.asm` if its going to hold constants used by hram.

home/vblank.asm
```diff
@@ -34,17 +34,20 @@ VBlank::
...
-	dw VBlank1
-	dw VBlank2
-	dw VBlank3
-	dw VBlank4
-	dw VBlank5
-	dw VBlank6
-	dw VBlank0 ; just in case
-
-VBlank0::
+VBlankHandlers:
```
vulcandth, line 37: Question; if the pointer table is not really being referenced elsewhere and is located directly after the routine that uses it, should it remain as a local label? Not sure if we have enforced any precedence on this. Just curious on your thoughts.

engine/battle_anims/anim_commands.asm
```diff
@@ -24,11 +24,11 @@ _PlayBattleAnim:
...
 
-	ld c, 1
+	ld c, VBLANK_CUTSCENE
 	ldh a, [rKEY1]
 	bit 7, a ; check CGB double speed mode
 	jr nz, .got_speed
-	ld c, 3
+	ld c, VBLANK_CUTSCENE_CGB
 
 .got_speed
```
vulcandth, line 33: This is not really a CGB check... Its a double speed check. So it wouldn't be accurate imo to call it VBLANK_CUTSCENE_CGB.

Furthermore, I think you got this backwards, unless i'm reading it wrong...

if bit 7 of rKEY1 is set (Double-speed mode), then we set `c` to `1`.

If bit 7 of rKEY1 is not set (Single Speed), then we set `c` to `3`.

vulcandth (changes requested):

home/vblank.asm
Rangi42: I think it's fine. There aren't many single-colon labels in home/, but there are some, so we don't have an absolute convention of "export every home label" (and probably can/should make more of them local, to be explicit about what *does* get referenced by ROMX).

engine/battle_anims/anim_commands.asm
Rangi42: The thing is, vblank handler 3 was unused in DMG-supporting G/S, but Crystal repurposes it to call `ForceUpdateCGBPals` if `[hCGBPalUpdate]` is true. Without reference to which code happens to use that handler, I think "CGB cutscene" makes sense for it.

(Compare with `VBLANK_SERIAL`, which is only used by the printer and could be called `VBLANK_PRINTER`, but what actually distinguishes it is calling `AskSerial`.)

(Compare also with `VBLANK_CREDITS`, which IMO is specifically tuned to the credits -- e.g. only updating SCX but not SCY.)

constants/wram_constants.asm
Rangi42: To do in a subsequent PR!

PR #1126, opened 2024-06-12
Add a section "How do I find absolute ROM offsets?" to the FAQ.md

risingPhil:
I saw in the discord that you'd be happy to accept documentation improvements.

So I'm doing this to describe what I learnt in the past few days and make it easier for others :)

I hope you're okay with this.

### DIFF
--- a/FAQ.md
+++ b/FAQ.md
@@ -16,6 +16,7 @@
   - ["Syntax error"](#syntax-error)
 - [How do I edit maps?](#how-do-i-edit-maps)
 - [How do I edit the colors of an image?](#how-do-i-edit-the-colors-of-an-image)
+- [How do I find absolute ROM offsets?](#how-do-i-find-absolute-ROM-offsets)
 - [How do I write new features?](#how-do-i-write-new-features)
 - [How do I share code on Discord?](#how-do-i-share-code-on-discord)
 - [I need more help!](#i-need-more-help)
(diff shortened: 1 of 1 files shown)


FAQ.md
```diff
@@ -85,6 +86,24 @@ Some image `.png` files are grayscale. This indicates that even though these ima
 
 It really depends on what image you're trying to change the colors of, where these colors are specified. Try looking for related files or `.pal` files.
 
+## How do I find absolute ROM offsets?
+When building external tools, it is usually necessary to know where the data you need is stored.
```
Rangi42, line 90: Have you been building tools that directly modify the ROM? You may want to be modifying the source files instead; this is what [Polished Map](https://github.com/Rangi42/polished-map), [Crystal Tracker](https://github.com/dannye/crystal-tracker), and [Tilemap Studio](https://github.com/Rangi42/tilemap-studio) do (as well as various Python scripts in the tools folder).

FAQ.md
```diff
@@ -85,6 +86,24 @@ Some image `.png` files are grayscale. This indicates that even though these ima
...
 It really depends on what image you're trying to change the colors of, where these colors are specified. Try looking for related files or `.pal` files.
 
+## How do I find absolute ROM offsets?
+When building external tools, it is usually necessary to know where the data you need is stored. 
+
+The first step is to browse the directories of the project and check files that look related to what you're looking for.
+
+Once you've found what you need there, you'll likely end up finding a name in one of the .asm files (example: PokemonPicPointers)
+
+Now it's time to look at the `pokecrystal.map` file. You'll get this file after building the project with `make`. Search the name you found earlier in this file to find a hexadecimal offset. Wait! You're not done yet!
```
Rangi42, line 96: The .sym file would be easier to work with for this use case.

To find the absolute offset of a label like `BaseData`, you just find the line for that label in pokecrystal.sym; in this case it's `14:5424 BaseData`. Then calculate `offset = (bank * $4000) + (address - $4000)`; in this case it's `($14 * $4000) + ($5424 - $4000) = $51424`.

(I changed your example because having the address be $4000 is confusing with the constant $4000s.)

FAQ.md
```diff
@@ -85,6 +86,24 @@ Some image `.png` files are grayscale. This indicates that even though these ima
 
 It really depends on what image you're trying to change the colors of, where these colors are specified. Try looking for related files or `.pal` files.
 
+## How do I find absolute ROM offsets?
+When building external tools, it is usually necessary to know where the data you need is stored. 
+
+The first step is to browse the directories of the project and check files that look related to what you're looking for.
```
Rangi42, line 92: A guide written at this level is probably more appropriate for a [wiki](https://github.com/pret/pokecrystal/wiki/Tutorials) page. You'd also have more immediate control that way, not needing a PR process to get your writing out there.

FAQ.md
Rangi42: No problem, just recommending .sym instead of .map since the bank and address are together on the same line. (Also some projects don't output all the symbols in the .map file, only the section names.)

Rangi42: Btw, be aware of the [`symbols` branch](https://github.com/pret/pokecrystal/tree/symbols) that makes the .sym and .map files available online without building yourself.

Rangi42:
Nice! You may also be interested in [how cheat codes work](https://github.com/pret/pokecrystal/wiki/Discovering-GameShark-cheat-codes).

PR #1132, opened 2024-07-28
Rename Bill's brother to Bill's older sister

metallicity:
I noticed that NPC attached to `BillsBrotherScript` had a sprite of `SPRITE_POKEFAN_F`, and wondered if the labels here were maybe incorrect. While I couldn't find anything beyond the sprite itself that seemed particularly definitive, the same house in HGSS also used a similar sprite, and [Bulbapedia](https://bulbapedia.bulbagarden.net/wiki/Bill) currently describes Bill as having two sisters (rather than a sister and a brother), so I went ahead and wrote up a commit for all the related renames.

Bill having two sisters does introduce some potential for naming collisions (i.e., `BillsSisterScript`), so I used `BillsYoungerSister` and `BillsOlderSister` to avoid this. The Goldenrod NPC being younger than Bill is perhaps a bit more clear from the in-game text/sprites, but Brent's phone gossip script about ["Bill's younger sister"](https://github.com/pret/pokecrystal/blob/9ae05312d6c039902d885de98d49407ffe3a3c79/data/phone/text/brent_caller.asm#L104) would technically be ambiguous if both were younger than Bill, and regardless the Fuchsia NPC would be the older of Bill's two sisters.


vulcandth:
Just wanted to leave a comment, that we haven't forgotten about this, but have been a bit busy with some life and other projects. I'll need to review your references and look over everything; hopefully have an update for you this weekend. Thank you for the contribution regardless 😄

PR #1141, opened 2024-09-21
Rename `dt` and `dd` to `bigdt` and `bigdd`

Rangi42:
This makes their big-endianness obvious and matches `bigdw`


vulcandth (approved):
Looks good to me.

PR #1153, opened 2024-11-29
Identify more bit flags

Rangi42:
Based on https://github.com/pret/pokegold/pull/120. Many mobile-related bits are still unidentified, but the rest should be on par with pokegold.


vulcandth (approved):
Looks good for a sync with pokegold's PR. The rest of the mobile stuff will come eventually in future PR's. Thanks!!!

PR #1172, opened 2025-03-08
Support `<` and `>` prefixes for high and low bytes in VC patch templates

Rangi42:
This is not needed for pokecrystal (nor for pokegold, pokered, or pokeyellow), but it *is* needed for https://github.com/Narishma-gb/pokegreen/commit/24ad50fa9c7295e7b2fd003a7b40386cc02a7d32. @Narishma-gb Can you please review this and verify that it works for you? The syntax is based on what some other assemblers besides RGBASM have used for low/high bytes, and it's easier to parse than `LOW()`/`HIGH()`. So in pokegreen, `LOW(Subanim_1StarBigMoving+7)` would become `<Subanim_1StarBigMoving+7`.


Narishma-gb:
This works perfectly in Narishma-gb/pokegreen@10335f1, thanks! I'll sync green (and blue) when this is merged in pokered.

vulcandth (approved):

PR #1179, opened 2025-04-28
Rename and document ``BrokenPlacePPUnits``

Brianum:


mid-kid:
Thanks for your contribution!
I see you added an extra line of code that shouldn't be there, changing the bytes in the ROM and breaking the CI with it. Can you look at that?

PR #1185, opened 2025-06-03
Make town map limits more intuitive

Brianum:
Address the hard-coded town map logic as outlined [here](https://github.com/pret/pokecrystal/wiki/Hard-coded-logic#landmark-limits-when-scrolling-in-the-town-map) to make things a little bit more intuitive.

### DIFF
--- a/constants/landmark_constants.asm
+++ b/constants/landmark_constants.asm
@@ -2,7 +2,6 @@
 	const_def
 	const LANDMARK_SPECIAL           ; 00
 
-; Johto landmarks
 DEF JOHTO_LANDMARK EQU const_value
 	const LANDMARK_NEW_BARK_TOWN     ; 01
 	const LANDMARK_ROUTE_29          ; 02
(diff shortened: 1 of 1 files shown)


Rangi42 (changes requested):
I like the idea of first/last constants, like how we have `JOHTO_POKEMON` for the first Gen 2 mon (Chikorita). However, I don't think the comments need to be added: they're restating what's clear from adjacent constant definitions.

data/maps/landmarks.asm
```diff
@@ -8,6 +8,7 @@ Landmarks:
 ; entries correspond to constants/landmark_constants.asm
 	table_width 4
 	landmark  -8, -16, SpecialMapName
+	assert_table_length JOHTO_LANDMARK
```
Rangi42, line 11:
```suggestion
```

data/maps/landmarks.asm
```diff
@@ -103,6 +104,7 @@ Landmarks:
 	landmark  20, 100, Route27Name
 	landmark  12, 100, TohjoFallsName
 	landmark  20,  68, Route28Name
+	assert_table_length OTHER_LANDMARK
```
Rangi42, line 107: You could add an `assert_table_length KANTO_LANDMARK`, but I don't think they need to be here -- every line is supposed to pair up with its constant already, so these asserts wouldn't catch extra errors.
```suggestion
```

Rangi42 (approved):
Thank you!

PR #1186, opened 2025-06-05
Replace hardware_constants.asm with the community-standard hardware.inc

### LINKED ISSUE #914: Use `hardware.inc` from gbdev
@ISSOtm mentioned this on discord today, that we should be using `hardware.inc` from https://github.com/gbdev/hardware.inc/blob/master/hardware.inc. Looking at the discord history, It sems this was discussed in the past, but I didn't see a clear definitive reason why or why not use it. A couple points on why not seemed to do with extra "noise", that it might be based on MBC5, and how pokecrystal uses more constants related to individual bits. 

Anyways, I'm going to let you all who have more experience weigh in on this. I'm just opening this to start a discussion. If we decide to take action with this, i'll help out.

Rangi42:
Fixes #914

Now that [hardware.inc](<https://github.com/gbdev/hardware.inc>) consistently defines bit numbers and flags for all the hardware registers, it's feasible to replace hardware_constants.inc (and much of gfx_constants.asm) with it. Although not all its naming conventions are great, they'll be more recognizable to GB programmers in general. And the old names are supported in legacy.asm.


Rangi42:
hardware.inc's naming convention is to use <code><i>REG</i>B_<i>NAME</i></code> for 0-7 bit indexes, and <code><i>REG</i>F_<i>NAME</i></code> for flags (<code>1 &lt;&lt; <i>REG</i>B_<i>NAME</i></code>). IMO it's not a *good* convention, but hardware register names are already pretty terse/obscure. And I do like that this gets rid of the `r` prefixes on non-register constants (e.g. we used to do `ld a, rRP_ENABLE_READ_MASK`; now it's `ld a, RPF_ENREAD`).

Rangi42:
> I very particularly dislike the naming convention `REGB_NAME` and `REGF_NAME`, I also dislike how it mismatches our own established convention... but, that being said I can get over it. I would request that we open an issue if this gets merged to talk about if we want to standardize on a convention.

I agree it's not a good convention, but wonder if it would be worth changing `REGB_NAME` to `REG_NAME_F` and `REGF_NAME` to `REG_NAME`. It would be more familiar for *us*, but no longer "in line with the majority of the Gameboy dev community." (Ideally we'd manage to convert the official hardware.inc itself to a new convention, but the most active users are also the most attached to tradition.)

> I do think we should keep the old names in the legacy file... or at least some of them. Like any of the ones that are likely to be copy+pasted. I'm thinking like if someone was going to copy+paste in sourapple's animation stuff. Those OAM flags should probably be aliased.

Yeah, definitely for the ones that get commonly used.

> Seems weird that hardware.inc is using `rVDMA` when the pandocs references it as `HDMA`. Likewise with `rKEY1` vs `SPD`... but this is either a nitpic with hardware.inc vs pandocs... and not for this PR.

The `rVDMA_*` aliases are new, and Pan Docs doesn't yet document them. `rSPD` is somewhat older, but still not listed there. (Pan Docs usually just documents the "official" register names, with some exceptions like `JOYP` and [mentioning](https://gbdev.io/pandocs/Audio_Registers.html#:~:text=One%20of%20the%20pitfalls) the `rNRxx` ones. ISSOtm and I would like it to be more in sync with hardware.inc, since both are official GBDev resources.)

vulcandth:
> I need some more opinions on this though; either one is too drastic for me to just commit to all the disassemblies.

*IF* we do this, that is switch to using hardware.inc, I only want to make this drastic change once... By that I mean, I don't want to switch to using `REGB_NAME` stuff, and then have hardware.inc change the convention, and us have to rename things *again*.

So if we *are* going to push to revamp hardware.inc... lets do that first, before this. However, if we get pushback from the rest of the gb dev community, then fine.. we can adapt to play ball.

That's just my opinion on our direction.. and we need some more input from pfero and dannye

Rangi42:
One motivator here is that currently hardware_constants.asm doesn't define a lot of bit flags for its registers. Now that hardware.inc defines them for all registers, it would be convenient to add its constants, and then replace some hard-coded values with them. The alternative is to invent our own names, which could be based on hardware.inc, but following pret's previous naming conventions.

Rangi42:
> Ultimately, I would probably vote for using gbdev's hardware.inc as-is (whatever that ends up being) without sprinkling our own changes on top, and then having a more brief hardware_constants.asm with the most important legacy aliases.

I expect they'll go in macros/legacy.asm with the rest, but yeah.

vulcandth:
> Although we shouldn't wholly adopt hardware.inc's conventions as our own, the `TEXTBOX_X`/`TEXTBOX_Y` constants do clash with the use of `_X`/`_Y` suffixes in `SCRN_X`/`SCRN_Y`, and those constants get used frequently near each other. We can just rename them to `TEXTBOX_X_COORD`/`TEXTBOX_Y_COORD`.

I'd actually be fine with aliasing some of the hardware.inc constants. At least by looking up one it will trace back to the original constant in hardware.inc

Rangi42:
> > Although we shouldn't wholly adopt hardware.inc's conventions as our own, the `TEXTBOX_X`/`TEXTBOX_Y` constants do clash with the use of `_X`/`_Y` suffixes in `SCRN_X`/`SCRN_Y`, and those constants get used frequently near each other. We can just rename them to `TEXTBOX_X_COORD`/`TEXTBOX_Y_COORD`.
> 
> I'd actually be fine with aliasing some of the hardware.inc constants. At least by looking up one it will trace back to the original constant in hardware.inc

@vulcandth Can you give some examples of what you mean here?

Rangi42:
hardware.inc 5.0 has improved a lot of the constant names, including changing the `F`/`B` convention, so this PR now uses it! Should be a much less painful change from hardware_constants.asm.

PR #1188, opened 2025-06-16
Document Time Capsule glitch

Brianum:
(See https://tcrf.net/Bugs:Pokémon_Gold_and_Silver#Time_Capsule_glitch)

### DIFF
--- a/docs/bugs_and_glitches.md
+++ b/docs/bugs_and_glitches.md
@@ -1775,7 +1775,7 @@ LearnLevelMoves:
  
 -	push hl
  	ld d, a
-+	cp STRUGGLE + 1
++	cp GEN_2_MOVE
 +	jr c, .old_move
 +
 +	ld a, [wLinkMode]
--- a/constants/move_constants.asm
+++ b/constants/move_constants.asm
@@ -171,6 +171,7 @@
 	const SLASH        ; a3
 	const SUBSTITUTE   ; a4
 	const STRUGGLE     ; a5
+DEF GEN_2_MOVE EQU const_value
 	const SKETCH       ; a6
 	const TRIPLE_KICK  ; a7
 	const THIEF        ; a8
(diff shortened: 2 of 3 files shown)


docs/bugs_and_glitches.md
```diff
@@ -1749,6 +1750,45 @@ When a Pokémon is withdrawn from the Day-Care, its Exp. Points are reset to the
...
+
+	ld b, a
+	ld a, [wCurPartyLevel]
+	cp b
+ 	ld a, [hli]
+ 	jr nz, .find_move
+ 
+-	push hl
+ 	ld d, a
++	cp STRUGGLE + 1
```
Rangi42, line 1778: Do this:
```asm
	const STRUGGLE     ; a5
DEF GEN_2_MOVE EQU const_value
	const SKETCH       ; a6
```

```suggestion
+	cp GEN_2_MOVE
```

docs/bugs_and_glitches.md
```diff
@@ -1749,6 +1750,45 @@ When a Pokémon is withdrawn from the Day-Care, its Exp. Points are reset to the
  	ret
 ```
 
+### Time Capsule glitch
+
+Generation I Pokémon that evolve through trading (Graveler, Kadabra, Machoke, and Haunter) can sidestep the check that prevents Pokémon with Generation II moves from being sent back to Generation I.
+
+To do this, they must be traded at the exact level they would learn a move in Generation II after evolving (for example, a Level 34 Graveler sent to Generation II will evolve to Golem and then immediately learn Rollout). After sending the Pokémon to Generation II, they will evolve and learn the move. From there, they can be immediately sent back to Generation I, and that will not trigger the move check. Any Generation II-exclusive moves will become glitch moves in Generation I.
```
Rangi42, line 1757: Pare down or remove this description, and give it a more descriptive name than "Time Capsule glitch". Maybe "Trade evolutions in the Time Capsule can send glitch moves to Gen 1".

PR #1191, opened 2025-06-29
Identify wCrystalData's content and rename memory locations accordingly

Brianum:


ram/sram.asm
```diff
@@ -215,10 +217,9 @@ s4_a03b:: ds 37 * 40
 s4_a603:: ds 8
 
 s4_a60b:: db
-s4_a60c:: db
-s4_a60d:: db
+s4_a60c:: dw
 
-s4_a60e:: dw
+sCrystalFlags:: ds 2
```
Rangi42, line 222: `wCrystalFlags` is one byte, right? So this can be `sCrystalFlags:: db`, and change the `ds $1f0` below to `$1f1`.

ram/wram.asm
```diff
@@ -2868,15 +2868,15 @@ wPlayerGender::
...
-wd473:: ds 1
-wd474:: ds 1
-wd475:: ds 1
-wd476:: ds 1
-wd477:: ds 1
-wd478:: ds 1
+; mobile profile
+wPlayerAge:: ds 1
+wPlayerPrefecture:: ds 1
+wPlayerZipcode:: ds 4
```
Rangi42, line 2874: Japan has [postal codes](https://en.wikipedia.org/wiki/Postal_codes_in_Japan), not ZIP codes.
```suggestion
wPlayerPostalCode:: ds 4
```

Rangi42:
Thanks for this PR! It's great to see the remaining JP mobile code being gradually identified.

How did you determine what these particular bytes were for?

Rangi42 (approved):
Thanks!

PR #1202, opened 2025-10-02
update trainer constants for rematches

shasocastris:
Attempt to better lineup trainer constants for rematchs, `.LoadfightX` and `.FightX` functions, and w*FightCount variable values; doesn't quite lineup for trainers like GAVEN and ANTHONY


constants/trainer_constants.asm
Rangi42, line 1: Is there any precedent for trainer constants to start from 0? I think it would be confusing by now for things like "`RALPH2`" to now mean something else. (Consider people merging this into their forks.)

Rangi42 (changes requested):

constants/trainer_constants.asm
shasocastris: The goal was to match the variable count in the rematches. It's admittedly a quality of life change and not necessary, I just found it something of a pain to keep track of the rematch logic with variables that were off by 1.

I guess in my mind, it is similar to changing BATTLETYPE_SHINY to BATTLETYPE_FORCESHINY. It will certainly require merging, but that type of breaking has occurred before.

Also, I don't know if it was intentional, but I definitely think I messed up with the RALPHX variable names, so I'll need to fix that

PR #1203, opened 2025-10-10
Update bugs_and_glitches.md

### LINKED ISSUE #1148: Verify and document another bug in the catching formula
According to this article... [https://www.dragonflycave.com/mechanics/gen-ii-capturing](https://www.dragonflycave.com/mechanics/gen-ii-capturing)

...**a wild Pokemon with more than 341 max HP will cause errors in the catching formula, including possible game freezes due to division by 0.** The article explains the mechanics of the error in detail, but I don't understand the assembly commands well enough to verify or fix it myself, so the most I can do is bring attention to it.

No wild Pokemon in vanilla can have this much HP, which is probably why this has gone unnoticed. But wild Chansey or Blissey exceeding Lv50 or so will start to enter this range, and that's hardly an unreasonable thing for someone to put in their hack.

This was originally brought up by Pyrois in the Discord server, who found it on Bulbapedia, which seems to have gotten it from the article linked above.

Pyrois81:
Add a fix for a potential failure condition in the catch rate function.

Fixes #1148

### DIFF
--- a/docs/bugs_and_glitches.md
+++ b/docs/bugs_and_glitches.md
@@ -47,7 +47,7 @@ Fixes in the [multi-player battle engine](#multi-player-battle-engine) category 
   - [Love Ball boosts catch rate for the wrong gender](#love-ball-boosts-catch-rate-for-the-wrong-gender)
   - [Fast Ball only boosts catch rate for three Pokémon](#fast-ball-only-boosts-catch-rate-for-three-pok%C3%A9mon)
   - [Heavy Ball uses wrong weight value for three Pokémon](#heavy-ball-uses-wrong-weight-value-for-three-pok%C3%A9mon)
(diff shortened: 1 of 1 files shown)


docs/bugs_and_glitches.md
```diff
@@ -1196,6 +1197,43 @@ Note that this fix only accounts for Pokémon that evolve via Moon Stone as thei
...
++.check_cur_low
+	ld a, c
+	and a
+	jr nz, .okay_1
+	ld c, $1
+.okay_1
+	ld b, e
+```
+
+Note: We only divide again if required in order to preserve as much precision as possible in all other cases.
```
Rangi42:
```suggestion
```

Rangi42 (changes requested):

PR #1204, opened 2025-11-08
Use features of RGBDS 1.0.0

Rangi42:
RGBDS 1.0.0 was released on October 31. So far we've found a couple bugs (both involving erroneous user input, so neither affects pret), and we want to release a 1.0.1 bugfix this year (so, within the next couple months). To that end, it will help if 1.0.0 gets more testing and feedback from real users.

- Use single-quoted character literals or `CHARVAL` when you need a character's numeric value; using strings as numbers is deprecated
- Use `?` to silence a node in warning/error location backtraces
- Use `===` and `!==` instead of `STRCMP`, and `++` instead of `STRCAT`
- Use `__SCOPE__` instead of checking for `..` and `.` separately
- Use `#string` instead of `"{string}"`
- All programs (rgbasm, rgblink, rgbfix, rgbgfx) support `-W` warnings
- `rgbgfx --colors dmg` is short for `rgbgfx --colors dmg=e4`


Makefile
```diff
@@ -149,16 +154,17 @@ $(foreach obj, $(pokecrystal11_vc_obj), $(eval $(call DEP,$(obj),$(obj:11_vc.o=.
...
-pokecrystal_debug_opt   = -Cjv -t PM_CRYSTAL -i BYTE -n 0 -k 01 -l 0x33 -m MBC3+TIMER+RAM+BATTERY -r 3 -p 0
-pokecrystal11_debug_opt = -Cjv -t PM_CRYSTAL -i BYTE -n 1 -k 01 -l 0x33 -m MBC3+TIMER+RAM+BATTERY -r 3 -p 0
-pokecrystal11_vc_opt    = -Cjv -t PM_CRYSTAL -i BYTE -n 1 -k 01 -l 0x33 -m MBC3+TIMER+RAM+BATTERY -r 3 -p 0
+RGBFIXFLAGS  += -Cjv -t PM_CRYSTAL -k 01 -l 0x33 -m MBC3+TIMER+RAM+BATTERY -r 3 -p 0
+pokecrystal.gbc:         RGBFIXFLAGS += -i BYTE -n 0
+pokecrystal11.gbc:       RGBFIXFLAGS += -i BYTE -n 1
+pokecrystal_au.gbc:      RGBFIXFLAGS += -i BYTU -n 0
+pokecrystal_debug.gbc:   RGBFIXFLAGS += -i BYTE -n 0
+pokecrystal11_debug.gbc: RGBFIXFLAGS += -i BYTE -n 1
+pokecrystal11_vc.gbc:    RGBFIXFLAGS += -i BYTE -n 1
```
dannye: Much nicer!

dannye (approved):
I tested it out and it LGTM!

PR #1206, opened 2025-11-10
Don't hard-code the Stadium data handling in tools/make_patch.c

Rangi42:
Instead, have a `--ignore addr:size` option that's set in the Makefile

This will let us copy the exact same make_patch.c file to all repos (even non-Pokémon ones).


Makefile
```diff
@@ -118,7 +118,7 @@ $(pokecrystal11_debug_obj): RGBASMFLAGS += -D _CRYSTAL11 -D _DEBUG
 $(pokecrystal11_vc_obj):    RGBASMFLAGS += -D _CRYSTAL11 -D _CRYSTAL11_VC
 
 %.patch: %_vc.gbc %.gbc vc/%.patch.template
-	tools/make_patch $*_vc.sym $^ $@
+	tools/make_patch --ignore 0x1ffde0:0x220 $*_vc.sym $^ $@
```
vulcandth: Lets add a comment here mentioning the ignore is for stadium data. This way if people are removing stadium data in their hacks and grep for the word "stadium" they will see they need to make a change here.

PR #1217, opened 2026-01-02
Implemented all bug fixes

ImJustHutch:


runlevel5:
@ImJustHutch why is this PR closed?

mid-kid:
@runlevel5 Not everyone can agree on what is or isn't a bug, and even where that's possible, everyone has different takes on how to fix them. We document everything in docs/bugs_and_glitches.md so hackers can use their own judgement. It provides an example of the simplest possible fix, but in quite a few cases a more proper fix would be nontrivial. In general, I don't recommend going around fixing everything, the vast majority are not noticeable during normal gameplay, or can only be exposed with specific modifications.

PR #1224, opened 2026-01-22
Update `link.asm` and WRAM link transfer related symbols

Narishma-gb:
This PR is a twin of pret/pokegold#141.

Aside from the mail translation system, trade functions are identical in order to preserve compatibility.
Opened as a draft, in order to perform a final check, in case I forgot to port some of the changes (comments, and some WRAM parts that would not show in the build check).


Narishma-gb:
This should (hopefully) bring this PR up to date with https://github.com/pret/pokegold/pull/141.

Rangi42 (approved):
Thank you!

PR #1226, opened 2026-01-26
Distinguish codepoints from characters

SnorlaxMonster:
In several places in the repo, character references are used to reference the codepoints those characters are defined at.

## Status quo
There are several occasions on which a new graphic is loaded into a particular codepoint, but at the moment the disassembly presents this as the _character_ being overwritten rather than the _codepoint_.

Additionally, there are two sections of the code that detect control characters, and to do so they define which ranges of characters are control characters and which are regular text (the actual code is based on the Japanese encoding, as pretty much all of the valid English codepoints are ≥0x7F). But at the moment, these ranges are defined using the characters at the endpoints (or next to the endpoints) instead of as ranges of codepoints. One particular issue with this approach is that sometimes adjacent characters have to be used due to the endpoint not being a defined character at all — for example, 0x05 ガ has to be used as an endpoint instead of 0x04 (which renders as オ゙), because there's no actual character defined at 0x04; and `<JP_18>` (which renders as ノ゙) is defined solely to be used as an endpoint in these ranges.

## Changes
In both cases, I think these values are best understood as codepoints rather than characters, so should be represented by hexadecimal codepoints rather than strings. I recognise that this change may be quite controversial, so I'm interested in feedback on whether the maintainers agree with my opinion here. (That's why despite making these changes locally in 2023, I didn't submit them as a PR at the time.)

Additionally, I've added some detailed comments to `CorrectNickErrors` and `CheckStringForErrors` to explain the differences between the two methods. I imagine that change will be less controversial, so I can submit those comment changes as a separate PR if desired.


Narishma-gb:
Recent updates tend to reduce the number of hardcoded values, so I'd be more inclined towards using a character in definitions: `DEF some_tile EQU '♂'`. However, I can admit that graphic font is a special case and could have its own rules.

About the `CorrectNickErrors` table, I suggest using the same structure as in spaceworld (and pokesilver). It's clearly intended for the Japanese charmap, as intervals match perfectly the gaps in the kana definitions. This would be:
```
	;      ≥                <
	db '<NULL>',        'ガ'
	db 'ド' + 1,        'バ'
	db 'ボ' + 1,        'が'
	db 'ど' + 1,        'ば'
	db 'ぼ' + 1,        'パ'
	db 'ぽ' + 1,        '<TRAINER>'
	db '<TRAINER> + 1', '　'
	db -1 ; end
```
This would probably need some comment, in order to explain that `<TRAINER>` aside,  what is permitted below $80 will result in duplications of some of the font characters, all those that use a common charmap with kanas that can accept diacritic (there is also a blank tile added on the line above, that will normally remain invisible, except maybe in a few cases such as single spaced menus).

Rangi42:
I think it will be easier because `charmap "◀", '<KE>' ; $71` tells people "Everywhere that the `◀` character is used in battle, it can safely replace `<KE>` / `<KE>` is not needed in those screens". So maybe they run out of room as they're rearranging their font, but now they know they can safely load `◀` on-demand in place of `<KE>`.

Also, while we're touching the charmap anyway, it would be worth changing the string literals to character literals. RGBDS supports `charmap "C", value` *and* `charmap 'C', value`, with no difference in functionality -- but `value` has to be numeric, so `$xx` works, and `'charvalue'` works, but `"strvalue"` does not. I think `charmap 'A', 'B'` will be simpler to read than `charmap "A", 'B'`.

PR #1241, opened 2026-02-10
Change Nightmare condition so that the pkm needs to be asleep

### LINKED ISSUE #1229: Bug: AI can use Nightmare if the player has a status condition that isn't sleep
This function is the issue, since only if the player has no status, then it will consider it redundant. 

```
AI_Redundant:

	...

.Nightmare:
	ld a, [wBattleMonStatus]
	and a
	jr z, .Redundant
	ld a, [wPlayerSubStatus1]
	bit SUBSTATUS_NIGHTMARE, a
	ret
```
The better choice is to consider anything except sleep as redundant:
``` diff
.Nightmare:
	ld a, [wBattleMonStatus]
-	and a
+	and SLP_MASK
	jr z, .Redundant
	ld a, [wPlayerSubStatus1]
	bit SUBSTATUS_NIGHTMARE, a
	ret
```

MichailTalReborn:
This PR changes the logic of applying Nightmare by the AI. Instead of checking for any status (bug), it checks whether the target is asleep as intended

Fixes #1229

### DIFF
--- a/engine/battle/ai/redundant.asm
+++ b/engine/battle/ai/redundant.asm
@@ -113,7 +113,7 @@ AI_Redundant:
 
 .Nightmare:
 	ld a, [wBattleMonStatus]
-	and a
+	and SLP_MASK
 	jr z, .Redundant
 	ld a, [wPlayerSubStatus1]
 	bit SUBSTATUS_NIGHTMARE, a


Pquasar:
for pokecrystal i think what the owners prefer is to note any bugs in bugs_and_glitches.md instead of actually fixing them

PR #1243, opened 2026-02-19
Rename `unleashenergy` to `startbide`

### LINKED ISSUE #1174: `unleashenergy` does not actually unleash energy
The `unleashenergy` (BattleCommand_UnleashEnergy) command does not actually unleash energy, nor does it have anything to do with unleashing energy.

This command initializes Bide's damage counter, determines the number of turns for Bide, and then it ends the turn. It is only called on the first turn of Bide; subsequent uses of Bide go through `storeenergy` which will skip over this command once Bide is finished.

It should probably be named something more accurate, like `initbide`, `startbide`, or `resetbide`.

runlevel5:
## Summary

- Renames the `unleashenergy` battle command (and `BattleCommand_UnleashEnergy`) to `startbide` (`BattleCommand_StartBide`), since it initializes Bide rather than unleashing energy.
- The actual energy unleashing is handled by `storeenergy` (`BattleCommand_StoreEnergy`) when Bide's turn counter reaches zero.
- Pure label/macro rename across 5 files; produces byte-identical ROMs (`make compare` passes).

Fixes #1174

### DIFF
--- a/data/battle/effect_command_pointers.asm
+++ b/data/battle/effect_command_pointers.asm
@@ -38,7 +38,7 @@ BattleCommandPointers:
 	dw BattleCommand_Conversion
 	dw BattleCommand_ResetStats
 	dw BattleCommand_StoreEnergy
-	dw BattleCommand_UnleashEnergy
+	dw BattleCommand_StartBide
 	dw BattleCommand_ForceSwitch
 	dw BattleCommand_EndLoop
 	dw BattleCommand_FlinchTarget
--- a/data/moves/effects.asm
+++ b/data/moves/effects.asm
@@ -797,7 +797,7 @@ Bide:
 	checkobedience
 	doturn
 	usedmovetext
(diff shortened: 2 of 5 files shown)


Rangi42 (changes requested):
The complete Bide move script has three unique commands:

<pre><code>
Bide:
	<b><i>storeenergy</i></b>
	checkobedience
	doturn
	usedmovetext
	<b><i>startbide</i></b>
	resettypematchup
	checkhit
	moveanim
	<b><i>bidefailtext</i></b>
	applydamage
	checkfaint
	buildopponentrage
	kingsrock
	endmove
</code></pre>

If the `storeenergy/unleashenergy` pair is misnamed, then I think both should be renamed, and maybe all three -- since Bide starts by storing energy, so it's not clear what the difference is between "`storeenergy`" and "`startbide`".

PR #1244, opened 2026-02-21
Indent Nested Unions

vulcandth:
Actually surprised there wasn't more in the pret repo's... Either way to aid readability nested unions should be indented.

### DIFF
--- a/ram/hram.asm
+++ b/ram/hram.asm
@@ -63,23 +63,23 @@ hEnemyMonSpeed:: dw
 UNION
 ; math-related values
 
-UNION
-; inputs to Multiply
-	ds 1
-hMultiplicand:: ds 3
-hMultiplier::   db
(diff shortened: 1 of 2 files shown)


Rangi42 (changes requested):
I'm also surprised we didn't have more nested UNIONs! I'll bet pokered has more though. As such I'd like @dannye's opinion on whether this is a reasonable formatting.

Also the STYLE.md guide should mention this:

```diff
 DEF BUT_ONLY_RELATED_CONSTANTS EQU 5
+
+; Unions are not indented
+UNION
+wOneOption:: dw
+NEXTU
+wAnotherOption:: db
+	; But nested unions are
+	UNION
+	wNested1:: db
+	NEXTU
+	wNested2:: db
+	ENDU
+ENDU
```

PR #1245, opened 2026-02-23
Fix the `GoldenrodUndergroundSwitchRoomEntrances` door-switch macros

Rangi42:
[Bulbapedia documents](https://bulbapedia.bulbagarden.net/wiki/Goldenrod_Tunnel#Goldenrod_Department_Store_basement) how the Goldenrod Underground switch room works:

> ...the basement maze involves three switches along the north wall, labelled 1 to 3 from right to left, that, when turned on or off, open or close doors in the maze. The doors are internally numbered according to the following diagram:
> 
> image

The `ugdoor_def` and `changeugdoor` macros in this PR correspond to that diagram.

The previous `ugdoor` and `doorstate` macros had a number of problems:

- They had swapped X and Y values
- They confused the OPEN and CLOSED block IDs
- They declared 16 doors, when actually there are 6 horizontal doors and 5 vertical ones, with the vertical ones using 2 blocks each
- They declared 14 `EVENT_SWITCH_` flags, when actually there are 3 switches and 11 door states

I did not keep the original macro names, partly because I think the new ones better match typical script commands (`changedoorstate` vs `changeblock`, `ugdoor_def` vs `warp_def`) and partly to avoid any confusion about whether someone's code has old vs new behavior.

Mostly what I'm looking for in a review is feedback on whether I'm going too far with macros and `for` loops.


Rangi42:
This is an example of what it looks like to shift/change all the door coordinates and IDs: https://github.com/Rangi42/polishedcrystal/commit/46f7d69de0d06e6ea82025807a3c56ac94bad2cc#diff-c0a6db620bab535fdf6d747b731136e0a9160cf1a95711d4801da7d7e1e34b3d

dannye (approved):
> Mostly what I'm looking for in a review is feedback on whether I'm going too far with macros and `for` loops.

Nope, makes sense to me and seems like a good idea to me 👍

PR #1252, opened 2026-05-05
Fix `tools/make_patch.c` to be endian-independent

Rangi42:
Many thanks to @ISSOtm for discovering this one. We've been playing around with an iMac G3, with its big-endian PowerPC processor, and successfully compiled RGBDS but ran into this problem with building pokecrystal.


Rangi42:
Note to self to port this change to the other repos.

mid-kid:
Damn, I don't think I'd ever have caught this as an endianness issue. Yet another reminder of why void pointers are evil. Thanks!

PR #1255, opened 2026-05-14
Add some comments/constants, minor fixes

Narishma-gb:
Here is the twin PR of pret/pokegold#145

It's still in draft mode, I have a couple banks left to do on pokesilver, and some additional things may appear.

The renaming in `battle_transition.asm` can definitely be improved. My goal was to make it more clear that the GFX is loaded for wild or trainer battle similarly. And comment the use of `wDecompressScratch` as a temporary solid tile pattern to later fill visible BGMap area with a single call to `Request2bpp`.


Narishma-gb:
Final edit here, syncing with the PR in pokegold. I went a bit sideways, as when checking a `vBGMap2` reference, I noticed several places where data is loaded in VRAM bank 1, but symbols names in bank 0 are used instead.
Addresses are identical, it's simply for code understanding. This second commit can therefore be ignored if it's not worth it, the actual code fixes are all in the first commit.

Rangi42 (approved):
👍 🚀

Issue #1230, opened 2026-01-30
Replace the tools/lzcomp compressor?


Rangi42:
I'd like to discuss replacing `tools/lzcomp` (our vendored copy of [ax6's repo](https://github.com/aaaaaa123456789/lzcomp/)) with [`lzss.c`](https://pastebin.com/vuRHsZwf) (after editing it to use our `tools/common.h` utilities).

Pros:

- lzss.c is shorter (320 lines of code, versus 995 dense lines across 13 files).
- lzss.c is easier to read as a self-documenting explanation of how LZ compression happens, and what the flags in `gfx/lz.mk` would be varying.
- lzss.c would be easier to edit if someone wants to change how the whole algorithm behaves (i.e. changes that affect the `home/decompress.asm` decompressor too). (For example, https://github.com/Rangi42/polishedcrystal/pull/1343.)

Cons:

- `lzcomp` will automatically give users better compression just by removing the `gfx/lz.mk` flags and letting it brute-force pick the best heuristic method.
  - Caveat: `lzcomp` is not actually capable of the *best* compression; "[`dpcomp`](https://github.com/meithecatte/lzcomp/commit/59f4b37fd9ffae00bd71da6ee42378a5a571b683)" does better or at least as well for every file I've tested it on.
- Users might be used to running `./tools/lzcomp`, so this could be disruptive.

I'd support us providing a `dpcomp`-only program in a separate repo, so users who want better compression can download it and replace `lzss`. That way the "best" compressor could be developed independently of pokecrystal, pokegold, pokegold-spaceworld, and other projects like pokesilver. It could even add opt-in support for other LZ-related formats, like [these variants](https://github.com/bonimy/MushROMs/tree/master/doc) or the experimental new opcodes linked above. Meanwhile the compressor bundled with pokecrystal would be just the essentials.

Rangi42:
lzss.c does not currently support *de*compressing LZ files, so that would have to be implemented. Its "`verbose`" output mode could also be removed.

vulcandth:
It is certainly easier to read than the multiple folder w/ files. I support this.

Rangi42:
Here's the source for *just* [dpcomp.c](https://pastebin.com/MYEXYK2n). Possibly it could be combined with a more lzss-like approach for matching and optimal compression in one program.

Narishma-gb:
I recently ran into some gfx that uses the `repetitions` option of `lzcomp` compressor for matching.
[...]

PR #1257, opened 2026-06-07
Tools accept `-` as a placeholder filename for stdin or stdout

Rangi42:
This is a common convention for CLI tools, and is already accepted by `lzcomp`.

When `-` stands for stdin, we need to allow reading it even though it's unseekable so its size cannot be measured in advance.

- Most of the tools call `read_u8()`, which now handles this case via `read_stdin()`.
- `scan_includes` does its own reading, and now has a stdin case because it could make sense to `cat` or `echo` something into `| scan_includes -`.
- `make_patch` calls `xfsize()` to get the size of the original ROM. This program takes four input file, and it doesn't make sense to pipe the ROM specifically as stdin, so it's okay for this to still fail with "Could not measure file" if you do that.

(Part of my motivation for this PR is that I'm working on the new dpcomp+matching LZ compressor from #1230, which really ought to still support `-` for input+output.)

### DIFF
--- a/tools/common.h
+++ b/tools/common.h
@@ -111,10 +111,13 @@ long xfsize(const char *filename, FILE *f) {
 uint8_t *read_stdin(long *size) {
 	uint8_t buffer[0x1000] = {0};
 	*size = 0;
-	uint8_t *data = xmalloc(*size);
+	uint8_t *data = malloc(0);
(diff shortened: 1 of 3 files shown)


mid-kid:
Nice improvement, thanks!

PR #1258, opened 2026-06-10
Replace lzcomp with lzcompress

### LINKED ISSUE #1230: Replace the tools/lzcomp compressor?

Rangi42:
Fixes #1230

This replaces our vendored copy of [ax6's `lzcomp`](https://github.com/aaaaaa123456789/lzcomp/) with a new `lzcompress` tool (named similarly to pokered's `pkmncompress`). Unlike `lzcomp`, `lzcompress` has just two compression methods: the `--matching` one (with various Boolean flags to reproduce all the original files, even in the non-English disassemblies), and the dynamic "`dpcomp`" one (ported from [mei's fork](https://github.com/meithecatte/lzcomp/), it does at least as well as `lzcomp`'s methods and often a little better on every file I've tested[^1]).

There's some redundancy between the code for the two compression methods -- they have different ways of storing the data associated with each LZ command, so they each need their own functions to measure command size and to output command bytes -- but overall it's smaller (and I hope easier to read/maintain/extend) than the previous files.

Chosen names and descriptions can be compared with those in:

- https://github.com/pret/pokecrystal/blob/master/home/decompress.asm
- https://github.com/pret/pokemon-reverse-engineering-tools/blob/master/pokemontools/lz.py
- https://github.com/bonimy/MushROMs/blob/master/doc/LC_LZ3%20Compression%20Format.md

For testing purposes, you can compile with a lot of warning and sanitizer flags enabled (mostly copied from rgbds's `make develop` debugging ones):

```make
lzcompress: CFLAGS := -O3 -std=c11 -Wall -Wextra -pedantic \
          -Wno-unknown-warning-option \
          -Wcast-align -Wcast-qual -Wduplicated-branches -Wduplicated-cond \
          -Wfloat-equal -Wlogical-op -Wnull-dereference -Wshift-overflow=2 \
          -Wstringop-overflow=4 -Wtrampolines -Wundef -Wuninitialized -Wunused -Wshadow \
          -Wformat=2 -Wformat-overflow=2 -Wformat-truncation=1 -Wno-format-nonliteral \
          -ggdb3 -Og -fno-omit-frame-pointer -fno-optimize-sibling-calls \
          -D_GLIBCXX_ASSERTIONS -D_LIBCPP_HARDENING_MODE=_LIBCPP_HARDENING_MODE_DEBUG \
          -fsanitize=address -fsanitize=undefined -fsanitize=float-divide-by-zero
```

[^1]: If you delete `gfx/lz.mk`, then every file in pokecrystal becomes at least as small with `lzcompress` or forked `dpcomp` than with original `lzcomp`. The total size is 368,817 bytes with matching compression; 350,501 bytes with optimal `lzcomp`; and 349,632 bytes with optimal `dpcomp`/`lzcompress`. I've also tested with pokegold, pokered, polishedcrystal, etc and gotten similar results.)


Rangi42:
@Narishma-gb Before we merge this or the pokegold PR, please try using `lzcompress` in pokegold-kr, see if its `--long-32` flag can work for those two nonmatching files.

Rangi42:
The single-file tools is pretty long (832 lines), since it has two compressors and one decompressor, plus the usual argument parsing. Here's its skeleton:

```c
#define PROGRAM_NAME "lzcompress"
#define USAGE_OPTS "[-h|--help] [-u|--uncompress] [-m|--matching [-a|--align alignment] [-bcdilnr]] infile outfile"

#include "common.h"

/******************************************** LZ format *******************************************/

// ...#define constants...

enum lz_command {
	// ...
};

struct command {
	// ...
};

/**************************************** Argument options ****************************************/

struct Options {
	// ...
};

struct Options options = {0};

void parse_args(int argc, char *argv[]) {
	// ...
}

/************************************** Matching compression **************************************/

// ...288 lines...

/********************************** Optimized dynamic compression *********************************/

// Original `lzcomp` implementation by ax6 <https://github.com/aaaaaa123456789/lzcomp>
// Dynamic programming `dpcomp` algorithm by mei <https://github.com/meithecatte/lzcomp>
// Licensed with the Unlicense into the public domain <https://unlicense.org>

// ...244 lines...

/****************************************** Decompression *****************************************/

// ...159 lines...

/********************************************** Main **********************************************/

int main(int argc, char *argv[]) {
	parse_args(argc, argv);

	argc -= optind;
	argv += optind;
	if (argc != 2) {
		usage_exit(1);
	}

	(options.decompress ? decompress : options.matching ? compress_matching : compress_dynamic)(argv[0], argv[1]);

	return 0;
}
```

Rangi42:
Note: after doing `make compare`, you can verify that the non-matching compression and decompression works for every file with:

```sh
for f in `find gfx -name '*.lz'`; do
  diff ${f%.lz} <(./tools/lzcompress ${f%.lz} - | ./tools/lzcompress -u - -);
done
```

(I've run this with `lzcompress` built with the high-warning/sanitizers-enabled compiler flags from above, and gotten zero errors.)

tools/lzcompress.c
```diff
@@ -0,0 +1,832 @@
...
+uint8_t *raw_data = NULL;
+unsigned int raw_data_size = 0;
+
+unsigned int min(unsigned int a, unsigned int b) {
+	return a < b ? a : b;
+}
+
+unsigned int command_size(struct command command) {
+	unsigned int header_size = 1 + (command.length > SHORT_MAX_LENGTH);
+	if (command.cmd & 4) {
```
Rangi42: Just check for each of the three lookback commands, `& 4` is too obscure.

tools/lzcompress.c
```diff
@@ -0,0 +1,832 @@
...
+	return n;
+}
+
+int encode_delta(int pos, int at) {
+	return at - pos >= -LOOKBACK_MAX_OFFSET ? at - pos : at;
+}
+
+void find_optimal_commands(void) {
+	best_size = xmalloc(sizeof(*best_size) * (raw_data_size + 1));
+	best_command = xmalloc(sizeof(*best_command) * (raw_data_size + 1));
```
Rangi42: `free` these if possible.

tools/lzcompress.c
```diff
@@ -0,0 +1,832 @@
...
+	free(commands);
+}
+
+/****************************************** Decompression *****************************************/
+
+struct command *read_commands(char const *filename, unsigned int *num_commands) {
+	long filesize;
+	uint8_t *data = read_u8(filename, &filesize);
+
+	struct command *command_data = xmalloc(sizeof(*command_data) * filesize);
```
Rangi42, line 665: Comment on why `filesize` is a sufficient upper bound (there's at most one command per compressed byte in the file).

mid-kid:
Just out of curiosity, is there any file that ends up *bigger* with dpcomp than with lzcomp, even if just by 1 byte?

tools/lzcompress.c
```diff
@@ -0,0 +1,832 @@
...
+int main(int argc, char *argv[]) {
+	parse_args(argc, argv);
+
+	argc -= optind;
+	argv += optind;
+	if (argc != 2) {
+		usage_exit(1);
+	}
+
+	(options.decompress ? decompress : options.matching ? compress_matching : compress_dynamic)(argv[0], argv[1]);
```
mid-kid: I know I also like writing cheeky lines like this but come the fuck on now

tools/lzcompress.c
```diff
@@ -0,0 +1,832 @@
...
+		}
+		case LZ_FLIP: {
+			unsigned int base = (command->offset < 0 ? output_size : 0) + command->offset;
+			for (unsigned int j = 0; j < command->length; j++) {
+				output_data[output_size++] = flip_f(output_data[base + j]);
+			}
+			break;
+		}
+		case LZ_REVERSE: {
+			unsigned int base = (command->offset < 0 ? output_size : 0) + command->offset;
```
mid-kid, line 804: base = command->offset
if (command->offset >= 0) base += command->offset;

tools/lzcompress.c
```diff
@@ -0,0 +1,832 @@
...
+	// Options to control the mode of input and output
+	bool decompress;        // [-u] Decompress LZ-compressed input, instead of LZ-compressing raw input
+	// Option to control the compression algorithm
+	bool matching;          // [-m] Use a less efficient compression algorithm to match the original ROMs
+	// Options that apply to the dynamic and the matching compression algorithms
+	uint8_t alignment;      // [-a] Alignment of the final compressed byte
+	// Options that only apply to the matching compression algorithm
+	bool no_lookback_3;     // [-c] Don't use LZ_FLIP/LZ_REVERSE/LZ_LONG for uncompressed length 3
+	bool odd_alternate;     // [-d] Allow odd-length LZ_ALTERNATE
+	bool prefer_alternate;  // [-r] Use LZ_ALTERNATE over LZ_ITERATE when their compressed lengths tie
```
mid-kid: These explanations belong in --help output

mid-kid:
The dpcomp code isn't exactly a joy to read, so I'll keep this review a bit superficial.
Is there any reason not to split this up into multiple files?
The only feature I'm missing is how lzcomp would allow printing the command stream as macros for debugging purposes, but I suppose that's rather niche and nobody would really use that.

Rangi42:
Thanks for your comments!

> Just out of curiosity, is there any file that ends up bigger with dpcomp than with lzcomp, even if just by 1 byte?

I think polishedcrystal *might* have encountered a few that are 1 or 2 bytes bigger, but that could be because of our algorithm changes. I don't recall any from the canon repos that got bigger, and will be keeping an eye out for such as this gets PRed elsewhere.

> Is there any reason not to split this up into multiple files?

Partly that, although 800+ lines *is* long, it's still more convenient to scroll through it than to browse/share multiple files. Partly so the Makefile doesn't need a special case for it.

(After updating based on the review comments, it's at 841 lines. Compared to the next-largest make_patch.c being 531 lines, or gfx.c being 300 lines, ...or rgbasm's lexer.cpp being 2437 lines. 😅)

> The only feature I'm missing is how lzcomp would allow printing the command stream as macros for debugging purposes, but I suppose that's rather niche and nobody would really use that.

Yeah. I considered adding them, but I've only seen [those macros](https://pastebin.com/h2uqE8Dv) in Prism, and frankly I'm not even sure why Prism isn't using `INCBIN "bingo.tilemap.lz"` instead of this: (I've PRed Prism to not do that anyway)

```asm
BingoTilemapCompressed:
	lzrepeat 4, $7f
	lzzero 1
	lzrepeat 14, $01
	lzdata $02
	lzrepeat 4, $7f
	lzdata $03, $0b, $0b, $0a
	lzcopy normal, 11, -3
	lzdata $07
	lzcopy normal, 25, -20
	lzdata $08, $08, $09
	lzcopy normal, 11, -3
	lzdata $07
	lzcopy normal, 224, -60
	lzdata $04
	lzrepeat 14, $05
	lzdata $06
	lzend
```

Also the macro names don't match the constants I chose to keep using in lzcompress.c. Also also, I'd have to decide whether to support compressing directly to a text stream, or simplify the code but the user would have to do `lzcompress in.bin - | lzcompress --text - out.txt`.

tools/lzcompress.c
Rangi42: Ehh, that would repeat four lines across the three cases. Also if I were `const`-ing everything that can be `const` (or using a language where it's more normal/idiomatic to do that, even for local variables, e.g. Typescript), I'd want this to be initialized in a single assignment so it's not `mut`.

Rangi42:
> The only feature I'm missing is how lzcomp would allow printing the command stream as macros for debugging purposes, but I suppose that's rather niche and nobody would really use that.

Besides `lzcomp -d`, there's also Prism's do-one-thing-well [lzdump.c](https://pastebin.com/tUiNeu4R) tool for this. (Which is also easier to extend if/when people change how the LZ format works, e.g. to add [extended opcodes](https://github.com/Rangi42/polishedcrystal/pull/1343).)

PR #1261, opened 2026-06-22
Make the `connection` macro more convenient and foolproof

Rangi42:
Currently, you have to specify a direction mask (e.g. `NORTH | SOUTH | EAST`) and then each `connection` in order. This can be confusing because the order of `|`ing direction constants together doesn't matter, but the order of `connection`s does. (It's not even the predictable clockwise order "`north`, `east`, `south`, `west`", nor the English-typical order "`north`, `south`, `east`, `west`", nor even the Japanese-typical 東南西北 order "`east`, `south`, `west`, `north`", but the unmemorable order "`north`, `south`, `west`, `east`".)

With this PR, the direction mask byte is derived from the subsequent `connection` macros, similarly to how the `def_*_events` macros derive their count bytes from the subsequent `*_event` macros. There's also some sanity checking for directions in the wrong order (or duplicated).

@dannye @vulcandth If this PR looks good, I'll do the same for the other Gen 1+2 repos before merging.

### DIFF
--- a/data/maps/attributes.asm
+++ b/data/maps/attributes.asm
@@ -2,7 +2,7 @@ MACRO map_attributes
 ;\1: map name
 ;\2: map id
 ;\3: border block
-;\4: connections: combo of NORTH, SOUTH, WEST, and/or EAST, or 0 for none
+	REDEF CURRENT_MAP_ID EQUS "\2"
 	DEF CURRENT_MAP_WIDTH = \2_WIDTH
 	DEF CURRENT_MAP_HEIGHT = \2_HEIGHT
 \1_MapAttributes::
@@ -13,7 +13,10 @@ MACRO map_attributes
 	db BANK(\1_MapScripts) ; aka BANK(\1_MapEvents)
 	dw \1_MapScripts
 	dw \1_MapEvents
-	db \4
+	db MAP_CONNECTIONS_\2
+	; Define `MAP_CONNECTIONS_\2` after `db`ing it so that the `db`ed value
+	; gets updated when subsequent `connection`s modify `MAP_CONNECTIONS_\2`.
+	DEF MAP_CONNECTIONS_\2 = 0
 ENDM
 
 ; Connections go in order: north, south, west, east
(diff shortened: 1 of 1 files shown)


data/maps/attributes.asm
```diff
@@ -13,7 +13,8 @@ MACRO map_attributes
 	db BANK(\1_MapScripts) ; aka BANK(\1_MapEvents)
 	dw \1_MapScripts
 	dw \1_MapEvents
-	db \4
+	db MAP_CONNECTIONS_\2
+	DEF MAP_CONNECTIONS_\2 = 0
```
ISSOtm, line 19:
```suggestion
	; Defined after the above so it evaluates to this symbol's value at the *end* of the file.
	DEF MAP_CONNECTIONS_\2 = 0
```

ISSOtm (changes requested):
I'm always in favour of one source of truth, even if it makes for a more complex conversion process. I only wish one load-bearing weirdness was documented.

data/maps/attributes.asm
Rangi42: To avoid indirect "the above"/"the below" language:
```suggestion
	; Defined after `db MAP_CONNECTIONS_\2` so its value can be updated by `connection`s
	DEF MAP_CONNECTIONS_\2 = 0
```

ISSOtm: IMO that's unclear, so allow me further bikeshedding:
```suggestion
	; Defined after the `db` above so it uses the value this symbol has at the *end* of the file,
	; after being modified by `connection`s.
	DEF MAP_CONNECTIONS_\2 = 0
```

mid-kid:
This macro is plenty complex already, so I suppose this doesn't hurt too much. I would've personally preferred just making the order not matter, but that'd probably have required pre-allocating all the `db`s for the connections in map_attributes, making the code even more complicated.

Rangi42:
> I would've personally preferred just making the order not matter, but that'd probably have required pre-allocating all the `db`s for the connections in map_attributes, making the code even more complicated.

Yeah, and it would also have required a new macro to terminate the series of connections and "flush" any allocated ones. (Ironically pokered *could* do this since it has `map_header`+`end_map_header` instead of `map_attributes`, but let's keep it consistent.)

